PR #42, opened 2022-06-08
Add support for doors and lifts

koonpeng:
### Implemented feature

Adds support for doors and lifts, also fixes a bunch of picking related bugs.

Some of the bugs fixes are:
* Objects still being selected when they are behind a egui widget.
* Some objects like walls that respawn their meshes losing their highlights.
* Unnecesssary extra traversal of the entity hierarchy when making meshes pickable.
* Camera mode keyboard shortcuts not being disabled when editing a text field.

### DIFF
--- a/rmf_sandbox/Cargo.lock
+++ b/rmf_sandbox/Cargo.lock
@@ -3648,11 +3648,11 @@ dependencies = [
  "futures-lite",
  "rfd",
  "serde",
+ "serde_json",
  "serde_yaml",
  "surf",
  "wasm-bindgen",
  "web-sys",
- "yaml-rust",
 ]
 
 [[package]]
--- a/rmf_sandbox/Cargo.toml
+++ b/rmf_sandbox/Cargo.toml
@@ -17,9 +17,9 @@ bevy_mod_picking = "0.7"
 rfd = "0.8.2"
 serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.8.23"
+serde_json = "1.0"
 wasm-bindgen = "0.2"
 web-sys = { version = "0.3.56", features = ["console"] }
-yaml-rust = "0.4"
 futures-lite = "1.12.0"
 bevy = { version = "0.7" }
 dirs = "4.0"
--- a/rmf_sandbox/assets/demo_maps/office.building.yaml
+++ b/rmf_sandbox/assets/demo_maps/office.building.yaml
@@ -233,5 +233,17 @@ levels:
       - [38, 24, {alpha: [3, 1], texture_name: [1, default]}]
       - [37, 38, {alpha: [3, 1], texture_name: [1, default]}]
 lifts:
-  {}
+  main_lift:
+    x: -0.185
+    y: 5.511
+    yaw: 0.021
+    depth: 1.895
+    width: 1.895
+    doors: {}
+    lowest_floor: L1
+    highest_floor: L1
+    initial_floor_name: L1
+    level_doors: {}
+    plugins: false
+    reference_floor_name: L1
 name: building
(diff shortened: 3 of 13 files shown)


mayman99 (approved):
I did run it and lifts seem to be okay :+1: 
p.s. didn't review code as I didn't have a chance to read on rust yet

luca-della-vedova:
I can't quite get the JSON input to work properly, it seems whatever I write is always considered "invalid json", i.e.:

https://user-images.githubusercontent.com/9111558/179193506-37872ae1-9da3-4e78-b087-a77cdcbc74ce.mp4

I'm not too sure how to debug it, I added a print statement to the `value_json` parameter that is passed to the function and saw the following (example):

```
Value is "{\n\"test\": 123\n}"
```

I'm not sure if it's intended behavior of the debug printer but I find the escaping of `"` a bit suspicious and I can see why a json parser would fail if it tried to read that string. I wonder if the escaping is happening the scenes and that might cause the json parser to fail.
I also tried to get rid of json altogether and use yaml (imho preferred option to avoid adding another dependency just for a temporary feature until we have a better UI to set lift doors) but I had a missing trait error and I'm not familiar enough with rust to figure out what might be happening. I noticed that the function signature for [serde_yaml::from_str](https://docs.serde.rs/serde_yaml/fn.from_str.html) and [serde_json::from_str](https://docs.serde.rs/serde_json/fn.from_str.html) are slightly different so I guess the reason might be there.

```
luca@luxury:~/ws_sim/rmf_sandbox/rmf_sandbox$ cargo run
   Compiling rmf_sandbox v0.0.1 (/home/luca/ws_sim/rmf_sandbox/rmf_sandbox)
error[E0277]: the trait bound `for<'de> T: Deserialize<'de>` is not satisfied
    --> src/widgets.rs:36:19
     |
36   |             match serde_yaml::from_str::<T>(self.value_json.as_str()) {
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `for<'de> Deserialize<'de>` is not implemented for `T`
     |
     = note: required because of the requirements on the impl of `DeserializeOwned` for `T`
note: required by a bound in `serde_yaml::from_str`
    --> /home/luca/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_yaml-0.8.23/src/de.rs:1445:8
     |
1445 |     T: DeserializeOwned,
     |        ^^^^^^^^^^^^^^^^ required by this bound in `serde_yaml::from_str`
help: consider further restricting this bound
     |
26   |     T: Deserialize<'de> + Serialize + for<'de> building_map::_::_serde::Deserialize<'de>,
     |                                     ++++++++++++++++++++++++++++++++++++++++++++++++++++

For more information about this error, try `rustc --explain E0277`.
error: could not compile `rmf_sandbox` due to previous error
```

koonpeng:
> I can't quite get the JSON input to work properly, it seems whatever I write is always considered "invalid json", i.e.:

The validator follows the format of the building yaml file, an example can be found here.

https://github.com/open-rmf/rmf_demos/blob/9cf5b279ae22f04ac5cd81a5bbcda9eb5ec404b8/rmf_demos_maps/maps/hotel/hotel.building.yaml#L1030-L1036

> I also tried to get rid of json altogether and use yaml (imho preferred option to avoid adding another dependency just for a temporary feature until we have a better UI to set lift doors) but I had a missing trait error and I'm not familiar enough with rust to figure out what might be happening. I noticed that the function signature for [serde_yaml::from_str](https://docs.serde.rs/serde_yaml/fn.from_str.html) and [serde_json::from_str](https://docs.serde.rs/serde_json/fn.from_str.html) are slightly different so I guess the reason might be there.

`serde_yaml` requires a `DeserializeOwned` instead of `Deserialize`, try changing the template constraints and see if it works.

koonpeng:
I remembered the reason why `serde_json` was used instead of `serde_yaml`. When converting the object to the string representation, I want it to use the json format. `serde_yaml::to_string()` converts to the yaml format and I couldn't find anyway to tell it to use the json representation.

mxgrey (approved):

PR #46, opened 2022-07-28
Feature/add cameras

PranayShirodkar:
Adding support for cameras to rmf_sandbox.

A new button "add camera" is made available to the user, which allows the user to position a camera in 3d space and configure the direction the camera faces.

The camera attributes include x, y, z, pitch and yaw. Z is required because cameras are typically positioned close to the ceiling. Pitch is required because ceiling mounted cameras are usually pitched downwards.

Other camera intrinsic properties:
1. Image FOV
2. Image width
2. Image height
3. Update rate

Preview:
Screenshot from 2022-07-28 11-34-28
The camera is at the bottom left of the office (not in the meeting room), selected in green
2nd camera is facing the office main door


rmf_sandbox/src/camera.rs
```diff
@@ -0,0 +1,35 @@
+use bevy::prelude::*;
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Component, Clone, Default)]
+pub struct Camera {
```
mxgrey, line 5: In Bevy 0.8 there's a new component called `Camera` in `bevy::prelude::*` so I think that's going to cause some naming conflicts with this struct.

Maybe we can name this struct `PhysicalCamera` instead?

PranayShirodkar: It looks like [Bevy 0.7 already has `Camera`](https://docs.rs/bevy/0.7.0/bevy/prelude/struct.Camera.html), how is it working now?

Since we want to differentiate between the camera that controls what is visible on the UI and the cameras inserted into the scene, I think `PhysicalCamera` or `CameraEntity` make sense

PranayShirodkar:
When the user inserts a camera, a box shape would represent the camera, but the user cannot tell which face of the box is the camera's front direction. For that reason, a new shape DirectionalBox is defined that is a box with a pyramid on the right face. The pyramid indicates the camera direction.
image

mxgrey (approved):

PR #47, opened 2022-08-02
Migrate to 0.8 + Cleaner UI/UX

mxgrey:
This PR migrates the sandbox to Bevy v0.8. I also did a major overhaul of the 3D user interactions that I think make the application much cleaner.


arjo129 (approved):
Works for me.

PR #50, opened 2022-08-04
Updates  demo world to be loaded from assets file.

arjo129:
This PR removes the giant string we have in demo_world and instead loads it from the assets folder.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>


codebot (approved):
LGTM, thank you for fixing this hack

PR #53, opened 2022-09-19
Fix support for cartesian_meters format

mxgrey:
This PR fixes support for `.building.yaml` files that use `caretesian_meters` and also makes several fields optional so the file parser is less strict.


mayman99:
Perfect, it works!

mayman99 (approved):
`coordinate_system: cartesian_meters` now works!

PR #64, opened 2022-10-05
Door widget and visual cues

mxgrey:
This PR adds a widget that lets users change the properties of their doors (e.g. swing, slide, double-swing, double-slide).

It also introduces visual cues rendered in the environment to help users know what kind of door each one is, and what its range of motion will be. See the screenshots below:

door_cues
door_cues_2
door_cues_3


codebot (approved):
This is awesome :tada: Tested on Linux and in the web browser. LGTM :+1:

PR #65, opened 2022-10-07
Correct the parsing of door properties

mxgrey:
When implementing the conversion from the legacy `.building.yaml` format to the new `.site.ron` format, I misinterpreted the meaning of some of the door parameters. This PR fixes the conversion to reflect the real meaning of those parameters.


codebot (approved):
Thanks! Works as expected :+1:

PR #66, opened 2022-10-07
flattened_FOO_offset is no longer saved by Traffic Editor I

codebot:
As of https://github.com/open-rmf/rmf_traffic_editor/pull/418/files#diff-fb5555f6057f74143c5aeb3cbec283ba77df9b1aaf91780957d140986dce2168L223-L224  the Qt-based Traffic Editor no longer saves `flattened_x_offset` and `flattened_y_offset` fields for levels in its YAML output. This PR just makes those fields optional. They have always been set to zero anyway.

### DIFF
--- a/rmf_site_format/src/legacy/level.rs
+++ b/rmf_site_format/src/legacy/level.rs
@@ -26,7 +26,9 @@ pub struct Level {
     #[serde(default)]
     pub drawing: LevelDrawing,
     pub elevation: f64,
+    #[serde(default)]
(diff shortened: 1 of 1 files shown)


mxgrey:
Those legacy parameters won't be used anymore in the new tool, so to make things as simple as possible I've push a commit to remove them from the parser entirely. When they exist in a file that's being parsed, they will simply be ignored.

mxgrey (approved):

PR #68, opened 2022-10-09
Allow default values for crowd sim and lifts

mxgrey:
Following the theme of https://github.com/open-rmf/rmf_site/pull/66 this should help make our building parsing more flexible towards the old building parameters that are optional.


codebot (approved):
LGTM, thanks! :tada:

PR #77, opened 2022-11-08
an attempt at adding opacity to drawings

codebot:
@luca-della-vedova  this is an attempt to add opacity to drawings. It may not be a step in the right direction, however, so I made it as a separate PR that can be :no_entry: if another approach is better.

To render both the drawing and lane graph, I set the drawing transparent and slightly higher in the Z stack than the floor, which is currently opaque.

Unfortunately, this means that clicks onto the floor surface and lane graph are registered on the drawing instead of the other objects. We may instead want to do what we do in Traffic Editor I, where the drawing is opaque and "below" the floor, and we render the floor as transparent instead.

There was also an attempt in here to make the floor opacity show up in the GUI Inspector, but that seemed to not be working. I need to learn more about how the Inspector is populated.


mxgrey:
Thanks for exploring this. I had a similar but sort of inverted idea for how to approach this feature.

Luca's PR is really meant to get the ball rolling on importing Drawings. There are several steps remaining that we're planning on leaving for future PRs since this isn't an urgent feature for our upcoming deliverable. The next steps would be:
* Have a "layering" system similar to most 2D drawing tools that lets us change the order that 2D elements lay on top of each other. We could then let users shift Drawings, lanes, and floors up/down in the layers so users can see what they're most interested in.
* Have a toggle for projecting a transparent copy of each Floor element into a layer that's exactly one higher than the Drawing layer(s). The user could set the transparency level from the inspector widget.

Issue #76, opened 2022-11-08
Better door displays


mxgrey:
In addition to the floor markings (which could be made to look nicer) we should:
- [ ] By default doors should be displayed as open. This should create a more open feel to the environment and make more sense for drawing lanes through it.
- [ ] Have animations for the doors, especially animations that can be triggered by site editor plugins.
- [ ] Improve the static visual cues that show how the doors move.

mxgrey:
Closed by https://github.com/open-rmf/rmf_site/pull/379

Issue #60, opened 2022-10-04
Preview of Lane Motions


mxgrey:
The terminology used for the parameters in the forward and reverse lane motions might not be clear. We can add tooltips to help explain their meaning, but ultimately it would be best to show an animated preview of what kind of motion a user should expect given their lane settings. That way there's no ambiguity about the meaning of the parameters.

mxgrey:
Currently being worked on by @althea28

PR #78, opened 2022-11-11
Add preview and parameter editing to physical cameras

luca-della-vedova:
### Implemented feature

This PR builds on top of #73 (the two features are quite decoupled but #73 introduced an inspector widget for arbitrary values and it's used here also, with a small refactor to make it generic to work with either floats, signed and unsigned integers).
The PR implements preview behavior for cameras, and sets up infrastructure for general previewable items, to help towards addressing #76  and #60.
image

A new empty marker component `PreviewableMarker` is added to all entities that can be previewed. The marker component will be used by the inspector widget to determine whether to add a preview button or not when the entity is selected, it will also be used to filter entities to process in the new `manage_previews` system.
For cameras itself, three systems are added, one to spawn the preview into a new window, one to read `Changed` events and update the live preview accordingly and one to handle window close events to keep a coherent state.

A few points I'm mostly unsure about:
* The marker component approach: The idea is that if in the future we want to make `Lane` objects previewable we just add the marker component to them and the `Preview` button, including its events, will be automatically added to the UI. However this marker component approach makes constructing the structs a bit more tedious (an extra empty field, but not too bad).
* Split vs unified `manage_previews` function, it started with a Query that would get all the previewable objects and `Option`s to figure out what they are exactly to spawn the previews. I'm figuring it might become a bit too large so it might be better to split into different functions.
* Bookkeeping and despawning previews: I tentatively implemented an `Option<Entity>` for events to spawn previews, the idea in the beginning was that only one preview would be available at a time, when it's `None` it would be cleared and when it was set it would be added / updated. However while working with cameras I realized there is value to having multiple previews open at the same time so the idea now is that `None` means "despawn all previews" and set means "spawn one preview", there is currently no "despawn some previews" available.


rmf_site_editor/src/interaction/preview.rs
```diff
@@ -0,0 +1,166 @@
...
+    mut camera_children: Query<(Entity, &mut Projection), With<Camera>>,
+    mut create_window_events: EventWriter<CreateWindow>,
+    mut preview_windows: ResMut<CameraPreviewWindows>,
+) {
+    for event in preview_events.iter() {
+        if let None = event.entity {
+            // TODO clear all previews
+        }
+        if let Some(e) = event.entity {
+            if let Some(open_window) = preview_windows.0.get_mut(&e) {
```
mxgrey, line 92: If `CameraPreviewWindow` were a component then you could add an `Option<&CameraPreviewWindow>` to the `previewable` query and check its value after querying.

rmf_site_editor/src/interaction/preview.rs
```diff
@@ -0,0 +1,166 @@
...
+                                camera_properties.horizontal_fov.radians() / aspect_ratio;
+                        }
+                        let window_id = create_camera_window(
+                            &mut commands,
+                            child_entity,
+                            &camera_name,
+                            &camera_properties,
+                            &mut create_window_events,
+                        );
+                        preview_windows.0.insert(e, window_id);
```
mxgrey, line 117: If `CameraPreviewWindow` were a component then this line would change to
```
commands.entity(e).insert(CameraPreviewWindow(window_id));
```

mxgrey (changes requested):
This new feature is so great. I literally had fun playing around with moving the camera around the environment and giving it funny parameters.

I only have one suggestion which is to change the `CameraPreviewWindows` resource into a `CameraPreviewWindow` component. A `HashMap<Entity, T>` is functionally equivalent to `T` simply being a component, but making it actually a component should be more ergonomic and efficient overall.

luca-della-vedova:
Thanks for the feedback! Done https://github.com/open-rmf/rmf_site/pull/78/commits/96bb0b793570cbc53a265714ff4036f7efdb8c2b That `HashMap` was indeed a bit silly and reimplemented an ECS feature, not as well as the ECS itself!
Removing it also removed the need of having the "bookkeeping" function since that is done by the ECS itself, which is great.

mxgrey:
> A bookkeeping function could be kept in a slightly different form if we want to have the feature that deleting a camera also despawns its preview window.

I think that could be accomplished with [removal detection](https://bevy-cheatbook.github.io/programming/removal-detection.html). Probably look at the `WindowId` which was removed and then close that window. We can save that for a future PR, though.

PR #79, opened 2022-11-14
Multi-floor + Lifts

mxgrey:
This PR introduces support for multi-floor sites and lifts. The changes in this PR are extensive, but here's a short list of some of the most significant changes:

### Levels Panel
* Toggle between which level to use
* Create a new level by clicking `Add`
* Change the name or elevation of each level
* Levels are shown in descending order of elevation
* Changing the elevation of a level will automatically reorder them

### Lift Inspector
* Inspect and edit properties of a lift
* Add lift cabin doors
* Toggle which floors each lift door can be used on
* Disabling a lift door on all floors will remove the door from the lift entirely

### Lift Door Place-mat Gizmo
* Each lift has four "place-mat" gizmos, one in front of each of its (potential) cabin doors
* The place-mat gizmo will be colored green if its lift door can open on the currently viewed floor, red if it cannot
* Clicking the gizmo will toggle the door availability on/off for that floor

All modifications are visible live while they are being made, as demonstrated in this video:

https://user-images.githubusercontent.com/1307217/201717559-7335961c-9d1b-42d9-a509-c8fe890ff63f.mp4

### Anchors
* Three types of reference frame for Anchors
  * site (cylinder): The anchor can be used at every level on the site and has the same (x, y) location at every level
  * level (sphere): The anchor can only be used inside of its parent level
  * lift (diamond): The anchor is associated with a lift. When the lift is translated or rotated, the anchor will move with it. The anchor can be used on any level.
* Categorized anchors can specify different points in space depending on which type of element (e.g. wall, floor, lane) is referencing it.
* Subordinate anchors have a position that is determined by the properties of some other entity. Currently this is only used for the lift door anchors, whose positions are determined by the lift cabin properties.

### Alignment
Legacy files in the `.building.yaml` format will automatically have their levels scaled and aligned based on measurements and fiducials.


luca-della-vedova:
I gave it a quick try and found an issue with coordinate transformation when adding waypoints inside lifts. The anchor coordinates seem off? Video showing the issue below:


https://user-images.githubusercontent.com/9111558/202068718-501eecfc-f55a-412b-abb0-134250256019.mp4

mxgrey:
> The anchor coordinates seem off?

Excellent catch. That used to work until I made some major changes to how anchors work, and I didn't think of rechecking that feature.

It should be fixed now by [da9004e](https://github.com/open-rmf/rmf_site/pull/79/commits/da9004ebfd54cb9528264e55925c40dad39a640f)

PR #88, opened 2022-11-18
Add support for local meshes and automatic lookup

luca-della-vedova:
### Implemented feature

This PR extends `AssetSource` to implement two additional variants, `Remote(String)` and `Search(String)`.
It also migrates `Model::Kind` to use `AssetSource` so all external assets use the same data structure.

To recap, `AssetSource` has the following variants:
* `Local(String)`: It encodes a path local to the building map file, currently only supported for drawings.
* `Remote(String)`: An asset name (the name itself is unchanged from the old `Kind` for now, so for example `OpenRobotics/Char`) to be searched in the remote server.
* `Search(String)`: An asset name (same as `Remote`) where hierarchical search will be performed, in sequence:
  * Look at the name as a subfolder of the path specified in the `$GZ_SIM_RESOURCE_PATH` environment variable, for example if its value was `/home/user` and the model's name was `OpenRobotics/Chair` it would try to load a `/home/user/OpenRobotics/Chair.glb` file.
  * Try the local cache, which is prepopulated with models fetched from remote.
  * Fetch from the remote server.

Most combinations of enum values and AssetSource work. The main corner case is when file is not set (i.e. for demo map), absolute paths for `Local` models can still be used but for drawings that is not possible.
Adding `Local` drawings for the demo map would work but introduce a warning when the map is loaded since the asset loader would try to look for it but fail, if the warning is OK I'll add the support.
My main worry is the hack I had to do to be able to use pattern matching at the `AssetServer` level. Since it only accepts strings and paths I implemented an `AssetSource::From(String)` and a `String::From(AssetSource)` to convert the data back and forth and not have to hardcode paths at the `AssetServer` level.

### Test it

* Export the `$GZ_SIM_RESOURCE_PATH` variable to a local folder.
* Add a model with the desired name and `Search` AssetSource to the project.
* Make sure there is a `model_name.glb` file in the folder pointed by the environment variable.


mxgrey (approved):
The `impl From<&AssetSource> for String` strategy is very clever, even if it feels a little dirty.

PR #89, opened 2022-11-20
Lights

mxgrey:
This PR adds a bunch of features for supporting physics-based lighting:
* Light panel
  * Toggle the default camera headlight (useful for looking at the scene, but interferes with the physics-based lights)
  * Toggle the rendering of the physics-based lights (makes the scene look good, but slows down performance)
  * Add a new light to the scene
  * Scroll through a list of existing lights and select them
* Light inspector
  * Change any property of a light, e.g. pose, color, type, brightness
  * Enable/disable casting shadows
* 3D Light Fixture
  * Simple 3D models are given for each light type
  * The model appearance depends on the light type
  * Click the 3D model to select it in the inspector
  * Click and drag the 3D model to move it around in the XY-plane
    * Z position cannot be dragged by the 3D model but it can be changed in the inspector panel after selecting the light
* File format supported in `.building.yaml`:
```
levels:
  L1:
    lights:
      - {pose: {trans: [1573, 783, 2.6]}, kind: {type: spot}}
      - {pose: {trans: [1273, 783, 2.6]}, kind: {type: point, enable_shadows: false}}
      - {pose: {trans: [1973, 633, 2.6]}, kind: {type: point, color: [0.5, 0.5, 1.0, 1.0]}}
```

A few notes of loose ends:
* There are no hover, selection, or dragging visual cues for the lights yet
* It seems that in Bevy, only **one** directional light can work at a time, and the camera headlight counts as a directional light. Directional lights won't be visible at all unless you turn off the headlight, and then only one of them will work per level.
* Spot lights are flaky (the light turns off) for certain (but not all) orientations that are rotated purely around the y-axis. I have absolutely no explanation for this, but I assume it's a bug in Bevy's shader.


rmf_site_editor/src/site/light.rs
```diff
@@ -15,15 +15,117 @@
...
+    physical_light_toggle: Res<PhysicalLightToggle>,
+    current_level: Res<CurrentLevel>,
+) {
+    for (e, parent) in &added {
+        // This adds all the extra components provided by all three of the
+        // possible light bundles so we can easily switch between the different
+        // light types
+        commands
+            .entity(e)
+            .insert(Visibility { is_visible: physical_light_toggle.0 })
```
luca-della-vedova, line 56: Should we add a `Category::Light` component to have the inspector plugin show the right label? Otherwise it just shows `<Unknown Type>`

rmf_site_format/src/legacy/building_map.rs
```diff
@@ -85,6 +85,16 @@ impl BuildingMap {
...
 
+            for light in &mut level.lights {
+                let p = tf.transform_point2(DVec2::new(
+                    light.pose.trans[0] as f64,
+                    light.pose.trans[1] as f64,
+                ));
+                light.pose.trans[0] = p.x as f32;
+                light.pose.trans[1] = -p.y as f32;
+                light.pose.rot.apply_yaw(Angle::Rad(-delta_yaw as f32));
+            }
```
luca-della-vedova, line 96: I don't know how deep we want to go into adding features to the legacy format but a yaw only angle doesn't sound quite sufficient for lights since all the current lights have symmetry on yaw and angle only has an impact on lights if roll or pitch are changed

luca-della-vedova (commented):
Works great! Just two minor points, the legacy format is not really critical so can be skipped if we don't want to go there

rmf_site_format/src/legacy/building_map.rs
mxgrey: This line of code is only for modifying the user-specified rotation based on how the levels were moved during the alignment process.

I'll tweak the example in the office demo to show a light with some pitch.

rmf_site_editor/src/site/light.rs
mxgrey: Whoops, good catch! Fixed: [2a8d533](https://github.com/open-rmf/rmf_site/pull/89/commits/2a8d533eea61512b420ecd49ede9c189e6916795)

rmf_site_format/src/legacy/building_map.rs
mxgrey: Now there's an example of [a spot light with some roll](https://github.com/open-rmf/rmf_site/blob/ce0fa6e9fe1c535445e50cdb61cff3aea52433fb/assets/demo_maps/office.building.yaml#L138) in the `.building.yaml`.

luca-della-vedova (approved):

PR #90, opened 2022-11-21
Allow light information to be exported

mxgrey:


codebot (approved):
This is super convenient! Thank you :bow: I just tested it in a few different ways and it works as expected.

The automatic headlight toggle on file load is also great; it's easy to disable the automatic headlight by having the file contain a directional light. Thanks!

PR #93, opened 2022-11-23
Try CI actions for occupancy grid branch

luca-della-vedova:
Introducing mapf in its current branch will break support for rust stable compiler, also the toolchain in this repo specifies using nightly so updating github actions to also use nightly rustc versions.

### DIFF
--- a/.github/workflows/ci_linux.yaml
+++ b/.github/workflows/ci_linux.yaml
@@ -14,19 +14,14 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        toolchain:
-          - nightly
-
     steps:
     - name: deps
       run: |
         sudo apt-get update
         sudo apt-get install libasound2-dev libudev-dev libgtk-3-dev
     - uses: actions/checkout@v3
-    - name: Setup toolchain
-      run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
+    - name: Setup toolchain to nightly
+      run: rustup default nightly
     - name: Build
       run: cargo build
     - name: Run tests
--- a/.github/workflows/ci_linux.yaml
+++ b/.github/workflows/ci_linux.yaml
@@ -14,12 +14,19 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        toolchain:
+          - nightly
+
     steps:
     - name: deps
       run: |
         sudo apt-get update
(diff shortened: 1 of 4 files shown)


luca-della-vedova:
Depends on https://github.com/open-rmf/mapf/pull/10, should probably merge that and revert the custom branch in the `Cargo.toml` before merging

mxgrey:
I'm a bit confused about why we would need nightly since the offending feature (GATs) [was stabilized in early November](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html). I wonder if the real issue is that the image used by the CI hasn't updated to the latest stable Rust toolchain. I'm going to see what happens if we run `rustup update` instead of  switching to nightly...

mxgrey:
Okay I'm noticing that `mapf` depends on a few other nightly features still...

To move the occupancy feature along I'm going to copy/paste over the `Cell` struct since it's the only thing from mapf we're actually using at the moment. I'd rather not begin requiring nightly for the site editor yet if we can avoid it.

luca-della-vedova:
> 



> I'm a bit confused about why we would need nightly since the offending feature (GATs) [was stabilized in early November](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html). I wonder if the real issue is that the image used by the CI hasn't updated to the latest stable Rust toolchain. I'm going to see what happens if we run `rustup update` instead of switching to nightly...

I think an issue I saw is that rustc does not seem to allow the `#![feature]` on stable compilers (i.e. [here](https://github.com/open-rmf/rmf_site/actions/runs/3524406049/jobs/5909765330#step:4:752)):

```
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> /home/runner/.cargo/git/checkouts/mapf-6acb4fbfb653d271/9c1fb85/mapf/src/lib.rs:17:1
   |
17 | #![feature(associated_type_bounds, type_alias_impl_trait)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This error is independent of GATs.

luca-della-vedova:
> Okay I'm noticing that `mapf` depends on a few other nightly features still...
> 
> To move the occupancy feature along I'm going to copy/paste over the `Cell` struct since it's the only thing from mapf we're actually using at the moment. I'd rather not begin requiring nightly for the site editor yet if we can avoid it.

Sounds good, will close this then, since it's not relevant anymore.

PR #94, opened 2022-11-23
Support CLI and improve main menu behavior

mxgrey:
This PR adds support for specifying a map to load from the commands line, e.g. from the root directory of this repo:

```
cargo run --release -- assets/demo_maps/office.building.yaml
```

The general format is
```
cargo run <cargo options> -- <site editor options>
```
but right now the only command line argument we only support is specifying one file to load at startup.

This PR also takes the opportunity to fix some other aspects of the main menu behavior. In the past I've noticed cases where an extra file selection dialogue pops up and gets permanently stuck. I believe this PR should fix that issue as well.

### DIFF
--- a/rmf_site_editor/Cargo.toml
+++ b/rmf_site_editor/Cargo.toml
@@ -28,6 +28,7 @@ dirs = "4.0"
 thread_local = "*"
 lyon = "1"
 thiserror = "*"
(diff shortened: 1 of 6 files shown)


codebot (approved):
Works as expected! This will be a _huge_ timesaver when quickly iterating on building generation scripts. Thank you! :tada: :bow: :bow: :bow:

PR #107, opened 2023-01-11
Prevent visual cues from showing for physical camera feeds

mxgrey:
Until now we've allowed visual cues that are meant for the user (e.g. visualizations of lanes and anchors) to appear in the physical camera preview image. That behavior is not desirable since it means the physical camera preview is incorrect.

This PR eliminates that problem using bevy's `RenderLayers` feature. The `RenderLayers` component allows us to control what entities get rendered in which camera views. The non-physical entities are put into their own render layer which only gets rendered by the user-camera.

In order to prevent the user interaction highlights from showing in the physical cameras, I needed to tweak the upstream implementation of `bevy_mod_outline`. I've [opened an issue](https://github.com/komadori/bevy_mod_outline/issues/8) to ask the maintainer to open a branch for `v0.2` so I can open a PR for the changes.

### DIFF
--- a/rmf_site_editor/Cargo.toml
+++ b/rmf_site_editor/Cargo.toml
@@ -15,7 +15,8 @@ name = "rmf_site_editor"
 bevy_egui = "0.15"
 bevy_mod_picking = "0.8"
 bevy_mod_raycast = "0.6"
-bevy_mod_outline = "0.2"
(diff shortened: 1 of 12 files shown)


mxgrey:
Upstream changes are merged and released, so I'm switching to the newly released version of `bevy_mod_outline` and merging this.

PR #108, opened 2023-01-15
Improve UI/UX for anchors

mxgrey:
Key highlights:
* anchors are smaller and less noisy
* anchors are hidden unless one of these conditions is met:
  * an element that depends on the anchor is hovered/selected
  * the cursor is close to the anchor
  * the anchor does not have any elements that depend on it
  * the editor is in a select-anchor mode
* when visible, anchors are rendered on top of everything
* anchors have an outline (white if hovered, black if not hovered)


codebot (approved):
This is awesome. It makes the world look so much nicer and less cluttered. I think those cyan hemispheres were part of the first POC as I was attempting to learn about Bevy :rofl: so they were definitely not well thought-out. Thanks for this! It's so cool to see how `RenderLayers` clean things up.

mxgrey:
> they were definitely not well thought-out

I couldn't think of any better way to do it until I recently learned about RenderLayers. We got some excellent mileage out of those medicine balls!

PR #111, opened 2023-02-13
Upgrade to bevy 0.9

luca-della-vedova:
### Implemented feature

This PR upgrades bevy from 0.8 to 0.9, it seems 0.10 is already behind the corner!

The bulk of the changelog is due to the deprecation of `spawn_bundle` / `insert_bundle` into `spawn / insert`, and the fact that resources need to be explicitly derived now.

Two main changes / regressions:

* `rmf_site_format` was changed to use `bevy::prelude` vectors instead of `glam`. I believe this might not be ideal since now the crate will depend on `bevy`. However I found a lot of very cryptic errors when mixing bevy and glam vectors when upgrading the dependency (even though from what I understand they are pretty much the same behind the scenes).
* Outline for floors is not displayed anymore, this I'm very puzzled about since there weren't too many changes in the outline file, I wonder if a change somewhere else in the stack (including in the `bevy_mod_outline` crate) might have caused it.

### DIFF
--- a/rmf_site_editor/Cargo.toml
+++ b/rmf_site_editor/Cargo.toml
@@ -12,10 +12,10 @@ path = "src/main.rs"
 name = "rmf_site_editor"
 
 [dependencies]
-bevy_egui = "0.15"
-bevy_mod_picking = "0.8"
-bevy_mod_raycast = "0.6"
-bevy_mod_outline = "0.2.5"
+bevy_egui = "0.19"
+bevy_mod_picking = "0.10"
+bevy_mod_raycast = "0.7"
+bevy_mod_outline = "0.3.2"
 smallvec = "*"
 serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.8.23"
@@ -23,7 +23,7 @@ serde_json = "1.0"
 wasm-bindgen = "0.2"
 web-sys = { version = "0.3.56", features = ["console"] }
 futures-lite = "1.12.0"
-bevy = "0.8"
+bevy = "0.9"
 dirs = "4.0"
 thread_local = "*"
 lyon = "1"
@@ -34,7 +34,7 @@ bitfield = "*"
 
 # only enable the 'dynamic' feature if we're not building for web or windows
 [target.'cfg(all(not(target_arch = "wasm32"), not(target_os = "windows")))'.dependencies]
-bevy = { version = "0.8", features = ["dynamic"] }
+bevy = { version = "0.9", features = ["dynamic"] }
 surf = { version = "2.3" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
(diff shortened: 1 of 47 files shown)


mxgrey:
> However I found a lot of very cryptic errors when mixing bevy and glam vectors when upgrading the dependency (even though from what I understand they are pretty much the same behind the scenes).

I think this is most likely caused by bevy changing which version of glam it's using. The fix should be simple: change `rmf_site_format`'s [glam dependency version](https://github.com/open-rmf/rmf_site/blob/e5bab612857d9d49885e365b8a3ae1e9a76a7d9a/rmf_site_format/Cargo.toml#L17) to match whichever version of glam Bevy 0.9 is using. I'd like us to try very hard to make a simple glam dependency work before we resort to adding all of bevy as a dependency for `rmf_site_format` because the difference in dependency footprint is enormous.

> Outline for floors is not displayed anymore

I think this is related to the fact that `bevy_mod_outline` v0.3 introduced a choice between "flat" and "real" depth for outlines. In v0.2 we only used "flat", and I think that's preferable for our use case.

In [672868b](https://github.com/open-rmf/rmf_site/pull/111/commits/672868b786f2bde548a903af71a158c3ee5d6611) I've updated our code to use the flat appearance, but I found that it won't get updated correctly without the changes in [this upstream PR](https://github.com/komadori/bevy_mod_outline/pull/10).

But even with that upstream fix there are some undesirable behavior changes. See in this screenshot how the floor's bottom outline is missing, and the outline on the right-hand side of the floor is much thinner than it should be:

Screenshot from 2023-02-13 19-19-27

I'm not sure what would have caused those changes. If I have some time I might look into the new shaders being used in `bevy_mod_outline` and compare them to the old shaders to see if I can spot where their behaviors would have diverged for the flat outline case.

luca-della-vedova:
> I think this is most likely caused by bevy changing which version of glam it's using. The fix should be simple: change `rmf_site_format`'s [glam dependency version](https://github.com/open-rmf/rmf_site/blob/e5bab612857d9d49885e365b8a3ae1e9a76a7d9a/rmf_site_format/Cargo.toml#L17) to match whichever version of glam Bevy 0.9 is using. I'd like us to try very hard to make a simple glam dependency work before we resort to adding all of bevy as a dependency for `rmf_site_format` because the difference in dependency footprint is enormous.
> 

Ah yes missed that! Reverted https://github.com/open-rmf/rmf_site/pull/111/commits/93238ce413ed88575be1612bd6944d994cc3e9b4 Do you think it's OK to get this in and base on Bevy 0.9 moving forward or would you rather look at the regression first?

mxgrey:
> Do you think it's OK to get this in and base on Bevy 0.9 moving forward or would you rather look at the regression first?

I'm not sure when I'll have time to look at the line issue, so there's not much point in holding things up on that. I think the difference isn't bad enough to hold up progress on this migration.

However, the upstream issue with updating the outlines is not okay, and I'm not sure when my PR for that will be merged, so I've updated this branch to use my fork of `bevy_mod_outline` for now.

mxgrey (approved):

luca-della-vedova:
Upstream `bevy_mod_outline` PR has been merged and re-released, I updated the tag to match the minimum required

PR #116, opened 2023-03-06
Add layering system and improve aesthetics

mxgrey:
This PR introduces a system that allows users to specify which order to render different "layers". "Layers" includes anything objects that are meant to be rendered flat on the ground, such as drawings, floors, and nav graphs. Here's an overview of the new features:
* Control the heights of same-type layers relative to each other
* Toggle the visibility of floor elements, including a global toggle
  * By default floors are opaque
  * When a drawing is present, the global default for floors changes to semi-transparent
* Fix the outline behavior of floors so that a uniform outline appears around the entire floor segment when hovered/selected

The improvements are most obvious in worlds that have many layers. E.g. the airport demo:

### Before
See the ugly z-fighting between the green and red lanes
airport_before

### After
All of the rendering is clean, no z-fighting anywhere
airport_after


luca-della-vedova (commented):
Few minor points as I go through the testing:

* It seems that assets are included in both .png and .svg format, is this intended or should we remove one?
* I also noticed that drawings are not selectable anymore by clicking on them but only through the layer -> select button, unlike floors that can be selected by clicking, same as whether this is intended or not

At least on my machine, the side bar's default width starts to be way too small to fit all the data, as in picture below:

image

Two of the side buttons are not displayed at all because of the narrowness, is it the same for you or is it a display resolution issue?

I also noticed that some outlines appear thicker than others, it almost seems that there are two outlines being drawn (i.e. in the bottom right in the picture below).
I noticed a comment pointing to an issue with the outline crate but it seems to be a different issue

image

mxgrey:
> It seems that assets are included in both .png and .svg format, is this intended or should we remove one?

This is intentional because the `.svg` is the original source that I created with inkscape, and I want to keep that in case we want to modify it in the future. Meanwhile the `.png` is the actual asset that gets baked into the application. Theoretically we could generate the `.png` as part of the build process using a `build.rs` script, but I think that would be way more involved than it's worth.

> I also noticed that drawings are not selectable anymore by clicking on them but only through the layer -> select button, unlike floors that can be selected by clicking, same as whether this is intended or not

Yes, this is intentional, although I'm open to other ideas on how to handle this. I don't think it will generally be too important to select drawings since users won't usually interact with them much once they're settled into place and you've started to draw floors on top of them. Instead I think it's much more likely that users will want to have easy access to selecting and editing floors.

One option might be to have a shortcut key that tells the cursor to only highlight drawings. But I'm skeptical of the usefulness for all the reasons I mention above.

> At least on my machine, the side bar's default width starts to be way too small to fit all the data

I've gotten accustomed to dragging it out further when needed, but I agree that it's not ideal. I've tried experimenting in the past with automatically expanding the side bar, but I concluded that egui just doesn't have a good way of doing that. I think we'll need to settle for less-than-ideal UIs until we have time to dig into proper UI frameworks like [kayak_ui](https://docs.rs/kayak_ui/latest/kayak_ui/) or [bevy_dioxus](https://docs.rs/bevy_dioxus/latest/bevy_dioxus/).

But anyway, feel free to play with different default sizes for the side panel and suggest a size that seems good to you. You can add a `.default_width(x)` to [this line](https://github.com/open-rmf/rmf_site/blob/1b695240e795a88833fcd45e61fccaca795f2c20/rmf_site_editor/src/widgets/mod.rs#L157).

> I also noticed that some outlines appear thicker than others, it almost seems that there are two outlines being drawn (i.e. in the bottom right in the picture below).
> I noticed a comment pointing to an issue with the outline crate but it seems to be a different issue

This is tangentially related to the issue that's linked in the PR. The inconsistencies and choppiness that you're seeing are because we're using the `SetOutlineDepth::Real` setting for the outline which attempts to blend the outline into the actual geometry. What we really want for everything is `SetOutlineDepth::Flat`, but the [linked issue](https://github.com/komadori/bevy_mod_outline/issues/14) shows a bug that can occur with the flat setting.

Since that bug has a high probability of being triggered for large outlines like floors and walls, we're temporarily using `Real`, which doesn't experience the bug but instead has these subtle quality losses. We'll switch back to `Flat` as soon as a fix for the bug is released.

luca-della-vedova:
> This is tangentially related to the issue that's linked in the PR. The inconsistencies and choppiness that you're seeing are because we're using the `SetOutlineDepth::Real` setting for the outline which attempts to blend the outline into the neighboring geometry. What we really want for everything is `SetOutlineDepth::Flat`, but the [linked issue](https://github.com/komadori/bevy_mod_outline/issues/14) shows a bug that can occur with the flat setting.
> 
> Since that bug has a high probability of being triggered for large outlines like floors and walls, we're temporarily using `Real` for those high-risk types of objects because `Real` doesn't experience the bug. However, `Real` has these subtle aesthetic losses. We'll switch back to `Flat` as soon as a fix for the bug is released.

It seems a fix was released and the module was re-released, however upon testing the new version with a `cargo update` outlines for floors / walls are not visualized anymore, I wonder if there is anything we need to change on our side or if it's another issue with the upstream crate?

mxgrey:
> with a cargo update outlines for floors / walls are not visualized anymore

Indeed it seems 0.3.4 caused the outlines to stop working for the `Real` setting. I've changed all of our outlines back to the `Flat` setting since that's working with the latest update, and it's what we want anyway. If you run the airport demo again you should see that the outlines look okay now.

I've reported the issue about `Real` upstream. It's not a blocker for us, though.

luca-della-vedova:
Generally looks good to me.
I was looking at the code for saving / loading and I couldn't see any addition to make sure that `RankingsInLevel` are serialized, but I noticed that rankings for navigation graphs instead are serialized.
Should there be a new query to iterate through to populate rankings in level to make sure that the information is not lost when the site is saved, closed and reopened?

mxgrey:
> Should there be a new query to iterate through to populate rankings in level to make sure that the information is not lost when the site is saved, closed and reopened?

Great catch! I absolutely meant to do that but it slipped my mind. Done by [45a4b5c](https://github.com/open-rmf/rmf_site/pull/116/commits/45a4b5c30a60d767ecf8ac1fb4ef1c37b78966ae)

luca-della-vedova (approved):

Issue #71, opened 2022-10-31
Status/Error Console


mxgrey:
Currently when an error happens or we want to convey some detailed information to the user, we have to print it to the terminal. This is often not visible for users since the terminal is often hidden behind the GUI application or possibly not visible at all (e.g. on the web browser). We should instead have a plugin for logging helpful information, status updates, and error messages to a GUI element that is more visible for users.

Visually it can be a bar in the [bottom panel](https://docs.rs/egui/latest/egui/containers/panel/struct.TopBottomPanel.html#method.bottom) which only shows the latest message message. The bar would have a button to expand the bottom panel into a scrollable console that shows all the messages that have been received.

We should have four different categories of messages:
* `Hint`: Displayed in the status bar but not saved in the log history, meant for giving advice to the user.
* `Status`: Notifies that a state-changing event has occurred. Saved in the log history.
* `Warning`: Notifies that the user might be trying to do something that isn't allowed. Saved in the log history.
* `Error`: Notifies that something has gone wrong in the program. These messages should be reported to Site Editor developers as they indicate bugs in the programming. Saved in the log history.

Additional features:
* Each category should be displayed differently, e.g. different text color or font settings
* When looking at the full history, users should be able to filter the messages based on the category
* Clicking on a message should copy the text to the clipboard
* The full console should have an option for copying the whole history to the clipboard
* The full console should have an option for copying the whole history of a specific category to the clipboard (especially needed for the `Error` category to help users report bugs).

Performance considerations:
* When rendering the full console, be mindful of how long the log history is. A very long log history could cause egui to struggle to render it. Consider limiting the number of log entries that are put into the history and having a `See more...` button at the bottom of the scrollable area.

PR #121, opened 2023-04-17
Logging console

xiyuoh:
This PR targets #71 to add a logging console panel in the site editor that collects printouts that would usually only appear in the terminal.
- Usually only shows the latest log entry, but expandable to reveal full log history
- Four categories of messages that can be filtered and identified based on colour
- Able to copy a single entry or the entire log history to the clipboard by clicking on the message/`Copy Log History` button
- Slider to limit the default number of entries displayed, with accompanying `See More`/`See Less` buttons to further customise number of entries shown

### Examples of use (updated)
Use `tracing`'s [shorthand macros](https://docs.rs/tracing/latest/tracing/#shorthand-macros).

### To test
Examples have been implemented in keyboard.rs with the `D` and `Del` keys.

### To do
Switch the `println!`s across site editor to this logger (done)


mxgrey (changes requested):
The console is looking really good!

I've left some inline comments, many of them are just style nitpicks to make the code more idiomatically "Rusty", but the one about tweaking the `LogCategory` enum is somewhat important.

I also have a few high level comments for improving the implementation:

### Do not use a `ResMut` for creating new log entries.

Every system function that requests a `ResMut<T>` for the same `T` will have to run in sequence instead of parallel. This is to make sure they don't make simultaneous conflicting changes to the `T`. Since we want the logging feature to be widely accessible and not have any negative impacts on performance, it would be really bad if we lose parallelism just by introducing logging to a system.

Instead we should take advantage of Bevy's Events feature. Any number of systems can request an `EventWriter<T>` for the same `T` and run in parallel. The written events will be collected during a command flush, and by the next iteration of the system schedule they will be read by any system with an `EventReader<T>`. We just need to add a system that reads logging events and collects them into the `LogHistory` resource.

For some added ergonomics we can define a custom "system parameter":

```
#[derive(SystemParam)]
struct Logger<'w, 's> {
    writer: EventWriter<'w, 's, Log>,
}

impl<'w, 's> Logger<'w, 's> {
    pub fn status(&mut self, message: impl Into<String>) {
        self.writer.send(Log {
            category: LogCategory::Status,
            message: message.into(),
        });
    }

    /* ... TODO: Other log categories ... */
}
```

Then users can request a `Logger` and call `logger.status("My Status");`

### To allow console messages from async functions, have them return a tuple of `(T, Vec<Log>)`

where `T` is whatever the async function's current return type is. After the async function is finished, we can move its log entries into to the Logger.

This should work fine for async functions that have a very short run time. In the future we may need to consider using [channels](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) so that long-running async functions can give status updates while they're still in progress, but we don't have any long-running async functions for now, so this isn't a priority.

luca-della-vedova:
I don't know if this was looked at, but should we consolidate on using `bevy_log` throughout the codebase and find a way to redirect its output to the logging console?

I can see arguments both ways (showing bevy logs would bring a fair bit of noise from the engine, but we could probably avoid reimplementing all the log filtering, severity levels, and would also not lose all the bevy logs once we run in a web app).

xiyuoh:
@luca-della-vedova I looked at it briefly, at the time it felt easier to implement the logger from scratch but i'm happy to give it a try again if its merits are worth it

> and would also not lose all the bevy logs once we run in a web app

Not sure if this is what you meant, but the current widget also prints logs to the terminal

luca-della-vedova:
> Not sure if this is what you meant, but the current widget also prints logs to the terminal

The logs I meant are the bevy specific ones, i.e. these:

image

We don't usually look at them much right now but it could be an idea to migrate our logging to use these instead, so we can get a single logging method and get their other useful features (color coding, pointing to which file generated the log).
Again I also see merits in keeping them fully decoupled so we don't have to deal with some noise coming from other bevy logs, so open to anything

mxgrey:
> Not sure if this is what you meant, but the current widget also prints logs to the terminal

What Luca means is that when we run the site editor in a web browser right now, we're not able to see any terminal output **at all** (there might be some workarounds, but it's very difficult). So anything that's being printed to the terminal is simply lost, including a lot of useful debug messages from inside of Bevy.

I definitely like the idea of wrangling the bevy log data into the console, but I would suggest doing that by adding a category like `Developer` or `Debug` which is hidden by default. I don't want every type of message to include info like the file and line of code where it originated. My vision for this console is to target end users more so than us developers, providing them with helpful context and a record of significant actions that they've taken.

But if we have a system that intercepts the bevy log messages and wraps each one into a `Dev`/`Debug` console log entry, and then users can toggle visibility for those on and off, I think that would be an awesome addition.

xiyuoh:
Gotcha! I'll dig a little deeper into the log plugin and see how to make it work. Thank you both for the explanation 😁

mxgrey:
Regarding the addition of crossbeam, it seems like the current version of [`std::sync::mpsc` is simply a copy of the `crossbeam-channel` in the standard library](https://users.rust-lang.org/t/for-beginners-using-channels-sync-mpsc-or-crossbeam/86547/4), so I would recommend just using the `std` implementation to avoid an unnecessary third party dependency unless there's a specific reason that I'm overlooking.

xiyuoh:
@mxgrey When passing the `mpsc::Receiver` into `LogHistory`, there was an error that says the `Receiver` object cannot be safely shared between threads. With crossbeam the `Receiver` can be cloned and passed along without problems. Seems like this trait is not available in `std::sync::mpsc`

mxgrey:
I see, then I guess crossbeam has made some progress which hasn't landed in `std` yet. Thanks for explaining!

xiyuoh:
PR is ready for review! With the following updates:
- Although Bevy's LogPlugin already uses the `tracing` and `tracing-subscriber` crates to collect events in the app, there doesn't seem to be a way to export these logs. Instead, the LogPlugin is now disabled, and a Subscriber and a Visitor are implemented in the console module to send logs over to the `LogHistory`.
- Switched the EventWriter implementation to using channels (Sender/Receiver) as suggested so that we can send messages across threads to the `LogHistory`
- Added a `Bevy` category that collects logs from Bevy itself. **Note:** current approach to differentiate between Bevy and RMF Site messages is to check whether the log's `event.metadata.target()` (String) contains "bevy". There might be a better way that i haven't thought of
- Building on top of the review comments about `LogCategory`, got rid of the `All` category and instead stored a HashMap in `LogHistory` to track which categories have their checkboxes toggled to true. Swapping out the drop-down menu for checkboxes allows for multiple categories to be selected at a time.
- Updated struct and function names as suggested

luca-della-vedova:
I believe this PR breaks wasm since opening the demo map freezes on the browser with this PR but works on `main`. I'll investigate further

xiyuoh:
@luca-della-vedova seems like the wasm issue was due to the tracing subscriber's formatting layer. Should be fixed now, the layer will only be added for non-wasm to preserve Bevy's output formatting in the terminal.

rmf_site_editor/src/widgets/console.rs
```diff
@@ -0,0 +1,471 @@
...
+    receiver: Receiver<Log>,
+}
+
+impl Default for LogHistory {
+    fn default() -> Self {
+        let mut filter_hashmap = HashMap::new();
+        filter_hashmap.insert(LogCategory::Status, true);
+        filter_hashmap.insert(LogCategory::Warning, true);
+        filter_hashmap.insert(LogCategory::Error, true);
+        filter_hashmap.insert(LogCategory::Bevy, false);
```
luca-della-vedova, line 76: I noticed that this filter only gets applied to the console that is expanded, not to the one that prints messages at every iteration, which means `[BEVY]` messages are still visibile by default (and spam that console a bit at the moment).
Is this intended?

rmf_site_editor/src/lib.rs
```diff
@@ -146,6 +148,8 @@ pub fn run(command_line_args: Vec<String>) {
     {
         app.add_plugins(
             DefaultPlugins
+                .build()
+                .disable::<LogPlugin>()
```
luca-della-vedova, line 152: This is a bit of an unfortunate necessity since the snippet below:

```
                .set(LogPlugin {
                    filter: "bevy_asset=error,wgpu=error".to_string(),
                    ..default()
                })
```

Allows us to filter the log based on crates and severity, since the plugin was disabled this is not possible anymore.

I guess we can either:
* If this can be reverted, keep the logging plugin and its filtering (I assume not possible based on what you said)
* If this cannot be reverted, either remove the snippet below since it doesn't do anything now or try to get some similar configuration into the current logging plugin.

rmf_site_editor/src/widgets/console.rs
```diff
@@ -0,0 +1,471 @@
...
+                    );
+                    ui.checkbox(
+                        self.events
+                            .display
+                            .log_history
+                            .category_present_mut(LogCategory::Bevy),
+                        "Bevy",
+                    );
+                    // Copy full log history to clipboard
+                    if ui.button("Copy Log History").clicked() {
```
luca-della-vedova, line 383: Not sure this works, I tried and all I could get is exporting of only the first message in the log

rmf_site_editor/src/widgets/console.rs
```diff
@@ -0,0 +1,471 @@
...
+        &mut self.display_limit
+    }
+
+    pub fn receive_log(&mut self) {
+        match self.receiver.try_recv() {
+            Ok(msg) => self.append_log(msg),
+            Err(TryRecvError::Disconnected) => println!("Unable to receive log: Disconnected"),
+            Err(TryRecvError::Empty) => (),
+        }
+    }
```
luca-della-vedova, line 183: It seems the log is updated once per frame with one message, which means that if, let' say, 30 messages were received in one frame and we were running at 30 FPS, the logs would be delayed by one second.

What do you think about, instead of consuming just one entry, we consume all the events that are currently available?
The main drawback would be that if we receive 10 events in one frame only the last one would be readable in the top widget, but on the other hand we wouldn't have these long delays if there happen to be a lot of events (which mostly happen at startups).

luca-della-vedova (commented):
Thanks for the wasm fix! Works for me now.
I tested the code more from a functionality point of view, left comments below:

rmf_site_editor/src/lib.rs
luca-della-vedova: Just saw the filter is available in the `console.rs` file, then we can just copy the updated filter below into the `default_filter` variable and remove it from here

rmf_site_editor/src/widgets/console.rs
xiyuoh: You're right, that should be the case. I've modified `receive_log()` to receive all available messages in the channel in [5a040a9](https://github.com/open-rmf/rmf_site/pull/121/commits/5a040a9e2ab9bce509df3798a13059d8ca948d6e)

rmf_site_editor/src/widgets/console.rs
xiyuoh: Ah, i think this only works on locally built editor. I could copy a single/full log and paste it into any open-ended field within the web app (example the `name` field of the Level), but not outside of the browser. Will look into this further

rmf_site_editor/src/widgets/console.rs
xiyuoh: Yup, currently the bar above the expandable console shows that latest log that comes through regardless of whether its category is being filtered or not. Do you think it would be necessary to have a button that disables Bevy logs being shown in the bar? Or should they not even be shown at all?

luca-della-vedova: :man_shrugging:  Personally I found it a bit confusing that the `BEVY` level is selected to be filtered out but was still displayed in one of the two logs, I would have expected the two displays to be coherent

xiyuoh: [12bdc1d](https://github.com/open-rmf/rmf_site/pull/121/commits/12bdc1d07916a14bf8c099abe330b02f5d16c151)

rmf_site_editor/src/widgets/console.rs
luca-della-vedova: Even in the local built editor I can only get it to copy the first message, I tried to click on a few messages as well but it doesn't seem to work

xiyuoh: Oh that's really weird, it seems to work for me. Also works on VM too. Let me try with another machine

xiyuoh: Could you try a `rustup update`? It wasn't working on another machine and after an update it works now

luca-della-vedova: We looked into it a bit more and it seems to be a clipboard issue and depends on the software that is pasted on, a few (non exhaustive) examples:

* Chrome works, but firefox doesn't
* VS code works, but vim, thunar and gedit don't

I suspect this might be a bit too hairy to solve

xiyuoh: As Luca pointed out there was a `\0` hidden in the tracing layer. Removed in [3ccac6d](https://github.com/open-rmf/rmf_site/pull/121/commits/3ccac6d0bb69be22becda93643c8e151626d23c3) and it should copy properly now.

luca-della-vedova (approved):
Thanks for iterating on this! Looks good to me

mxgrey (approved):
This is such a great feature! I just push a little cleanup:
* Simplified the UI logic
* Refactored the log history into its own plugin with a separate system for collecting logs
* Now users can push logs as bevy events instead of using the trace mechanism. This unlocks the `HINT` category for us.

PR #123, opened 2023-05-11
Add support for spawning robots in locations

luca-della-vedova:
### Implemented feature

Spawn a robot visual in locations with a `spawn_robot_name` and `spawn_robot_type` property

The implementation is quite naive and currently just spawns a single static model in spawning locations, there is a TODO in place for future plans when robot behavior is added to the site editor, including making it non static and / or adding other components to distinguish models from robots.

While working on it I noticed that querying for whether a Location has a `SpawnRobot` tag is not terribly ergonomic since all tags are contained in a `LocationTags` component that is a `Vec<LocationTag>`, so each system dealing with location behavior will need to query for components with `LocationTags`, iterate over the whole vector and verify whether the location has the tag it needs (as is done here for `SpawnRobot`).

On the other hand, if each tag was a different component it would be possible to filter relevant entities at the query level, improving both performance and code simplicity.

Still, I kept this out of this PR since it would be quite a large refactor and didn't want to get into it in case it's not a good idea.

### DIFF
--- a/rmf_site_editor/src/site/location.rs
+++ b/rmf_site_editor/src/site/location.rs
@@ -22,6 +22,9 @@ use bevy::prelude::*;
 // experience z-fighting.
 const LOCATION_LAYER_HEIGHT: f32 = LANE_LAYER_LIMIT + SELECTED_LANE_OFFSET / 2.0;
 
+#[derive(Component, Debug, Clone)]
+pub struct LocationRobotModel(Entity);
+
 // TODO(MXG): Refactor this implementation with should_display_lane using traits and generics
 fn should_display_point(
     point: &Point<Entity>,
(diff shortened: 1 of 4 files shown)


luca-della-vedova:
Note that there is currently a panic when removing then adding a spawn robot again. This panic is at the `model.rs` level and also appears when changing the `AssetSource` of models, I haven't figure that out yet but it is not strictly introduced in this PR.

luca-della-vedova:
Small followup, panic is related to #145 so it's about the AABB plugin

luca-della-vedova:
Closing for now. There will be much more work needed to support robots as shown in #153, while this PR only implements naive rendering

PR #124, opened 2023-05-11
[WIP] Undo Moves

arjo129:
### Implemented feature

This PR adds an undo feature to the editor. Currently it only undos moves.

undo

### Implementation description

Currently, this uses an `enum` to dispatch various types of undo commands. In the future the hope is to use some type of trait system to provide a scalable dispatch.

PR #125, opened 2023-05-15
Undo move functionality

arjo129:
Supersedes #124.

### Implemented feature

This PR adds an undo feature to the editor. Currently it only reverts moves.

undo

Currently, this uses an `enum` to dispatch various types of undo commands. In the future the hope is to use some type of trait system to provide a scalable dispatch.

### DIFF
--- a/rmf_site_editor/src/interaction/anchor.rs
+++ b/rmf_site_editor/src/interaction/anchor.rs
@@ -92,12 +92,37 @@ pub fn remove_interaction_for_subordinate_anchors(
 }
 
 pub fn move_anchor(
-    mut anchors: Query<&mut Anchor, Without<Subordinate>>,
+    mut anchors: Query<(&mut Anchor, &Parent), Without<Subordinate>>,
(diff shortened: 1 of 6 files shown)


mxgrey:
Thanks for getting this feature kicked off.

I understand the appeal of using a simple enum to start, but I think we can leapfrog straight to the generalized solution.

I recommend introducing some traits like this:

```
trait Undoable {
    fn undo(self, world: &mut World) -> Box<Redoable>;
}

trait Redoable {
    fn redo(self, world: &mut World) -> Box<Undoable>;
}
```

We create simple structs for each type of action that we want to be able to undo/redo and implement these traits for those structs.

Then we can have a resource containing buffers of `Vec<Undoable>` for the undo stack and `Vec<Redoable>` for the redo stack. The redo stack gets cleared each time any undoable action is performed.

I don't think we need to use `VecDeque` since pushing/popping to a `Vec` is sufficient for a stack.

PR #126, opened 2023-05-15
POC fix for legacy lifts

luca-della-vedova:
### Fixed bug

When importing projects containing lifts with rectangular shapes, things can go quite wrong, from swapped widths and depths to failure in importing the project.
This PR is little more than a bug report, it's a "hacky fix that seems to work", there might be a better fix but I'm not familiar with how lifts are stored and parsed in the site editor.

### Fix applied

I just swapped widths and depths in the legacy lift importing code path, not sure how the coordinate system is supposed to be but this seems to fix all the scenarios I could find.

Test it!

* Create a world with a rectangular lift, for example change [this line](https://github.com/open-rmf/rmf_demos/blob/3f29a858ae64f2702696c5b1e5481c969c8f08a4/rmf_demos_maps/maps/hotel/hotel.building.yaml#L1046) to a large number like 10.0
* Open the map in traffic editor to see what it looks like.
* Try to open the map in site editor `main` branch, import should  fail: `Failed converting to site InvalidLiftCabinDoorPlacement { lift: "Lift1", door: "lift1_door" }`
* Try the PR, import should be successful and the lift should match its traffic editor rendering

### DIFF
--- a/rmf_site_format/src/legacy/lift.rs
+++ b/rmf_site_format/src/legacy/lift.rs
@@ -45,8 +45,8 @@ impl Lift {
         // TODO(MXG): Rewrite this with glam now that we've accepted it as a dependency
         let x = self.x as f32;
         let y = self.y as f32;
-        let d = self.depth as f32 / 2.0;
-        let w = self.width as f32 / 2.0;
+        let d = self.width as f32 / 2.0;
+        let w = self.depth as f32 / 2.0;
         let theta = self.yaw as f32;
         let rotate = |x, y| {
             (
@@ -95,8 +95,8 @@ impl Lift {
 
             let dx = door.x as f32;
             let dy = door.y as f32;
-            let half_width = self.width as f32 / 2.0;
-            let half_depth = self.depth as f32 / 2.0;
+            let half_width = self.depth as f32 / 2.0;
+            let half_depth = self.width as f32 / 2.0;
 
             let cabin_face = if dx.abs() < 1e-3 {
                 // Very small x value means the door must be on the left or right face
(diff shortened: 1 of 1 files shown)


arjo129 (changes requested):
While these changes work for me, reading the code is really confusing. Consider the following options:
* Add a comment as to why we are flipping the width and the depth before
* Flip the usage of `d` and `w` in functions calling them.

arjo129 (changes requested):
While these changes work for me, reading the code is really confusing. Consider the following options:
* Add a comment as to why we are flipping the width and the depth before
* Flip the usage of d and w where they are used in function calls.

luca-della-vedova:
Yea I have no hope of getting this merged, it's really meant to be just a bit more of a bug report, a "bug report with a way to work around it that hopefully sheds some light into what a proper fix might be".
Will change it to draft to reflect this

mxgrey (approved):
Thanks for noticing and reporting this. You're very correct, the width and depth should be flipped when we do the conversion. That's actually the correct fix, and the reason is because of a change in coordinate axis conventions.

The legacy format used a simple y-forward, x-sideways convention for depth and width, probably because when the lift has zero rotation then the x-axis "feels" like the dimension that should be called "width". However for the site editor I decide to switch to the robotics convention of the local x axis being forward/backward and the local y axis being side-to-side, because I believe that will be more consistent with conventions that we'll use elsewhere, and I believe it will make more sense for lifts that have some rotation.

I've simply added a note about this into the PR because it's admittedly confusing. No other action should be needed to fix the problem.

PR #127, opened 2023-05-16
Hotfix for outline issue in wasm

luca-della-vedova:
### Fixed bug

There seems to be an issue with outline rendering in wasm builds, as documented in https://github.com/komadori/bevy_mod_outline/issues/19.

### Fix applied

This PR circumvents the issue by temporarily switching to Real outlines for wasm builds.
It's already been deployed for testing and can be verified [on the browser](https://open-rmf.github.io/rmf_site/)

### DIFF
--- a/rmf_site_editor/src/interaction/outline.rs
+++ b/rmf_site_editor/src/interaction/outline.rs
@@ -81,17 +81,20 @@ impl OutlineVisualization {
         }
     }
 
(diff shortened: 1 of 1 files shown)


luca-della-vedova:
Upstream issue is fixed but only released on bevy 0.10. This fix will not be needed anymore once we upgrade.

mxgrey (approved):

PR #131, opened 2023-05-22
Fix panic with new workspaces

### LINKED ISSUE #130: Clicking on Create->Lane in an empty world causes editor to panic.
### Description of the bug
Clicking on `Create->Lane` in the editor when no anchor is selected causes the editor to panic. The offending line is here:
https://github.com/open-rmf/rmf_site/blob/f796e44ce5a1d13a2878b5cc2cd7d93de0bbee0b/rmf_site_editor/src/interaction/anchor.rs#L44

1. File -> new
2. Click on create->lane.
3. Click on the empty space.

### Expected behavior
A panic feels extreme. The system should gracefully do nothing or give a warning at this point

### Actual behavior
The application panics.

luca-della-vedova:
### Fixed bug

Fixes #130.

### Fix applied

The fix is reasonably straightforward, just change the `Default` implementation of sites to have a new empty level. I'm not 100% sure putting a site ID of 0 is sound, and whether this would create any issue if we had multiple workspaces open

### DIFF
--- a/rmf_site_editor/src/workspace.rs
+++ b/rmf_site_editor/src/workspace.rs
@@ -17,6 +17,7 @@
 
 use bevy::{prelude::*, tasks::AsyncComputeTaskPool};
 use rfd::AsyncFileDialog;
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 
 use crate::interaction::InteractionState;
@@ -24,7 +25,7 @@ use crate::site::LoadSite;
 use crate::workcell::LoadWorkcell;
 use crate::AppState;
 use rmf_site_format::legacy::building_map::BuildingMap;
-use rmf_site_format::{Site, SiteProperties, Workcell};
+use rmf_site_format::{Level, Site, SiteProperties, Workcell};
 
 use crossbeam_channel::{Receiver, Sender};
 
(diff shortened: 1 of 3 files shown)


mxgrey:
I don't think this is the way we should fix the issue. `rmf_site_format` is meant to be used by more than the editor GUI. It's also meant for use by scripts or other programs which may be surprised when a default initialization of a site gets automatically filled with exactly one level.

If anything, I think we should modify [this code](https://github.com/open-rmf/rmf_site/blob/f796e44ce5a1d13a2878b5cc2cd7d93de0bbee0b/rmf_site_editor/src/workspace.rs#L141) to create a default (empty) site and add one level before we dispatch it to `load_site`.

luca-della-vedova:
> I don't think this is the way we should fix the issue. `rmf_site_format` is meant to be used by more than the editor GUI. It's also meant for use by scripts or other programs which may be surprised when a default initialization of a site gets automatically filled with exactly one level.
> 
> If anything, I think we should modify [this code](https://github.com/open-rmf/rmf_site/blob/f796e44ce5a1d13a2878b5cc2cd7d93de0bbee0b/rmf_site_editor/src/workspace.rs#L141) to create a default (empty) site and add one level before we dispatch it to `load_site`.

Alright! Reverted the original fix and changed at the event sending level https://github.com/open-rmf/rmf_site/pull/131/commits/2be086317ffc06c95109c6784fa3523fc3462dbc
Added a `Default` derive for `Level` as well, all the fields implement default anyway so that helps reducing boilerplate

arjo129 (approved):
Works for me!

mxgrey (approved):

Issue #117, opened 2023-03-16
Measurements and Fiducials should be children of Drawings


mxgrey:
In the old traffic editor, each Level had one Drawing, so there was no distinction between whether a Measurement/Fiducial belongs to a Level or to a Drawing.

Currently Measurements/Fiducials are parsed in as children of the Level, but this no longer makes sense, especially since
1. We now represent Measurements/Fiducials in real-world Cartesian space
2. Each Level is allowed to have multiple Drawings

We should make the following changes:
1. Measurement/Fiducial endpoints are represented in pixel-space
2. Each Measurement/Fiducial is the child of a Drawing
3. We have an "Edit Drawing" mode to add, remove, and modify the Measurements/Fiducials on a single Drawing

Then we can provide a UI that provides a button to enter the Edit Drawing mode, and another button to ask a Drawing to be automatically scaled and aligned according to its Measurements and Fiducials.

PR #132, opened 2023-05-25
WIP: Drawing editor

luca-della-vedova:
### Implemented feature

Addresses #117.
Detailed overview once implementation gets more finalized. 

TODO


luca-della-vedova:
There is currently a different user experience for matching fiducials between different levels (for inter-level alignment) and different drawings on the same level (for robot - main drawing alignment).

Inter level uses fiducial names as in traffic editor, while intra level are matched by drawing constraints, also like traffic editor. Drawing of constraints is quite tricky and required a fair bit of additional logic in `select_anchor` to make sure that they can only be drawn between different drawings. If we adopted the same logic based on fiducial names the code path there could be greatly simplified, and we would also be able to remove the need to serialize constraints since they are auto-determined, but not sure that is the best way forward (drawing is still quite useful visually).

mxgrey (approved):

audrow:
From using `git bisect` it seems like this PR somehow removes the "Export from URDF" option for workcells, or maybe I'm missing something. @luca-della-vedova, any thoughts on why this might be?

mxgrey:
There were a lot of merge conflicts to deal with by the time this PR was being merged. Most likely it got wiped out while I was attempting to resolve conflicts.

PR #135, opened 2023-05-26
Adds a setting for default lane size

arjo129:
This PR adds a setting for default lane size. It can be found by going to `Edit -> Preferences...`. We need to find a way to persist this. Note this setting does not change per-lane widths as those should be handled differently.

### DIFF
--- a/rmf_site_editor/src/widgets/preferences.rs
+++ b/rmf_site_editor/src/widgets/preferences.rs
@@ -3,35 +3,10 @@ use bevy_egui::{
     egui::{self, DragValue, Slider},
     EguiContext,
 };
-use rmf_site_format::{AnchorParams, AssociatedGraphs, Edge};
+use rmf_site_format::{AnchorParams, AssociatedGraphs, Edge, SiteProperties, Preferences};
 
 use crate::site::{should_display_lane, update_lane_visuals, LaneEnds, LaneSegments, SiteAssets};
 pub struct PreferenceEvent;
-
-const LANE_DEFAULT_SIZE: f32 = 0.5;
-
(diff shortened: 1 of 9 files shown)


rmf_site_editor/src/widgets/preferences.rs
```diff
@@ -0,0 +1,106 @@
...
+            props.preferences = Some(Default::default());
+            println!("Setting default value");
+        }
+
+        if let Some(preferences) = props.preferences.as_mut() {
+            egui::Window::new("Preferences").show(egui_context.ctx_mut(), |ui| {
+                ui.horizontal(|ui| {
+                    ui.label("Default Lane Width: ");
+                    let lane_width_entry = DragValue::new(&mut preferences.default_lane_width)
+                        .clamp_range(0.0..=10000.0);
```
luca-della-vedova, line 50: `0.0` is a bit of a risky lower bound that would make it completely invisible, I'd suggest making it something that is not completely invisible but still covers the lowest common case (maybe 0.1 m?)

luca-della-vedova (changes requested):
A good start on the feature, not sure whether it's ok for the long term but probed a bit around the implementation and found a few issues that should be addressed if we want to get this in.

It seems that when the default lane width is set and a new lane is created a fair bit of glitches appear, as in the video below:


Specifically the beginning and end of each lane go back to the original width. Measurements are also scaled

arjo129:
Will fix these however in another conversation with @mxgrey he mentioned the following crate: https://github.com/ForesightMiningSoftwareCorporation/bevy_polyline

This removes the need for this feature. Will probably use this instead.

arjo129:
Closing in favor of #152

PR #136, opened 2023-06-08
Add support for wall and floor textures

luca-della-vedova:
### Implemented feature

Targets #132, will retarget `main` once target branch is merged.
This PR updates the `texture` structure and implements parsing and live updates for texture parameters. For now only texture source, height, width and alpha are implemented. Texture rotation is not implemented yet since the UV calculation is not trivial, use cases are limited and it was also never actually parsed in `rmf_traffic_editor`.

A [dummy model](https://app.gazebosim.org/OpenRobotics/fuel/models/RMF_Materials) has been uploaded to Fuel to contain materials and textures, on parsing of legacy buildings, texture names will be converted to point to the fuel path and fetched at runtime.
The `AssetSource` components was used which means textures can also be local files if needed.
UV calculation for walls and floors has been updated to use the texture width and height parameters. These values specify the size, in meters, of the texture and can be used to scale it.
Finally the alpha parameter has been added to walls to allow users to make them semi-transparent.
An implementation detail is that bevy stretches textures by default instead of tiling them, so I had to set the default sampler to tiling textures. This could have been done on a per-asset basis but it doesn't seem it had any repercussion since our models don't rely on texture stretching.

A demo video of the functionality is below, key takeaways:

* Floors now respect the material specified in the project (before it was a plain grey color).
* Textures can be updated, fetched from fuel, and displayed at runtime.

[Screencast from 2023-06-08 11-47-54.webm](https://github.com/open-rmf/rmf_site/assets/9111558/737ccf32-67c1-43e5-bf68-554628e6fe1d)


mxgrey (approved):

Yadunund:
I think this PR might have mistakingly reversed changes from #161 when resolving the merge conflict https://github.com/open-rmf/rmf_site/pull/136/commits/44616601ac08c107b53a922472d590049cdfe6e3

PR #137, opened 2023-06-16
Add a Pre-Commit hook.

arjo129:
`Pre-commit` is a tool which checks commits before they are made. This can help reduce the amount of hammering our CI takes in the event we forget to style something.

Install `pre-commit`:
```
sudo apt install pre-commit
# OR
pip install pre-commit
# OR
brew install pre-commit
# OR
conda install -c conda-forge pre-commit
```

Once installed run `pre-commit install`. Next time you commit, you will be stopped if you don't pass style checks.

Gazebo uses it here: https://github.com/gazebosim/gz-sim/pull/1792


mxgrey (approved):
Looks like a nice way to spare us from unnecessary style iterations.

When I tested it on my machine it took longer than I would expect to run all the commit hooks, but since this is an opt-in feature I suppose it's okay.

PR #138, opened 2023-06-21
Add display for double doors

luca-della-vedova:
### Implemented feature

This PR is part of the work towards https://github.com/open-rmf/rmf_site/issues/76, no animation is done yet to keep the PR simple but display support for double doors and asymmetric doors (`left_right_ratio` parameter) has been added.
Future work will be based on these changes and add animation on top to fully address the feature request.

A new `DoorBodyType` enum has been added to store the mesh entities for different door elements, the animation plugin will match this enum to figure out how to actuate doors.
Double doors will now spawn two meshes with a small gap inbetween, live updates for the `left_right_ratio` parameter will update the scale of the doors to match the value.


rmf_site_editor/src/interaction/outline.rs
```diff
@@ -134,7 +134,7 @@ pub fn update_outline_visualization(
     mut commands: Commands,
     outlinable: Query<
         (Entity, &Hovered, &Selected, &OutlineVisualization),
-        Or<(Changed<Hovered>, Changed<Selected>)>,
+        Or<(Changed<Hovered>, Changed<Selected>, Changed<Children>)>,
```
luca-della-vedova, line 137: This is an unfortunate side effect of this PR. Changing a door from single to double will spawn an additional mesh but the system that updates the hovering and displays outlines will not be triggered since the `Hovered` and `Selected` components didn't change, meaning that the newly spawned door will not have its outline displayed until the mouse hovers on it again.

This approach works but the performance implication might not be great since now this system will be triggered much more often (`Changed<Children>` is fairly common).

Other approaches that I considered:
* Just triggering a change detection by doing a mutable access to the root door entity's `Hovered` component when a new door is spawned. Works and only a few lines but a bit hacky.
* Manually copying the `Outline` components when spawning a new door, probably the best option from a performance point of view but would introduce a fair bit of outline related code in the door module

luca-della-vedova: https://github.com/open-rmf/rmf_site/pull/138/commits/5461c93e68ecf678a1871c9c525f83b06baf756e is a tentative approach to optimize this, it's a new system that looks at newly spawned meshes, iterates over their parents and copies the outline components to the new mesh if any is found. It's a bit hard to benchmark performance but `Added<Handle<Mesh>>` should be triggered quite a bit less than `Changed<Children>` so it should be a better solution.

mxgrey: The outline copying system for new meshes is an interesting idea, but outline behavior is supposed to be based on [`VisualCue` settings](https://github.com/open-rmf/rmf_site/blob/571b397031fa45fd185d57befef32d2065e95d5b/rmf_site_editor/src/interaction/visual_cue.rs#L71-L82), not just simple propagation.

Overall I actually like the idea of dirtying the `Hover` component to trigger the recalculation. We do something similar [here](https://github.com/open-rmf/rmf_site/blob/571b397031fa45fd185d57befef32d2065e95d5b/rmf_site_editor/src/site/lift.rs#L492) to update lifts using `.set_changed()`.

luca-della-vedova: Changed to manual `set_changed` to `Hovered` component https://github.com/open-rmf/rmf_site/pull/138/commits/e2b35dd839abbe397cb2378b60e45325754e3b2a

rmf_site_format/src/door.rs
```diff
@@ -186,12 +186,26 @@ impl From<SingleSwingDoor> for DoorType {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct DoubleSwingDoor {
     pub swing: Swing,
+    /// Length of the left door divided by the length of the right door
+    pub left_right_ratio: f32,
```
mxgrey, line 190: I've always found this ratio to be kind of clunky and unintuitive. It's something we carried over from traffic editor, but I'm not sure it's the best representation of the split in doors.

I wonder if we should do a value `0.0 -> 1.0` representing what percentage of the frame width is for the "left" door, then the right door will just be `1.0 - x`.

mxgrey (approved):
I think the changes in this PR are good, but I thought of something related that's been irking me for a while so I left a comment on it. We can address it in this PR or in a follow up, either is fine with me.

PR #139, opened 2023-06-23
Add components for opening / closing doors and door behavior preview

### LINKED ISSUE #76: Better door displays

luca-della-vedova:
### Implemented feature

This PR builds on top of #138 and closes #76. The door behavior is currently more of a "teleport to destination" rather than smoothly open or close but doors can be commanded and their states read.

[Screencast from 2023-06-23 15-36-37.webm](https://github.com/open-rmf/rmf_site/assets/9111558/8f5b676a-f2c9-4039-b768-b3e73e1bb80d)

To avoid dependency on physics engines the transforms for door opening / closing are currently done manually.
A `DoorComand` and `DoorState` component have been added, they are mostly the same but while `DoorCommand` is only `Open` or `Close`, `DoorState` has an additional `Moving` variant.
Since doors effectively teleport right now this state is never reached, however all the infrastructure is in place to populate that once a door's transform is detected to be neither fully open nor fully closed.
There are some TODOs in place for better system ordering, had to hack around to avoid off-by-one-frame glitches, however since we will need to pretty much rewrite our whole state / system pipeline when migrating to bevy 0.10+ I didn't put too much effort into creating new stages and setting them up.
Ideally there would be a new `Simulate` stage separate from the others so animation systems don't conflict with other update systems (i.e. a system that opens a door + a system that moves the door because its anchors' positions changed).

I have a feeling the `preview` model should not live in `interaction` anymore and each site module should manage its own preview, but I didn't do the refactor here to avoid making the diff explode with physical camera preview changes.

### DIFF
--- a/rmf_site_editor/src/interaction/mod.rs
+++ b/rmf_site_editor/src/interaction/mod.rs
@@ -172,7 +172,8 @@ impl Plugin for InteractionPlugin {
                             .after(update_gizmo_release),
                     )
                     .with_system(handle_lift_doormat_clicks.after(update_gizmo_click_start))
-                    .with_system(manage_previews)
+                    .with_system(make_new_entities_previewable)
+                    .with_system(manage_camera_previews)
                     .with_system(update_physical_camera_preview)
                     .with_system(dirty_changed_lifts)
                     .with_system(handle_preview_window_close),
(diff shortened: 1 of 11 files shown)


luca-della-vedova:
Closing for now, this might be implemented following a different architecture / pattern.

PR #140, opened 2023-06-26
Add missing systems for model spawning in workcell editor mode

luca-della-vedova:
### Fixed bug

When refactoring the sdf module into separate systems, the new systems were not added to the workcell editor mode, making SDF support and scale updating not work.

### Fix applied

Add the missing systems so SDFs can be spawned in workcell editor mode and model scales can be updated at runtime. Now the "Spawn visual" and "Spawn collision" should work. Also live scale update should work.


mxgrey (approved):
I've also added in the `clear_model_trashcan` system to make sure the workcell editor is also doing the model cleanup.

PR #141, opened 2023-06-27
Add icons for location tags

### LINKED ISSUE #101: Add icons to the location markers
Locations have a special ring marker with a few small boxes circling the ring. We should add icons to those boxes based on what kind of "tags" (e.g. charger, parking spot, holding point) the location has been assigned.

luca-della-vedova:
### Implemented feature

Closes #101. Adds icons to the small rectangles around the locations to show their properties on first inspection. An example of the behavior is below:

[Screencast from 2023-06-27 10-28-28.webm](https://github.com/open-rmf/rmf_site/assets/9111558/8d38c725-da2c-422d-b6ad-aceb9475425c)

I adapted the available icons from traffic editor and sketched two new ones for workcells and robot spawning. They don't look great but I'm not sure how to do them better, happy to substitute them with anything.
The code is very verbose because of the tags being contained in a vector, I'll play around a bit and see if I can simplify it a bit (but that might require refactoring the `LocationTags` component).


luca-della-vedova:
Cleaned it up a bit, I suspect some of the duplicated code to go through new tags and spawn meshes if necessary could be refactored into a separate function but since the duplicated functions differ a bit (one always spawns all tags on new location, one needs to check for differences on updated tags) it would be not too immediate / introduce a bit of overhead, but that can be done if needed.

luca-della-vedova:
With https://github.com/open-rmf/rmf_site/pull/141/commits/557d68fef3e07118fd0c44a6b0cc3f8f2cec14a4, removed spawn robot and workcell tag visualization since these will be handled by the upcoming Scenarios

mxgrey (approved):

PR #146, opened 2023-07-14
Add panic hook for debugging wasm builds

luca-della-vedova:
### Implemented feature

Add a panic hook to help debugging wasm builds. With this change the panic error is propagated all the way to the web console, making debugging a lot easier.

Implementation follows the example from the crate


mxgrey (approved):
Very helpful for debugging in web browsers 👍

PR #149, opened 2023-07-14
Add fuel asset gallery

luca-della-vedova:
### Implemented feature

This PR adds an asset gallery that is allows spawning models from the [Gazebo App](https://app.gazebosim.org/dashboard) database.
To access the asset gallery navigate to Create -> New Model -> Browse fuel, as shown below:

[Screencast from 2023-07-14 14-49-12.webm](https://github.com/open-rmf/rmf_site/assets/9111558/2d005020-9fe4-4978-9a2c-85fdce3286bc)

The core implementation of the fuel client is in a separate [gz-fuel](https://github.com/luca-della-vedova/gz-fuel/) crate I wrote for the purpose. I'm happy to either release it, keep a source dependency or just copy the code here to remove the dependency. More detailed list of changes:

* Added Asset gallery, including model preview widget based on a camera + preview model with a dedicated `RenderLayer` to make sure they are not visible to other views.
* Assets can be filtered by owner, tag or private / public property. Filters are chained to allow more fine grained control of the visualized models. The entire database is above 2500 models so the feature is quite necessary.
* The `gz-fuel` crate saves the model cache to a json file when possible (every platform except wasm). Users can trigger an update of the cache through a button in the GUI. Update is `async` and goes to a background task in the IO thread pool to avoid having too much overhead.
* Support for private models in fuel, users will need to set an API key to access them.
* More helpful error messages, hinting to users whether the model fails to visualize because it's not supported (i.e. `dae` files that are very prevalent in Fuel) or because it was not found in the path / fuel.
* Event infrastructure to trigger a re-download of failed models if the API key is updated.
* Improve support for SDF, previously only `AssetSource::Search` sdfs would work, this won't be the case anymore and both `AssetSource::Local` and `AssetSource::Remote` SDFs should work. For example browsing through the `Local` asset source to a `model.sdf` file should successfully spawn the asset to the site.
* Minor cleanups to model / asset source logic. Models had a system that had two queries and iterated on both of them, trivially split into two systems with one query each to improve parallelization.

Note that sadly because a lot, or even the majority of, the models in fuel are in `.dae` and there is no support for Collada, a lot of models will actually fail to spawn but that is out of the scope of this PR.


rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
-        AssetSource::Remote(model_path.to_owned() + "/" + uri)
-    } else {
-        AssetSource::Remote("".into())
+    let mut asset_source = AssetSource::from(path);
+    match asset_source {
+        AssetSource::Remote(ref mut p) | AssetSource::Search(ref mut p) => {
+            let binding = p.clone();
+            *p = if let Some(stripped) = uri.strip_prefix("model://") {
+                // Get the org name from context, model name from this and combine
+                let org_name = binding.split("/").next().unwrap();
```
mxgrey, line 41: I'm concerned about using `.unwrap()` here. Couldn't a user crash this by inputting some bad text? What if we do something like `.unwrap_or("")` and allow it to just fail to find the asset later in the pipeline?

rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
+        }
+        AssetSource::Local(ref mut p) => {
+            let binding = p.clone();
+            *p = if let Some(stripped) = uri.strip_prefix("model://") {
+                // Search for a model with the requested name in the same folder as the sdf file
+                // Note that this will not play well if the requested model shares files with other
+                // models that are placed in different folders or are in fuel, but should work for
+                // most local, self contained, models.
+                // Get the org name from context, model name from this and combine
+                let model_folder = binding.rsplitn(3, "/").skip(2).next().unwrap();
```
mxgrey, line 63: I'm also concerned about the use of `.unwrap()` here. Maybe do `.unwrap_or("")` here too?

rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
+            *p = if let Some(stripped) = uri.strip_prefix("model://") {
+                // Get the org name from context, model name from this and combine
+                let org_name = binding.split("/").next().unwrap();
+                org_name.to_owned() + "/" + stripped
+            } else if let Some(path_idx) = binding.rfind("/") {
+                // It's a path relative to this model, remove file and append uri
+                let (model_path, _model_name) = binding.split_at(path_idx);
+                model_path.to_owned() + "/" + uri
+            } else {
+                println!(
```
mxgrey, line 48:
```suggestion
                error!(
```

rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
+                    "WARNING: Invalid model found path is {} and model uri is {}",
```
mxgrey, line 49:
```suggestion
                    "WARNING: Invalid SDF model path. Path is [{}] and model uri is [{}]",
```

rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
+                // most local, self contained, models.
+                // Get the org name from context, model name from this and combine
+                let model_folder = binding.rsplitn(3, "/").skip(2).next().unwrap();
+                model_folder.to_owned() + "/" + stripped
+            } else if let Some(path_idx) = binding.rfind("/") {
+                // It's a path relative to this model, remove file and append uri
+                let (model_path, _model_name) = binding.split_at(path_idx);
+                model_path.to_owned() + "/" + uri
+            } else {
+                println!(
```
mxgrey, line 70:
```suggestion
                error!(
```

rmf_site_editor/src/site/sdf.rs
```diff
@@ -29,21 +29,56 @@ use rmf_site_format::{
...
+                    "WARNING: Invalid model found path is {} and model uri is {}",
```
mxgrey, line 71:
```suggestion
                    "WARNING: Invalid SDF model path. Path is [{}] and model uri is [{}]",
```

rmf_site_editor/src/site_asset_io.rs
```diff
@@ -61,12 +64,13 @@ impl SiteAssetIo {
...
     ) -> BoxedFuture<'a, Result<Vec<u8>, AssetIoError>> {
         Box::pin(async move {
-            let bytes = surf::get(remote_url.clone())
-                .recv_bytes()
-                .await
-                .map_err(|e| {
-                    AssetIoError::Io(io::Error::new(io::ErrorKind::Other, e.to_string()))
-                })?;
+            let mut req = surf::get(remote_url.clone());
+            if let Some(key) = &(*FUEL_API_KEY.lock().unwrap()) {
```
mxgrey, line 68: It's not likely to be a problem, but we should still avoid using `.unwrap()` here. An `Err` result means the mutex was poisoned, which means whatever thread had the mutex locked panicked while holding it. The information inside might be fine or it might be mangled in some unpredictable way.

We can use [`into_inner`](https://doc.rust-lang.org/std/sync/struct.PoisonError.html#method.into_inner) on the `PoisonedError` to access the underlying data, and we can then reset it to `None` and print a `warn!` to the user telling them to input their API key again. This would be a way to gracefully recover from a mutex poisoning.

mxgrey (changes requested):
This is very awesome. I've just left a few comments, mostly about places where `.unwrap()` is used where I think we can have safer alternatives.

rmf_site_editor/src/site/sdf.rs
luca-della-vedova: For both this and the similar issue, changed it to return empty string but logging an error to denote exactly what went wrong, instead of doing it silently. It should be a bit more helpful in diagnosing issues https://github.com/open-rmf/rmf_site/pull/149/commits/91a8256eee8a8ff6ffffade21b518a20c7df31f1

rmf_site_editor/src/site_asset_io.rs
luca-della-vedova: I tried to address it but maybe not in the best way in https://github.com/open-rmf/rmf_site/pull/149/commits/91a8256eee8a8ff6ffffade21b518a20c7df31f1, lock poisoning wouldn't happen in the current use case (unless a string allocation failed, but in that case we would probably have many more issues to worry about!). For now it returns an error and fails to load the asset since the variable might be corrupted. I've seen the http library panicking when headers are given invalid values with messages like:

```
thread 'IO Task Pool (2)' panicked at 'called `Result::unwrap()` on an `Err` value: http::Error(InvalidHeaderValue)', /usr/local/google/home/yadunund/.cargo/registry/src/index.crates.io-6f17d22bba15001f/http-client-6.5.3/src/isahc.rs:58:42
```

So propagating a potentially corrupted string might cause other issues down the line. Anyway if you feel we should still consume it feel free to push directly!

mxgrey: Your solution is definitely better than what I was proposing.

I did one last little iteration on it to clear out the key to `None` just in case it contains some corrupted data: [b13f1f0](https://github.com/open-rmf/rmf_site/pull/149/commits/b13f1f076425932de0abec6b933e3bf29dfd614c)

mxgrey:
I was ready to approve this PR until I did a manual test and discovered that something broke for model scaling. Here's a screenshot from the hotel demo world:

Screenshot from 2023-08-24 12-16-55

I might have mangled something while resolving merge conflicts. I'll see if I can fix it.

mxgrey:
Never mind, the scaling issue was fixed after refreshing my model cache.

mxgrey (approved):

PR #150, opened 2023-07-14
Fix panic when updating model sources

### LINKED ISSUE #145: [Bug]: Crash when removing, then re-adding a model
### Description of the bug

A reproducible crash can happen when updating a fully spawned model to something else that doesn't exist, then back to the original.

1. Run the site editor in the demo world
2. Change a model to trigger a new asset lookup to something that doesn't exist, i.e. a `OpenRobotics/AdjTable` to `OpenRobotics/AdjTable2`.
3. Change the model back to the previous, i.e. `OpenRobotics/AdjTable`
4. Panic!

After spending some time debugging the issue I narrowed it down to a complex interaction between the code to spawn models and the `Aabb` plugin, specifically:

* When the asset source for a model is changed, its entire scene, including entities that will contain `Handle<Mesh` components is despawned .
* The [Aabb](https://github.com/open-rmf/rmf_site/blob/main/rmf_site_editor/src/aabb.rs) plugin internally keeps track of what entities have what mesh handle, however it only does cleanup when the `Handle<Mesh>` component is removed. The query for removed `Handle<Mesh>` components does _not_ get triggered when an entity is despawned, so the internal hash map falls out of sync with what entities actually exist and contain the mesh handle.
* Next time the same handle is added any entity, this for loop will be triggered and because an old entity, that has since been despawned, refers to that handle, a command for an entity that doesn't exist will be triggered [here](https://github.com/open-rmf/rmf_site/blob/f68041b75f4f763a13e782e44773bae1cf7ae10f/rmf_site_editor/src/aabb.rs#L120) and the application will panic.

This issue is quite painful when spawning / despawning models, however the code was copied from an old PR in bevy. It seems mesh aabb update is still in progress in main so it's not going to land anytime soon and we will need to keep relying on our custom aabb module.

luca-della-vedova:
### Fixed bug

Fixes #145.

### Fix applied

The fix is sadly to just comment out, with a note explaining why, the code that causes the panic.
It seems the only place where we use the pattern of accessing a mesh by its handle, mutating it, and the event being caught by this system is in the model logic, specifically [here](https://github.com/open-rmf/rmf_site/blob/369766219e0c9e58debe872ec959f5b292db7028/rmf_site_editor/src/site/model.rs#L340-L349). Now this is also the section that causes the panic in the first place since the missing cleanup issue is there. Furthermore in this specific case:

* The aabb doesn't really change because we are only mutating the mesh to add outline normals.
* It's not necessary to propagate to all the meshes that share the same handle, since this is done for each newly spawned mesh anyway.

The workflow that might have worked before but will not after this PR is:

* Create a mesh handle.
* Access the mesh data through its handle.
* Change the mesh in a way that would change its bounding box.
* Bounding box for all the entities that use that mesh handle *was* updated, it won't be anymore.

But again we only use this pattern in the one case that causes a panic so it is too brittle anyway.

An alternative fix could be to avoid using `despawn_recursive` throughout the codebase and sending `Delete` events, together with having the aabb systems track the Delete events to do the cleanup, but this would also be a bit brittle if deletion was to fail for any reason.
Another way could be to despawn in two steps, removing the `Handle<Mesh>` components first, making the `RemovedComponent<Handle<Mesh>>` query behave as expected, but this would be quite clunky.
A proper fix would involve having a way to catch despawn events, but I am not aware of any way to do this.

### DIFF
--- a/rmf_site_editor/src/aabb.rs
+++ b/rmf_site_editor/src/aabb.rs
@@ -88,7 +88,7 @@ pub fn update_bounds(
     meshes: Res<Assets<Mesh>>,
     mut mesh_reassigned: Query<(Entity, &Handle<Mesh>, &mut Aabb), Changed<Handle<Mesh>>>,
     mut entity_mesh_map: ResMut<EntityMeshMap>,
-    mut mesh_events: EventReader<AssetEvent<Mesh>>,
+    //mut mesh_events: EventReader<AssetEvent<Mesh>>,
     entities_lost_mesh: RemovedComponents<Handle<Mesh>>,
 ) {
     for entity in entities_lost_mesh.iter() {
@@ -105,6 +105,15 @@ pub fn update_bounds(
         }
     }
 
+    // Note (luca) This has been removed since entity despawns are not caught to cleanup the
+    // entity_mesh_map and there is a possibility of panic in case we try to add an aabb to an
+    // entity that has since been despawned (i.e. when updating models' AssetSource and their scene
+    // is despawned).
+    // Tracking issue here with the details https://github.com/open-rmf/rmf_site/issues/145
+    // It seems we don't currently change meshes themselves outside of the model systems that cause
+    // the panic, with the general pattern in the editor being to reassign new handles
+    // so this event is never created anyway
+    /*
     let to_update = |event: &AssetEvent<Mesh>| {
         let handle = match event {
             AssetEvent::Modified { handle } => handle,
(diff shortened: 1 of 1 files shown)


mxgrey:
I ran into a similar problem while working #142 . I solved it in that PR by introducing a `ModelTrashcan` which delays the despawning of model descendants until a later stage, after all of their other commands have been flushed.

I believe the root conflict for `update_bounds` is that it runs during the same stage as `update_model_scenes` so some of these `Aabb` insertion commands may happen on entities that were deleted earlier during the same command flush. With the changes I made in #142 that should no longer be an issue so I believe it's safe to close this PR and not make any further changes for this.

Feel free to reopen the PR if you think I'm overlooking anything.

PR #153, opened 2023-07-24
Introducing scenarios concept

mxgrey:
This PR introduces the concept of scenarios and begins moving the implementation of "models" in a new direction:
* A "model" is a blueprint for something that can exist in the world, e.g. mobile robot, workcell, or furnishings
* Each asset that is seen in the world is an "instance" of a model
* Scenarios are defined by what "instances" they contain and how those instances are positioned

This PR leaves the data structures in a limbo where `Model` still exists to refer to furnishings that are laid out around the levels. I left those as-is to minimize how disruptive this PR is. They will also be migrated into the new scenario concept after the existing PRs are merged in and there's less risk of merge conflicts. They will probably be referred to as `Decor` or `Furnishings`.

There is also a flaw in this PR where robots will become invisible when their parent anchor gets hidden. This will be fixed in the future after we migrate to the latest version of Bevy.

### DIFF
--- a/rmf_site_format/src/location.rs
+++ b/rmf_site_format/src/location.rs
@@ -19,67 +19,32 @@ use crate::*;
 #[cfg(feature = "bevy")]
 use bevy::prelude::{Bundle, Component, Deref, DerefMut, Entity};
 use serde::{Deserialize, Serialize};
-
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-pub enum LocationTag {
-    Charger,
-    ParkingSpot,
-    HoldingPoint,
-    SpawnRobot(Model),
-    Workcell(Model),
-}
-
-impl LocationTag {
-    pub fn label(&self) -> &'static str {
-        match self {
-            Self::Charger => "Charger",
-            Self::ParkingSpot => "Parking Spot",
-            Self::HoldingPoint => "Holding Point",
-            Self::SpawnRobot(_) => "Spawn Robot",
-            Self::Workcell(_) => "Workcell",
-        }
-    }
-
-    pub fn is_charger(&self) -> bool {
-        matches!(self, Self::Charger)
-    }
-    pub fn is_parking_spot(&self) -> bool {
-        matches!(self, Self::ParkingSpot)
-    }
-    pub fn is_holding_point(&self) -> bool {
-        matches!(self, Self::HoldingPoint)
-    }
-    pub fn spawn_robot(&self) -> Option<&Model> {
-        match self {
-            Self::SpawnRobot(model) => Some(model),
-            _ => None,
-        }
-    }
-    pub fn workcell(&self) -> Option<&Model> {
-        match self {
-            Self::Workcell(model) => Some(model),
-            _ => None,
-        }
-    }
-}
+use std::collections::{BTreeSet, HashSet, HashMap};
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "bevy", derive(Bundle))]
 pub struct Location<T: RefTrait> {
     pub anchor: Point<T>,
+    #[serde(flatten)]
     pub tags: LocationTags,
     pub name: NameInSite,
     pub graphs: AssociatedGraphs<T>,
 }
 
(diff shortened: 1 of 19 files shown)


rmf_site_format/src/location.rs
```diff
@@ -19,67 +19,32 @@ use crate::*;
...
 
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-#[serde(transparent)]
-#[cfg_attr(feature = "bevy", derive(Component, Deref, DerefMut))]
-pub struct LocationTags(pub Vec<LocationTag>);
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+#[cfg_attr(feature = "bevy", derive(Component))]
+pub struct LocationTags {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub charger: Option<String>,
```
luca-della-vedova, line 38: What does `String` contain (since it was bool before)? Is it meant to be something like a `json` that will be sent to the robot, or just a human readable description? This could also be "documented" in the widget through hover text or a label. +100 for the change to a `LocationTags` struct since the `Vec<LocationTag>` Component made tag search and update a bit painful before.

rmf_site_format/src/mobile_robot.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg_attr(feature = "bevy", derive(Component))]
+pub enum MobileRobotKinematics {
+    DifferentialDrive(DifferentialDrive),
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct DifferentialDrive {
+    pub translational_speed: f32,
+    pub rotational_speed: f32,
```
luca-della-vedova, line 43: Would it be worth it to add linear / angular accelerations as well? So we can wrap all the data that would be in the fleet adapter `limits` structures

luca-della-vedova (changes requested):
I gave it a spin and not too much to say other than what is in the description, few minor comments scattered around.

I did however run into a major issue, where it seems project loading / saving is broken, specifically by running the demo scenario, saving it, then trying to load it, loading failed with the following error:

```
2023-07-25T04:39:46.735174Z ERROR librmf_site_editor::workspace: Failed loading site Error { code: ExpectedIdentifier, position: Position { line: 254, col: 14 } }
```

Which specifically points to the first location:

```
43: {"anchor": 42, "charger": Some(""), "parking": Some(""), "holding": Some(""), "name": "tinyRobot1_charger", "graphs": All},
```

> There is also a flaw in this PR where robots will become invisible when their parent anchor gets hidden. This will be fixed in the future after we migrate to the latest version of Bevy.

This might require a bit more thought about the parenthood. It is true that there will be a "always visible" variant for the `Visibility` component, which ignores hierarchy's visibility. However, since we use the hierarchy visibility to show / hide levels, the moment we set an entity to always visible this will be ignored and we might see all the robots in other levels in the current view.

rmf_site_format/src/mobile_robot.rs
mxgrey: I've come to the conclusion that for the next generation of RMF we don't actually want to use the acceleration data. The acceleration numbers provided by users are usually made up, and don't seem to have a meaningful effect on the overall system anyway. Reducing the parameters to just the nominal first-order speed is a strategic choice.

If we need to export backwards-compatible config files, e.g. to be consumed by the current `rmf_fleet_adapter` implementation, we can just automatically fill in 1.0 for linear and 6.0 for angular accelerations.

rmf_site_format/src/location.rs
mxgrey: I'm still deliberating this. For now it's a placeholder meant for us to experiment with. My current expectation is it can be a `json` containing information about what type of charger it is. E.g.
```
{
  "brand": "Xnergy",
  "height_range": [0.1, 2.0],
  "current": 2.0,
  "voltage": 15
}
```
then there can be application logic somewhere to decide which robots are compatible with the charger.

mxgrey:
Closing this in favor of #242

Issue #61, opened 2022-10-04
Audit / Diagnostic tool


mxgrey:
Provide a tool that looks for common mistakes or potential issues within the site data. The user would be provided with a list of warnings with advice on how to fix it.

In the event that the user wants to ignore certain warnings, they could click a button to suppress an item in the list. That suppression will be saved to their site file so it is remembered next time. There would also be a way to view what warnings have been suppressed from inside the GUI.

Some examples of possible warnings:
- [ ] Two anchors are very close to each other but not connected
- [ ] Multiple doors/lifts/models share the same name
- [ ] A fiducial is missing its label

Feel free to comment with suggestions for more diagnostics.

luca-della-vedova:
Together with the fiducials with no name, there are several issues currently captured by `warn` messages in the (future) drawing editor mode that could be changed to be captured by this tool, such as:

* Building has multiple levels but < 2 matching fiducials between the reference level and each of the other ones. Site editor would fail to perform level alignment.
* < 2 constraints found between the level drawing and a robot map, same issue as above when optimizing transforms.

We could potentially also do additional validation on the navgraph, for example make sure that there is at least one charger location per graph id, to avoid robots being unable to charge.

PR #154, opened 2023-07-25
Implement a diagnostic tool

luca-della-vedova:
### Implemented feature

Addresses most of #61, excluding the additional recommendations and the model issue (since lift / door models are not supported yet). 
It is currently in draft and based on `luca/drawing_editor` because it needed some of the UI refactoring that went into that PR. However it also needs some of the refactoring that is in the fuel asset widget PR, so will move out of draft once that is finalized (and this widget is moved into a left panel rather than a floating window).

This PR adds the following features:
* Checks for common types of issues, such as doors / lifts / docks with the same name, fiducials without a label and anchors that are very close but unconnected.
* Suppressing issues by category.
* Suppressing single issues.
* Serializing / deserializing suppression data in the `SiteProperties` to make it persistent across sessions.

Example of the current behavior in the video below:

[Screencast from 2023-07-25 11-20-31.webm](https://github.com/open-rmf/rmf_site/assets/9111558/f20c1295-010f-40b7-ae97-be3d78058966)

The architecture of the issues finding / inserting is fully decentralized. Issues are entities spawned as children of the workspace and systems are responsible for registering them (mapping a UUID to the issue type through a new `App` trait) and spawning them, for now only as a response to validation events.

There is one slightly brittle aspect to the solution, which caused the odd system ordering. Specifically, since issues are entities it is necessary to despawn them when a new validation request is received, to avoid having duplicates.
At the same time, it would be a fair bit of code duplication to have every system be responsible for cleaning up its own issues, then respawning them afterwards, so I created a system that performs the cleanup. Right now the events are sent in `Update`, cleanup is performed in `PostUpdate` and population with new issues in `PreUpdate`, if the ordering was changed in a refactoring this feature might behave unexpectedly (i.e. the cleanup system clearing all the newly created issues if it runs after their creation and the issue list always being empty).


luca-della-vedova:
Sadly there is a fair bit of noise coming from `cargo fmt` since `let else` was just stabilized, hopefully most of these changes should be straightforward

mxgrey (approved):
I made some tweaks; one important one to make sure that saving/loading works correctly.

This feature is awesome, and I look forward to adding more warnings to it over time.

PR #155, opened 2023-07-27
Save workcell as URDF

audrow:
The current workcell is saved as a URDF when you click "Export URDF" in the "File" menu.

Note this doesn't include mass/inertia or joint data. I'll make separate PRs for those.

<details><summary>Output URDF</summary>

```xml
<robot name="test_workcell">
  <link name="world">
    <inertial>
      <origin xyz="0 0 0" rpy="0 -0 0" />
      <mass value="0" />
      <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0" />
    </inertial>
    <visual name="">
      <origin xyz="1.2212629318237305 -2.776434898376465 0" rpy="0 0 0" />
      <geometry>
        <mesh filename="package://nexus_poc/meshes/SCARA_Table_only.stl" scale="0.009999999776482582 0.009999999776482582 0.009999999776482582" />
      </geometry>
    </visual>
  </link>
  <link name="top_anchor">
    <inertial>
      <origin xyz="-1.0387370586395264 -2.961110830307007 1.2591593265533447" rpy="0 0 1.5707963705062866" />
      <mass value="0" />
      <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0" />
    </inertial>
  </link>
  <link name="goal_1">
    <inertial>
      <origin xyz="0 0.5 0" rpy="0 0 -1.5707963705062866" />
      <mass value="0" />
      <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0" />
    </inertial>
  </link>
  <link name="goal_2">
    <inertial>
      <origin xyz="-0.5 0 0" rpy="0 0 -1.5707963705062866" />
      <mass value="0" />
      <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0" />
    </inertial>
  </link>
</robot>
```

</details>

<details><summary>Workcell</summary>

```json
{
  "name": "test_workcell",
  "id": 0,
  "frames": {
    "2": {
      "parent": 0,
      "Pose3D": {
        "trans": [
          -1.038737,
          -2.9611108,
          1.2591593
        ],
        "rot": {
          "yaw": {
            "deg": 90.0
          }
        }
      },
      "name": "top_anchor",
      "mesh_constraint": {
        "entity": 1,
        "element": {
          "Vertex": 0
        },
        "relative_pose": {
          "trans": [
            -2.26,
            -0.18467593,
            1.2591593
          ],
          "rot": {
            "yaw": {
              "deg": 90.0
            }
          }
        }
      }
    },
    "3": {
      "parent": 2,
      "Pose3D": {
        "trans": [
          0.0,
          0.5,
          0.0
        ],
        "rot": {
          "yaw": {
            "deg": -90.0
          }
        }
      },
      "name": "goal_1"
    },
    "4": {
      "parent": 2,
      "Pose3D": {
        "trans": [
          -0.5,
          0.0,
          0.0
        ],
        "rot": {
          "yaw": {
            "deg": -90.0
          }
        }
      },
      "name": "goal_2"
    }
  },
  "visuals": {
    "1": {
      "parent": 0,
      "name": "",
      "geometry": {
        "Mesh": {
          "filename": "package://my_pkg/meshes/robot.stl",
          "scale": [
            0.01,
            0.01,
            0.01
          ]
        }
      },
      "pose": {
        "trans": [
          1.2212629,
          -2.776435,
          0.0
        ],
        "rot": {
          "yaw": {
            "deg": 0.0
          }
        }
      }
    }
  },
  "collisions": {}
}
```

</details>


luca-della-vedova:
Very nice! The general approach looks good I see some of the parts are still left to do where:

* Visuals / Collisions should be pretty straightforward, since  the implementation is pretty much there. They are just children of frames so by doing iteration over children of frames when creating links you should be able to populate them.
* Mass and inertial data should be still quite simple, the components are there and they should be deserialized 
* Joints data is more tricky, I think we would require need to adjust the `Joint` component to make sure it can hold references to parent and child frames, then we can just export them as they are, at least until we have a more functional implementation.

The first point should be pretty much doable in the current implementation.
The second point will need a bit of thinking as to what is a `Frame` and what is a `Link`. In order of increasing complexity we could:
* Add `Link` components to a Frame when loading a Urdf (i.e. [here](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_editor/src/workcell/load.rs#L40)), then when saving add a `Option<&Link>` to the frame and, if the component is present, serialize its data.
* The above plus have a system that adds `Link` components to all frames, then the query above would just be for `&Link` and data would always be serialized.
* Any of the above but with `Link` being a new entity that is a child of `Frame` instead of a component of `Frame`, to have a bit more encapsulation.
We will also need to add the [missing data](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L92) to the inertial structs.

The third point will be tricky, first of all the `Joint` [struct](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L117-L119) will have to be updated to have the missing joint data (like type, limits etc) and hold references to child and parent frames. References to entities can be implemented through a generic with types that implement `RefTrait` as is done in [MeshConstraint](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L58-L62). The idea is that both `u32` and `Entity` implement `RefTrait` and we serialize / deserialize `u32` while using `Entity` internally.

Now the tricky part is that some of these points will be hard to do without a functional `Load` implementation for workcells, so I'm happy to split up the work into separate PRs as you see fit.

The only suggestion I have on the current implementation is on the `From` function. While `From` makes conversion quite simple, is has the issue that there is no significant way to encode failure since it is forced to return a `urdf_rs::Robot`. This means that, if anything went wrong during conversion, you are only left with the choice of panicking, logging an error and continuing, or return an empty structure of sorts to encode failure and leave the function caller to figure out that it went wrong.

By contrast, if you had your own conversion function you could return either an `Option<urdf_rs::Robot>` or a `Result<urdf_rs::Robot, UrdfConversionError>`, an example of this being done is when we convert legacy building projects to sites [here](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/legacy/building_map.rs#L122)

audrow:
> The only suggestion I have on the current implementation is on the `From` function. While `From` makes conversion quite simple, is has the issue that there is no significant way to encode failure since it is forced to return a `urdf_rs::Robot`

In trying to implement this, it seems that there isn't a reason that the code would fail here. As long as the workcell is correctly defined, it should just be a data mapping, and I think that Rust will stop us if there is a mismatch in the data structures. Otherwise, am I missing something?

audrow:
> * Visuals / Collisions should be pretty straightforward, since  the implementation is pretty much there. They are just children of frames so by doing iteration over children of frames when creating links you should be able to populate them.

I added visuals in bb09860 and collisions in ccb7915. The `test.workcell.json` example doesn't have any collisions, so I haven't been able to test that, but I follow a similar process to what I do for visuals. I updated the example above with the visuals included.

audrow:
> The second point will need a bit of thinking as to what is a `Frame` and what is a `Link`. In order of increasing complexity we could:
> 
> * Add `Link` components to a Frame when loading a Urdf (i.e. [here](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_editor/src/workcell/load.rs#L40)), then when saving add a `Option<&Link>` to the frame and, if the component is present, serialize its data.
> * The above plus have a system that adds `Link` components to all frames, then the query above would just be for `&Link` and data would always be serialized.
> * Any of the above but with `Link` being a new entity that is a child of `Frame` instead of a component of `Frame`, to have a bit more encapsulation.
>   We will also need to add the [missing data](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L92) to the inertial structs.
> 
> The third point will be tricky, first of all the `Joint` [struct](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L117-L119) will have to be updated to have the missing joint data (like type, limits etc) and hold references to child and parent frames. References to entities can be implemented through a generic with types that implement `RefTrait` as is done in [MeshConstraint](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/workcell.rs#L58-L62). The idea is that both `u32` and `Entity` implement `RefTrait` and we serialize / deserialize `u32` while using `Entity` internally.

Maybe we put these in a separate PRs?

rmf_site_format/src/workcell.rs
```diff
@@ -280,6 +280,137 @@ pub struct Workcell {
...
+            rpy: match pose.rot {
+                Rotation::EulerExtrinsicXYZ(arr) => urdf_rs::Vec3(arr.map(|v| match v {
+                    Angle::Rad(v) => v as f64,
+                    Angle::Deg(v) => v.to_radians() as f64,
+                })),
+                Rotation::Yaw(v) => match v {
+                    Angle::Rad(v) => urdf_rs::Vec3([0.0, 0.0, v as f64]),
+                    Angle::Deg(v) => urdf_rs::Vec3([0.0, 0.0, v.to_radians() as f64]),
+                },
+                _ => todo!("Unsupported rotation type for conversion to urdf pose"),
```
mxgrey, line 295: This will cause the code to panic if `pose.rot` is a `Rotation::Quat`. That's a big problem since it would be perfectly reasonable for someone to use quaternions to describe a pose.

Another issue with this pattern is if we ever add another variant to `Rotation` then we'll never notice that we need to update this pattern matching block. We should prefer for this block to produce a compilation error if we ever add another `Rotation` variant.

So let's get rid of this line and add a pattern matcher for `Rotation::Quat(_)`. We can draw inspiration from the implementation [here](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/misc.rs#L261-L270), except I believe that was implemented before we accepted `glam` as a dependency. `glam` is the crate that bevy uses for math, and it's [now a dependency for `rmf_site_format`](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/Cargo.toml#L17) so we can feel free to use glam for things like this.

Another tip is that we don't need to do any pattern matching for `Angle` because `Angle` has a [`radians()`](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/misc.rs#L146-L151) method. E.g. consider this alternative for the first matching arm:
```
Rotation::EulerExtrinsicXYZ(arr) => urdf_rs::Vec3(arr.map(|v| v.radians())),
```

rmf_site_format/src/workcell.rs
```diff
@@ -280,6 +280,137 @@ pub struct Workcell {
...
+                        .collect();
+                    let collision: Vec<Collision> = collisions_and_parents
+                        .iter()
+                        .filter(|(_, parent)| parent == &f.1.parent)
+                        .map(|(collision, _)| collision.clone())
+                        .collect();
+
+                    let pose: urdf_rs::Pose = match frame.anchor {
+                        Anchor::Pose3D(pose) => pose.into(),
+                        _ => todo!(),
```
mxgrey, line 382: Let's make sure we implement this before we merge.

rmf_site_format/src/workcell.rs
```diff
@@ -280,6 +280,137 @@ pub struct Workcell {
...
+            Geometry::Primitive(MeshPrimitive::Capsule { radius, length }) => {
+                urdf_rs::Geometry::Capsule {
+                    radius: radius as f64,
+                    length: length as f64,
+                }
+            }
+            Geometry::Primitive(MeshPrimitive::Sphere { radius }) => urdf_rs::Geometry::Sphere {
+                radius: radius as f64,
+            },
+            _ => todo!("Only meshes and primitives are supported for conversion to urdf geometry"),
```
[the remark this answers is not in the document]
audrow: Gotcha, thank you for the clarification.

rmf_site_format/src/workcell.rs
audrow: Done in dfbf3d3.

rmf_site_format/src/workcell.rs
audrow: Done in dfbf3d3.

rmf_site_format/src/workcell.rs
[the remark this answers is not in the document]
audrow: I made it so that we don't panic in 5e376ea.

audrow:
@mxgrey, thanks for the feedback!

I believe that I've implemented the changes. I'm fairly new to Rust (and loving it), so please let me know if you see opportunities for me to do things in a more Rust-like way.

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,110 @@ impl Workcell {
...
+                let visual: Vec<Visual> = visuals_and_parents
+                    .iter()
+                    .filter(|(_, parent)| parent == &f.1.parent)
+                    .map(|(visual, _)| visual.clone())
+                    .collect();
+                let collision: Vec<Collision> = collisions_and_parents
+                    .iter()
+                    .filter(|(_, parent)| parent == &f.1.parent)
+                    .map(|(collision, _)| collision.clone())
+                    .collect();
```
luca-della-vedova, line 398: I wonder if there is a more efficient way to do this filtering.
What is happening here is a nested iteration where, for each link, we iterate through all the visuals / collisions to check if any of them is a parent of the link.
What if you did a HashMap where either:
* Instead of a `Vec<Link>` you have a `HashMap<u32, Link>` then populate it once when iterating through all the links with their ID and data or
* You keep a `Vec<Link>` but build in parallel a `HashMap<u32, &mut Link>` (don't quote me on this not sure it would actually work! It would in principle avoid doing the conversion from HashMap -> Vec when exporting).

Then you only iterate through visuals / collisions once and for each item you find which link to append it to by doing a hash map lookup.
This should flatten the nested loop and make the export a lot faster

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,110 @@ impl Workcell {
...
+
+                let pose: urdf_rs::Pose;
+                if let Anchor::Pose3D(p) = frame.anchor {
+                    pose = p.into();
+                } else {
+                    return Err(WorkcellToUrdfError::InvalidAnchorType(frame.anchor));
+                }
```
luca-della-vedova, line 405: This can be "simplified" using the return type of the `if let` expression, such as:

```
let pose = if let Anchor::Pose3D(p) = frame.anchor {
  p.into()
} else {
  return Err(WorkcellToUrdfError::InvalidAnchorType(frame.anchor));
};
```

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,110 @@ impl Workcell {
...
+                    collision,
+                    visual,
+                };
+                Ok(link)
+            })
+            .collect();
+        let links = match links {
+            Ok(links) => links,
+            Err(e) => return Err(e),
+        };
```
luca-della-vedova, line 431: There is one of my favorite operators `?` it basically "unwraps" the result returning the error if it failed.
You can just apply it to the structure above by removing the type annotation from `link` and changing the `collect()` to:

```
.collect::<Result<Vec<_>, _>>()?;
```

What _I believe_ this does is still communicate the `collect` function to bring the errors in the inner `map` function to the external collection, then unwrap it or return the error.
In this way you can get rid of this match statement entirely.

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,110 @@ impl Workcell {
...
+    }
+
+    pub fn to_urdf_string(&self) -> Result<String, WorkcellToUrdfError> {
+        match self.to_urdf() {
+            Ok(urdf) => match urdf_rs::write_to_string(&urdf) {
+                Ok(string) => Ok(string),
+                Err(e) => Err(WorkcellToUrdfError::WriteToStringError(e)),
+            },
+            Err(e) => Err(e),
+        }
```
luca-della-vedova, line 448: Similar to the above, you can change the outer match statement to:

```
let urdf = self.to_urdf()?;
[inner match statement]
```

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,110 @@ impl Workcell {
         serde_json::ser::to_string_pretty(self)
     }
 
+    pub fn to_urdf(&self) -> Result<urdf_rs::Robot, WorkcellToUrdfError> {
+        let workcell = self.clone();
```
luca-della-vedova, line 348: I'm a bit on the fence on the early `clone` here, while it makes the code below a lot simpler it could create unnecessary copies of data, compared to only iterating through references in `self` and ad-hoc cloning of only necessary variables. This might not matter too much now but if in the future we added new fields to workcells that don't map to URDF, this `clone()` call would copy all that data unnecessarily as well. 

Also, visuals and collisions are actually cloned twice, once in this line and once [here](https://github.com/audrow/rmf_site/blob/5e376ea789f44311b5e3bfdc10d086c7cd8eb067/rmf_site_format/src/workcell.rs#L392). This sounds unnecessary for what is a single export.

I can be convinced that the early clone to save on all the `.clone()` calls scattered throughout the function can be worth it, but we should at least get rid of the double clone (which should be addressed if you go for the hash map approach).

rmf_site_format/src/workcell.rs
```diff
@@ -280,6 +282,59 @@ pub struct Workcell {
...
 }
 
+impl From<Pose> for urdf_rs::Pose {
+    fn from(pose: Pose) -> Self {
+        urdf_rs::Pose {
+            rpy: match pose.rot {
+                Rotation::EulerExtrinsicXYZ(arr) => urdf_rs::Vec3(arr.map(|v| v.radians().into())),
+                Rotation::Yaw(v) => urdf_rs::Vec3([0.0, 0.0, v.radians().into()]),
+                Rotation::Quat([x, y, z, w]) => {
+                    let (x, y, z) = glam::quat(x, y, z, w).to_euler(EulerRot::XYZ);
```
luca-della-vedova, line 292: I actually don't know but I noticed that in the matching function in [misc.rs](https://github.com/open-rmf/rmf_site/blob/d2fb61f8785e66021b56286fb64b6fda454f9ffe/rmf_site_format/src/misc.rs#L266-L267) we actually apply the rotation in ZYX, not sure what should be done but just thought I'd point out the difference.

rmf_site_editor/src/workcell/save.rs
```diff
@@ -265,6 +265,21 @@ pub fn save_workcell(world: &mut World) {
...
+                    }
+                };
+                match std::io::Write::write_all(&mut std::io::BufWriter::new(f), urdf.as_bytes()) {
+                    Ok(()) => {
+                        info!("Save successful");
+                    }
+                    Err(err) => {
+                        error!("Save failed: {err}");
+                    }
+                }
```
luca-della-vedova, line 282: Is there a way to keep the I/O functions in `rmf_site_format` the same way the saving for other formats is done? Maybe a `workcell.to_urdf_writer(f)`?

luca-della-vedova (commented):
This is looking really great. High level design looks good to me, I went into a more detailed review

luca-della-vedova:
I also just noticed that the pose of links might be wrong without support for joints.
We have a hierarchy of `Frame`s and reading a Frame's `Pose` will define its pose relative to its parent. The issue is that since links in URDF have no parent, if we naively set the link's pose to be the frame's pose, all the information about the hierarchy (i.e. the global transform of the parent) would be lost.

I believe a way to make this work would be to look at each link, its parent, then add a fixed joint between the two at export time, but I'm not sure if there are any other/better ways.

I'm OK with dealing with this in the next step of supporting joints but just thought I'd point that out so we keep it in mind.

rmf_site_format/src/workcell.rs
audrow: I like that much more. Thanks for letting me know!

audrow:
@luca-della-vedova, thanks for the feedback - I feel like I'm learning a lot! I'll iterate on this a bit later this week.

rmf_site_format/src/workcell.rs
audrow: That's awesome - thanks for the tip!

rmf_site_format/src/workcell.rs
audrow: It's amazing that it doesn't need a type here and that it infers it from the usage later in the code. Magic.

rmf_site_format/src/workcell.rs
audrow: I switched to use zyx in c8ef823.

rmf_site_format/src/workcell.rs
audrow: Done in c8ef823.

rmf_site_format/src/workcell.rs
audrow: Done in 77883bd.

rmf_site_format/src/workcell.rs
```diff
@@ -18,16 +18,18 @@
 use std::collections::{BTreeMap, HashSet};
 use std::io;
 
+use super::misc::Rotation;
```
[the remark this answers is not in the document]
audrow: Done in c8ef823.

rmf_site_editor/src/workcell/save.rs
audrow: Done in cba2ebc.

audrow:
> I'm OK with dealing with this in the next step of supporting joints but just thought I'd point that out so we keep it in mind.

That's a good thing to point out. Let's handle it in another PR.

rmf_site_format/src/workcell.rs
audrow: I tried to remove the unnecessary clone calls in 41e04ff.

rmf_site_format/src/workcell.rs
audrow: I went with a hashmap of vectors for the links and collisions in 41e04ff.

I think this is better as we're not iterating over everything, every time. We just do it once now.

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,129 @@ impl Workcell {
...
+                            izz: 0.0,
+                        }
+                    },
+                    mass: urdf_rs::Mass { value: 0.0 },
+                };
+
+                let parent = &f.1.parent;
+                let collision = parent_to_collisions
+                    .get(parent)
+                    .map(|collisions| collisions.clone())
```
luca-della-vedova, line 429: I think the hierarchy calculation here might be off.
This line (and the same afterwards for visuals), do a lookup in the visual / collision hash map for the parent which is fine.
However, the `parent` variable doesn't contain the link's id but the link's parent id, so I believe this line would assign, to the current link, the visuals / collisions that are children of the link's parent link.
The `parent` should be the frame's ID `f.0` rather than `&f.1.parent`, for readability can just unpack the tuple at the iteration level by changing:

```
.map(|f| {
```

Into

```
.map(|(frame_id, frame) {
```

With this change you should be able to also simplify the HashMap access to move values and remove the clone.
Since in every loop iteration the key you use to access the map is different, you can just move the visuals / collisions out of the map and avoid the cloning as such:

```
let visual = parent_to_visuals
    .remove(frame_id)
    .unwrap_or_default();
```

luca-della-vedova: I noticed that hierarchy is also off in the example in your first post, the visual has a parent id of `0` that corresponds to the ID of the root of the workcell, while in the exported `urdf` its parent is set to the `top_anchor` frame which has id 2.
Now there is a minor challenge in the fact that the root might not be a frame itself but that should be pretty easy to fix by adding a dummy link at the origin

luca-della-vedova:
Another small thing I noticed is that the root level tag is `Robot` instead of `robot`, however this should probably be fixed at the `urdf-rs` level. Did you try to import the exported `urdf` into any software to see whether parsers are case sensitive or not?

rmf_site_format/src/workcell.rs
```diff
@@ -289,6 +344,129 @@ impl Workcell {
...
+                (visual, v.1.parent)
+            })
+            .collect();
+        let mut parent_to_visuals = HashMap::new();
+        visuals_and_parents.iter().for_each(|(visual, parent)| {
+            parent_to_visuals
+                .entry(*parent)
+                .or_insert_with(Vec::new)
+                .push(visual.clone());
+        });
```
luca-della-vedova, line 370: This (and the snippet below) is still a bit inefficient since you are iterating over a vector to create another vector, then you iterate over the new vector to create a hash map, also cloning each element twice in the process.

I think you can just skip the creation of the first vector altogether and compress the logic in a single iteration that creates the hashmap, something like:

```
        let mut parent_to_visuals = HashMap::new();
        for (parent, visual) in workcell.visuals.iter() {
            let visual = &visual.bundle;
            let visual = urdf_rs::Visual {
                name: Some(visual.name.clone()),
                origin: visual.pose.into(),
                geometry: visual.geometry.clone().into(),
                material: None,
            };
            parent_to_visuals
                .entry(*parent)
                .or_insert_with(Vec::new)
                .push(visual);
        }
```

This should remove the creation of the intermediate vector, as well as removing the additional `visual.clone()` to just move the value.

audrow:
> Another small thing I noticed is that the root level tag is `Robot` instead of `robot`, however this should probably be fixed at the `urdf-rs` level. Did you try to import the exported `urdf` into any software to see whether parsers are case sensitive or not?

You were right. `rviz2` needs lowercase.

rmf_site_format/src/workcell.rs
audrow: This should be fixed now in 6c5999c.

rmf_site_format/src/workcell.rs
audrow: Should be fixed in 6c5999c.

audrow:
I updated the `urdf` in the PR's description.

It still doesn't load into Rviz, but I think it will once we add joints. 

I think that the visual and collisions are being included correctly now.

luca-della-vedova:
> > Another small thing I noticed is that the root level tag is `Robot` instead of `robot`, however this should probably be fixed at the `urdf-rs` level. Did you try to import the exported `urdf` into any software to see whether parsers are case sensitive or not?
> 
> You were right. `rviz2` needs lowercase. I added a string replace [6c5999c](https://github.com/open-rmf/rmf_site/commit/6c5999cd5549a69c367815435c9e8d4998cc592a) and made an issue on the `urdf_rs` repo.

It was a one-liner and they already fixed and re-released the crate, `0.7.3` contains the fix. I removed the workaround and set the minimum crate version to 0.7.3 in https://github.com/open-rmf/rmf_site/pull/155/commits/0e22813d9c1f4b9bab368381e891c9f2136ac862

luca-della-vedova (approved):
I think this is a good first step, we can tackle the remaining features in followups

audrow:
@luca-della-vedova, great! I don't have merge access. Feel free to merge if you think it's ready.

audrow:
Closing in favor of #177.

PR #159, opened 2023-08-11
Fix compilation for rmf_site_format without the bevy feature

luca-della-vedova:
### Fixed bug

Fixes #158 

### Fix applied

Fix compilation without the `bevy` feature for `rmf_site_format`, as well as add workflows for all supported platforms to make sure the issue doesn't reappear.

PR #160, opened 2023-08-11
Support build with colcon-cargo

Yadunund:
Requires #159


mxgrey (approved):
I think we can go ahead and merge these even if the upstream requirements aren't met. The files don't have any negative impact in the meantime.

PR #164, opened 2023-08-17
Add level fiducials to reference drawing features

luca-della-vedova:
### Fixed bug

This PR fixes map alignment when importing legacy projects that include a list of features / constraints but no fiducials, a typical case is a building with only one level (hence no fiducials) but one or more robot maps (hence the sets of constraints and features).

### Fix applied

Previously, no level anchors were spawned so the optimization ran during map alignment would also transform the reference drawing and find an average of all the drawings. With this PR, level anchors are spawned and added to the fiducials in the position where the reference drawing is. The result is that the reference drawing will be "sticky" and stay in its original position with only the imported robot layers being translated.

I had to increase the scope of the reference drawing site id which required wrapping it in an `Option` to handle the "no reference drawing" case. This added a few (safe) boilerplate unwraps.

Before this PR:

[Screencast from 2023-08-17 10-54-01.webm](https://github.com/open-rmf/rmf_site/assets/9111558/1d7ca83b-37bd-4d17-9d07-6bb73df941aa)

After this PR:

[Screencast from 2023-08-17 12-41-16.webm](https://github.com/open-rmf/rmf_site/assets/9111558/c78e6678-ab05-448f-8fb4-1bc19d2706b0)

### DIFF
--- a/rmf_site_format/src/legacy/building_map.rs
+++ b/rmf_site_format/src/legacy/building_map.rs
@@ -1,7 +1,7 @@
 use super::{level::Level, lift::Lift, PortingError, Result};
 use crate::{
     alignment::align_legacy_building, Affiliation, Anchor, Angle, AssetSource, AssociatedGraphs,
-    DisplayColor, Dock as SiteDock, Drawing as SiteDrawing, DrawingProperties,
+    Category, DisplayColor, Dock as SiteDock, Drawing as SiteDrawing, DrawingProperties,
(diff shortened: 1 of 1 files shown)


mxgrey (approved):
I just made two tweaks:

* Keep `primary_drawing_id` as a u32 instead of `Option<32>` so we don't need to unwrap so much. Unwrapping should generally be avoided except in rare cases where we are absolutely sure from the immediate context of the code that it will not panic (or that panicking is the only acceptable response to the unwrap failing).
* Store both `primary_drawing_id` and `drawing_tf` inside of `primary_drawing_info` so we don't need to recalculate the `drawing_tf` later.

PR #168, opened 2023-08-22
Fix dragging behavior when FPS is low

luca-della-vedova:
### Fixed bug

Starting and finishing dragging is done by two separate systems where:

* System to start drag behavior listens for "button left - just pressed" and initializes dragging.
* System to end drag behavior listens for "button left - just released _and_ we are dragging" and stops dragging.

This can cause an issue in cases when the FPS of the application is very low and the button is pressed and released on the same frame. We can get stuck in dragging behavior with the second system failing to cleanup the dragging.

### Fix applied

This PR changes the drag start system to only start dragging if the button was pressed but not released in the current frame.

### DIFF
--- a/rmf_site_editor/src/interaction/gizmo.rs
+++ b/rmf_site_editor/src/interaction/gizmo.rs
@@ -282,10 +282,13 @@ pub fn update_gizmo_click_start(
         }
     }
 
-    let clicked = mouse_button_input.just_pressed(MouseButton::Left)
+    // Ignore if button was pressed and released in the same frame, to avoid being stuck in
+    // dragging behavior in cases when the frame rate is low.
+    let clicking = (mouse_button_input.just_pressed(MouseButton::Left)
+        && !mouse_button_input.just_released(MouseButton::Left))
         || touch_input.iter_just_pressed().next().is_some();
 
-    if clicked {
+    if clicking {
         if let GizmoState::Hovering(e) = *gizmo_state {
             click.send(GizmoClicked(e));
             if let Ok(Some(intersection)) = intersections.get_single().map(|i| i.position()) {


rmf_site_editor/src/interaction/gizmo.rs
```diff
@@ -282,10 +282,13 @@ pub fn update_gizmo_click_start(
         }
     }
 
-    let clicked = mouse_button_input.just_pressed(MouseButton::Left)
+    // Ignore if button was pressed and released in the same frame, to avoid being stuck in
+    // dragging behavior in cases when the frame rate is low.
+    let clicking = (mouse_button_input.just_pressed(MouseButton::Left)
+        && !mouse_button_input.just_released(MouseButton::Left))
         || touch_input.iter_just_pressed().next().is_some();
```
mxgrey, line 289: Is it possible to have a similar problem with the touch input?

We aren't trying very hard to support touch inputs so I'm fine with ignoring this for now, it's just a thought that occurred to me.

mxgrey (approved):
Oh very good catch. I experienced this one time and had no idea what caused it. Then I was never able to reproduce it, so I needed to move on and prioritize other issues.

rmf_site_editor/src/interaction/gizmo.rs
luca-della-vedova: I am honestly not sure since I don't have a touch device handy. The system in charge of disabling dragging does actually not even listen to touch inputs so I'm not sure how that would have worked

mxgrey: That seems problematic. Maybe we should just remove touch input here to be safe.

mxgrey (approved):

PR #175, opened 2023-08-29
Fix level saving when multiple workspaces are open

luca-della-vedova:
### Fixed bug

Originally reported by @arjo129, when multiple workspaces are open site saving generates corrupted levels.
Specifically, a `HashMap` of levels is populated and entities update it based on its `SiteID` value. However, multiple entities can have the same `SiteID`, as long as they are in different workspaces, hence the level data of the saved site might be silently overwritten by entities that are children of a level with the same `SiteID` that however belongs to a different site.

Steps to reproduce:
* Open a map, for example the demo map.
* Create a new workspace by clicking on File -> New.
* Draw something
* Save the workspace.

The resulting save file will be corrupted and contain a mix of data from the first and the second workspace. This should be addressed by the PR.

### Fix applied

Instead of checking for each entity whether it is the parent of the requested level, iterate through the children of the site / level. This made the code a bit simpler (we don't need to query for the parent of each entity) and possibly slightly more efficient (we only iterate through the entities that we care about, not all entities that currently exist across all workspaces).

### DIFF
--- a/rmf_site_editor/src/site/save.rs
+++ b/rmf_site_editor/src/site/save.rs
@@ -259,7 +259,8 @@ fn generate_levels(
     site: Entity,
 ) -> Result<BTreeMap<u32, Level>, SiteGenerationError> {
     let mut state: SystemState<(
-        Query<(&Anchor, &SiteID, &Parent)>,
+        Query<&Children, With<NameOfSite>>,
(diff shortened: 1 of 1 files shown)


arjo129 (approved):
Tested this. It works as expected.

PR #177, opened 2023-08-31
Urdf import, export and workcell editor fixes

luca-della-vedova:
### Implemented feature

Based on #155 by @audrow, extends urdf export to also export joint and inertial data, as well as add a urdf import feature to convert URDF into a `Workcell` object. Having both allows adding an integration test that performs a roundtrip conversion in `rmf_site_format` and makes sure that the files match.
I am also refining various parts of the workcell editor mode and fix issues I found along the way. Will change it to in review and add a more detailed description of the changes once it's ready.

* Removed rapier dependency, it was meant to be used for physics simulation but it was effectively never used, we can reintroduce it when needed. Saves us some compile time and runtime overhead.
* Fix mesh primitives that were not selectable.
* Add a system that "explodes" SDFs imported in the workcell editor mode to unpack all the visuals and collisions contained in them.
* Fix parenthood of models that was not set properly.
* Add joint components / data to the `Workcell` structure, including conversion back and forth from `urdf`.
* Add a basic joint inspection widget, as well as a button to create a joint when two frames are connected directly in the hierarchy without a joint inbetween.
* Update the mesh primitive inspector widget to only have read-only labels, until we implement systems to update primitives based on sizes / type.
* Add a keybinding (Ctrl-E) to export to URDF. For the moment I didn't add a menu button since that would have needed adding a state dependent logic to the top menu bar.
* Add inertial data to the `Workcell` format, including saving and loading to a site. No inspector is implemented yet though.
* Added a temporary workaround system that replaces `NameInSite` components for `NameInWorkcell` if the spawned object is a child of a `Workcell`. This will probably be removed in the future by having a `NameInGroup` component that unifies the two.
* Add support for opening `urdf` files through the `Open File` window.


luca-della-vedova:
Setting this to ready for review. There are likely new issues to come up with further testing for the workcell editor mode but keeping adding them here might make the PR grow to an even more unmanageable size.

rmf_site_format/src/workcell.rs
```diff
@@ -99,23 +107,115 @@ pub struct Inertial {
...
+        Self {
+            xyz: urdf_rs::Vec3(axis.0.map(|v| v as f64)),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Joint {
     pub name: NameInWorkcell,
+    pub joint_type: JointType,
```
mxgrey, line 194: I think we're missing a huge opportunity here to take advantage of Rust enums.

Every different joint type is potentially going to have very different property fields. For single-dimensional joint types like prismatic and revolute you can mostly reduce the properties down to a single set of joint limits and a single axis, but other joint types, such as a [universal joint](https://en.wikipedia.org/wiki/Universal_joint) may require you to define two or more axes, and multiple dimensions of joint limits. There may be arbitrarily exotic joints that we want to support in the future.

So I'd recommend we do something more like this:

```
pub struct Joint {
    pub name: NameInWorkcell,
    pub properties: JointProperties,
}

pub enum JointProperties {
    Fixed,
    Prismatic(SingleDofJoint),
    Revolute(SingleDofJoint),
    Continuous(SingleDofJoint),
}

pub struct SingleDofJoint {
    pub limit: Option<JointLimit>,
    pub axis: JointAxis, // This should be required, there's no reasonable default
}
```

Then we can add more variants at any time without disrupting the existing single-DoF variants.

rmf_site_format/src/workcell.rs
```diff
@@ -89,7 +90,14 @@ pub struct Mass(f32);
 
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
 #[cfg_attr(feature = "bevy", derive(Component))]
-pub struct Inertia {}
+pub struct Inertia {
```
mxgrey, line 93: Having structs named `Inertial` and `Inertia` doesn't make a lot of sense imo.

I'd recommend changing the current `Inertial` to `Inertia` and change this `Inertia` struct to [`Moment`](https://en.wikipedia.org/wiki/Moment_of_inertia).

In general, the word "inertia" is meant to refer to a body's resistance to acceleration. So it's reasonable to bundle up mass, center of mass, and moment of inertia into one `Inertia` struct. "**Moment** of inertia" refers to resistance to rotational acceleration, so I think it makes sense to use "moment" as its keyword.

rmf_site_format/src/workcell.rs
```diff
@@ -89,7 +90,14 @@ pub struct Mass(f32);
...
+    pub ixx: f32,
+    pub ixy: f32,
+    pub ixz: f32,
+    pub iyy: f32,
+    pub iyz: f32,
+    pub izz: f32,
+}
 
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
 #[cfg_attr(feature = "bevy", derive(Bundle))]
```
mxgrey, line 103: GitHub isn't letting me comment on the `origin: Pose` line, so this comment is a little bit off the mark.

I'd gently suggest changing the field name `origin` to `center` so that its relationship to "center of mass" is more clear. I don't consider this to be a big deal either way, but I think people will generally be more familiar with the concept of a "**center** of mass" than an "**origin** of the Inertial frame".

mxgrey (changes requested):
I've left some comments about the format that I think should help keep it stable as we continue to iterate on workcell related features.

Here are some additional thoughts about the existing format that isn't touched by changes in this PR so I can't leave in-line comments:
- I recommend changing the name of `MeshPrimitive` to `PrimitiveShape`
- Why does `Geometry::Mesh` use `filename: String` instead of `source: AssetSource`?

rmf_site_format/src/workcell.rs
luca-della-vedova: https://github.com/open-rmf/rmf_site/pull/177/commits/53357a6cf5cdf9ec2cb9327d11f0437b104325ac. There is now a new `RangeLimits` enum to contain the different types of limits

rmf_site_format/src/workcell.rs
luca-della-vedova: https://github.com/open-rmf/rmf_site/pull/177/commits/9405bdc2192af3290d99cdc2ccff0cbdff5f2ea7 for Inertia -> Moment and https://github.com/open-rmf/rmf_site/pull/177/commits/422a610a09b6e392d9242f09110e2ed480753a89 for Inertial to Inertia

rmf_site_format/src/workcell.rs
luca-della-vedova: https://github.com/open-rmf/rmf_site/pull/177/commits/aa70ca6c08b68314759143ce06ff60facd9ad465

luca-della-vedova:
> * I recommend changing the name of `MeshPrimitive` to `PrimitiveShape`

https://github.com/open-rmf/rmf_site/pull/177/commits/b57b53f85a73f41cee25763e4bad2e58aaf45152

> * Why does `Geometry::Mesh` use `filename: String` instead of `source: AssetSource`?

Good point, I think it was just to make conversion to urdf simpler but on second thought it really doesn't make a lot of sense. Fixed https://github.com/open-rmf/rmf_site/pull/177/commits/bff825832eb143e56abd916135aebd38f73311e7

I also found the issue with broken dependent references and fixed it in https://github.com/open-rmf/rmf_site/pull/177/commits/769d485036d0a19698d6e198aa572cc5f3312620.

There is still non-optimal behavior when flattening SDFs since all the models are assigned to the same parent. This issue actually originates from the sdf loading code since all visuals / collisions [are assigned to the root element](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_editor/src/site/sdf.rs#L202-L205) so there is no real hierarchy parsing there. I suspect we never bumped into this issue so far because all the SDFs we used have always been very simple with no complex hierarchies but this will probably become an issue once we start importing (for example) complex SDF workcells.
My plan was to exploit the fact that Urdf is pretty much a subset of Sdf and unify the code for parsing of both, but I never got around to do it.

audrow:
Hey, I'm getting one of the two errors when I'm trying to open a workcell with a mesh. I'm using the same workcell that I was using when developing the export URDF function.

```
2023-09-21T13:03:21.418787Z  INFO librmf_site_editor::workspace: Opening workcell file
2023-09-21T13:03:21.419368Z ERROR librmf_site_editor::workspace: Failed loading workcell Error("missing field `source`", line: 101, column: 5)
```

or

```
2023-09-21T12:57:54.856592Z  INFO librmf_site_editor::workspace: Opening workcell file
2023-09-21T12:57:54.857053Z ERROR librmf_site_editor::workspace: Failed loading workcell Error("no variant of enum Anchor found in flattened data", line: 70, column: 5)
```

luca-della-vedova:
Sorry forgot to reply to this!

> Hey, I'm getting one of the two errors when I'm trying to open a workcell with a mesh. I'm using the same workcell that I was using when developing the export URDF function.
> 
> ```
> 2023-09-21T13:03:21.418787Z  INFO librmf_site_editor::workspace: Opening workcell file
> 2023-09-21T13:03:21.419368Z ERROR librmf_site_editor::workspace: Failed loading workcell Error("missing field `source`", line: 101, column: 5)
> ```
> 
> or
> 
> ```
> 2023-09-21T12:57:54.856592Z  INFO librmf_site_editor::workspace: Opening workcell file
> 2023-09-21T12:57:54.857053Z ERROR librmf_site_editor::workspace: Failed loading workcell Error("no variant of enum Anchor found in flattened data", line: 70, column: 5)
> ```
> 

The file format changed, among other things the mesh filename changed from a `String` to an `AssetSource` and `MeshConstraint` is gone for normal anchors, which might cause the errors above. Sadly the only way to fix that right now is to recreate a project.

> Also, I still don't see the "Export URDF" button in the "File" menu.

True, I still need to do the refactor to only show some buttons in some states, didn't do it yet to avoid making this PR even more complex than it already is since that would touch the code in many areas that are not really workcell specific

audrow:
Okay, let me know if I can help with this.

luca-della-vedova:
There are a few rough edges here and there but this is "stable enough" that it should be pretty much OK.

mxgrey (approved):
I know this has been baking for quite a while and much of the implementation details are subject to change as we obtain better foundational tools in Bevy, so I'm only doing a cursory review to unblock this.

I only noticed a few things that seem off, and they're relatively minor, so I've opened issues #198 #199 #200.

PR #178, opened 2023-09-04
Remove add_children uses

luca-della-vedova:
### Implemented feature

Remove all uses of `add_children` throughout the codebase.

In preparation for the upgrade to bevy 0.11, the add_children function was [removed](https://github.com/bevyengine/bevy/pull/6942). This PR refactors the code to remove all uses to make migration easier without breaking the code (yet!). Two main cases were encountered:

* Cases in which the return value was actually ignored, a simple `add_children` -> `with_children` change works.
* Cases in which the return value was used, usually to spawn entities and return their id, remove the `add_children`, spawn all the entities and manually add them through either `add_child` or `set_parent`, depending on the case.

### DIFF
--- a/rmf_site_editor/src/interaction/anchor.rs
+++ b/rmf_site_editor/src/interaction/anchor.rs
@@ -46,7 +46,8 @@ pub fn add_anchor_visual_cues(
             _ => site_assets.site_anchor_mesh.clone(),
         };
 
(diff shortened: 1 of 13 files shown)


luca-della-vedova:
At this point it's probably easier to just work on #180 instead, closing this

PR #180, opened 2023-09-05
Update to Bevy 0.11

luca-della-vedova:
### Implemented feature

This PR performs basic migration to Bevy 0.11. There are a lot of cleanups, refactors and additional features we could do in this upgrade that have been left out for the sake of simplicity, for example:

* 16 size limit on `SystemParam` has been removed, we can remove all the manual splitting we had to do in `widgets/mod.rs`. Done in https://github.com/open-rmf/rmf_site/pull/180/commits/2060bdf9d15f43920391cc5f2442ae6da1419846
* A new [crate](https://github.com/iiYese/aery) for entity relations is available. We could potentially cleanup a lot of the `<Entity>` components, such as `Point`, `Edge`, `MeshConstraint`, etc.
* States were simplified, we could potentially refactor them to reduce duplication that was introduced when `WorkcellEditor` was added. Done in https://github.com/open-rmf/rmf_site/pull/180/commits/26c4a360a2531d67f950cbe30de91f8dee665d22, can be reverted if necessary.
* The [EntityCommand trait](https://docs.rs/bevy/0.10.1/bevy/ecs/system/trait.EntityCommand.html) has been added, we could use it to unify our entity spawning logic and only spawning all the necessary components in one place.
* `Visibility` component has now a new variant for "always visible". To preserve behavior I'm not using it but it seems it might help in a few places (i.e. drawing_editor has a todo mentioning it).
* WebGPU support has been introduced, although it seems it's not supported in many browsers.

Other than the `Visibility` change that also introduced a lot of boilerplate, the other parts can imho be addressed in followup PRs to avoid the complexity of this exploding dramatically.

Other than what was necessary to make the code compile, a few other changes have been made:

* Remove `Label`. It was previously mainly used for fiducials, however the use was removed in favor of `Affiliation`. It was still present for `Measurements` but not really used and set to `None` by default. The reason this had to be done is that now the bevy prelude includes a `Label` and we would have had to change all our imports statements to avoid name clashes.
* Remove `constraint.rs`. They are not used anymore and were effectively dead code.
* Remove `urdf.rs`. Same as above.
* Remove `bevy_mod_picking` dependency. It was not really used since we do manual raycasting through `bevy_mod_raycast` and intersection checking. Removing it had barely any side effect.
* Fix the `assign_drawing_parent_to_new_measurements`. During refactoring it pretty much became a no-op system and measurements were not saved to new projects.

The rest was just what was needed to make it run, the most complex change is rewriting the app for the new schedule, that might use an extra pair of eyes.

TODO:

- [x] Wasm build currently panics in bevy_render.
- [x] Fix panics when deleting entities.


mxgrey:
Here's a before (left) : after (right) screenshot for this PR:
Screenshot from 2023-10-06 11-54-55
I notice that the coloring became a bit dull, and the specularity has softened. Is that due to some change in Bevy's rendering pipeline or default settings? I know it's subjective but I think I would prefer the previous look which feels more vibrant.

I'll check Bevy's changelog to see if I can find a likely cause. Then I'll try to fiddle with our rendering settings to see if we can get back to the original effect.

mxgrey (approved):

luca-della-vedova:
Found / fixed a migration issue when removing `add_children` that caused wrong fiducial spawning in https://github.com/open-rmf/rmf_site/pull/180/commits/4d66fb887f5710f22da6249bc9efdba4c3ecf2a8.
Tested on web / Linux / Mac and it seems to help, good news is that 0.9 was panicking on Mac and updating to 0.11 fixes it. Merging.

PR #182, opened 2023-09-26
Add package exporter

audrow:
This PR adds 
* a library to export a package from a `urdf_rs::Robot` and
* a binary demo of the package exporter that is run from the CLI.

Currently, this isn't connected to the `rmf_site_editor` since I think it'd be good to get this in and connected after #177 is merged in.

We can also move this anywhere else that we think makes sense.


luca-della-vedova:
Not really a thorough review but I'm trying to understand from the PR what the intended final workflow is.

From what I see, the utility expect its input to be an `urdf` that already has all the meshes in a ros2 package and with a `package://` prefix. However this conversion is really the largest missing piece right now since the site editor can accept files from a variety of sources (i.e. local files, gazebo app server) and is not really able to export them to a ros2 package.
This means that, to my understanding, with this PR we wouldn't really have a full pipeline yet, what would you envision this conversion to be like? Would it be something done by this tool or something done at the site editor level?

audrow:
I think we just focus on the local case for now.

For the local case, the package paths for the assets are updated.

The idea of how this integrates is that you call the `generate_package` function when you want to export a package. This function takes the `urdf_rs::Robot` and some context for the new package to create. It then outputs the package to a location you select.

luca-della-vedova:
My main concern is that with this limitation there seems to be a large missing piece in the pipeline.

This utility takes as an input a urdf that only refers to files in a ROS 2 package, however `rmf_site` has currently no way to generate anything like that, since its files will most likely exist in a variety of places.
So for this utility to be useful we would need to have some way to start from a `.workcell.json`, either through:

1. A capability in the site editor that exports all meshes into ROS2 packages that can be used by this tool (probably not the right way to go, it would be very complex).
2. Having the site editor export a "wrong urdf" as it does now, which contains references to files on fuel, local file system, etc. and can't really be parsed by other tools, just to have this tool ingest it and create an actually correct one.
3. Having this tool (and the future site editor) take `.workcell.json` files as an input, and parse all types of asset sources, and generate a valid package.

In general sounds like assuming the input is a URDF in the right format is very limiting and would force us to either do option 1. (which defeats the purpose of this tool) or option 2. (which relies on this weird intermediate product that is an invalid file).

I believe the more general solution is really 3., where either:
* This capability is added to `rmf_site` and it exports a full package when doing the `export urdf` feature (or a new `export urdf package` feature).
* This tool keeps being separate and it takes a workcell.json as an input.

I have other feedback (mostly on trying to reuse code from libraries rather than reimplementing functionalities) but it's not as important as this architectural decision.

luca-della-vedova:
My main concern is that with this limitation there seems to be a large missing piece in the pipeline.

This utility takes as an input a urdf that only refers to files in a ROS 2 package, however `rmf_site` has currently no way to generate anything like that, since its files will most likely exist in a variety of places.
So for this utility to be useful we would need to have some way to start from a `.workcell.json`, either through:

1. A capability in the site editor that exports all meshes into ROS2 packages that can be used by this tool (probably not the right way to go, it would be very complex).
2. Having the site editor export a "wrong urdf" as it does now, which contains references to files on fuel, local file system, etc. and can't really be parsed by other tools, just to have this tool ingest it and create an actually correct one.
3. Having this tool (and the future site editor) take `.workcell.json` files as an input, and parse all types of asset sources, and generate a valid package.

In general sounds like assuming the input is a URDF in the right format is very limiting and would force us to either do option 1. (which defeats the purpose of this tool) or option 2. (which relies on this weird intermediate product that is an invalid file).

I believe the more general solution is really 3., where either:
* This capability is added to `rmf_site` and it exports a full package when doing the `export urdf` feature (or a new `export urdf package` feature).
* This tool keeps being separate and it takes a workcell.json as an input.

I have other feedback (mostly on trying to reuse code from libraries rather than reimplementing functionalities) but it's not as important as this architectural decision.

luca-della-vedova:
So there are currently a [few asset sources](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_format/src/asset_source.rs#L27-L38)

```
pub enum AssetSource {
    Local(String),
    Remote(String),
    Search(String),
    Bundled(String),
    Package(String),
    OSMTile {
        zoom: i32,
        latitude: f32,
        longitude: f32,
    },
}
```

We can safely ignore `Bundled` and `OSMTile` and for the sake of a first implementation we can also ignore `Search`.
A good reference on how to parse these variants is in [site_asset_io.rs](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_editor/src/site_asset_io.rs#L274).

* `Package` is pretty much what you have done for your package variants, just that I would suggest using the `urdf_rs` function to get the path as is done [here](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_editor/src/site_asset_io.rs#L314C30-L314C49) instead of writing your own.
* `Local` should just be a [path relative to the file that was loaded](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_editor/src/site_asset_io.rs#L297), so it's quite straightforward.
* `Remote` is a bit tricky but we can do a simple approach for a first iteration. The site editor will save all downloaded files to a [cache path](https://github.com/open-rmf/rmf_site/blob/fbc5fbf4dbf416f5b9686ea03bcf41b7545a0902/rmf_site_editor/src/site_asset_io.rs#L20-L25), if you are on Linux that will be under `~/.cache/open_robotics/rmf_site_editor`. This is probably easier than trying to fetch them from a remote server, where we would also have to handle API keys if the models were private.

It is also true that if the utility was to be bundled in `rmf_site` you wouldn't actually have to do any of this parsing, you should be able to use [load_path](https://docs.rs/bevy/0.9.0/bevy/asset/trait.AssetIo.html#tymethod.load_path) to do:

```
// Convert AssetSource object to Path first.
asset_server.asset_io().load_path(path);
```

Which would return you a vector of bytes that you could then process however you want.

audrow:
Closed to prefer https://github.com/open-rmf/rmf_site/pull/186.

PR #185, opened 2023-10-05
Add a Cargo.lock with a set of working packages

luca-della-vedova:
### Fixed bug

`main` is currently not working because of a change in `async-executor` from version 1.5.1 to version 1.5.3 that causes the app to freeze.
Since the dependency has the same minor version, it is automatically downloaded when building the package.

### Fix applied

Commit a `Cargo.lock` with a frozen version to 1.5.1, going forward we should keep it updated when relevant.


arjo129 (approved):

mxgrey:
Since this only introduces a lockfile, the problem will happen again the next time we need to `cargo update`, e.g. to migrate to 0.11.

I believe we'll also need to fix our selection of dependency versions in the `Cargo.toml`s.

PR #186, opened 2023-10-09
Export package from a workcell

audrow:
This PR adds 
* a library to export a package from a workcell and
* a binary demo of the package exporter that is run from the CLI.

This relies on #177, since it uses code to get the URDF from a workcell. Perhaps we retarget this PR to main once #177 is merged in.

With multiple binaries, you now need to use `--bin` to select the executable. If we like things where they are, I can update the README to add this change.


audrow:
Note, I'm having trouble viewing the generated URDF. I'll take a look at this in the coming days.

package_exporter/src/generate_package.rs
```diff
@@ -0,0 +1,195 @@
...
+) -> Result<(), Box<dyn Error>> {
+    let asset_paths = get_mesh_paths(&workcell)?;
+    copy_files(&asset_paths, &meshes_directory_path)?;
+
+    let new_workcell = convert_to_package_paths(&workcell, new_package_name, mesh_directory_name)?;
+
+    let urdf_robot = new_workcell.to_urdf()?;
+    let urdf_file_path = urdf_directory_path.join("robot.urdf");
+    let urdf_string = urdf_rs::write_to_string(&urdf_robot)?;
+    std::fs::write(urdf_file_path, urdf_string)?;
```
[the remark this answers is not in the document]
luca-della-vedova: Any update on this? I wonder if https://github.com/openrr/urdf-rs/issues/87 could have been the cause? There was a missing namespace issue. I guess can try to update to the latest `urdf-rs` and see if it helps

package_exporter/src/generate_package.rs
```diff
@@ -0,0 +1,195 @@
...
+use std::io::{Error as IoError, ErrorKind as IoErrorKind};
+use crate::template;
+
+pub fn generate_package(
+    workcell: &Workcell,
+    package_context: &template::PackageContext,
+    output_directory: &String,
+) -> Result<(), Box<dyn Error>> {
+
+    let output_directory = "output".to_string();
```
luca-della-vedova, line 14: Since you redefine `output_directory` here the argument is effectively ignored and the output folder is always `output` regardless of what the user specifies.

package_exporter/src/generate_package.rs
```diff
@@ -0,0 +1,195 @@
...
+        if let Geometry::Mesh { source: asset_source, scale } = &mut visual.bundle.geometry {
+            let path = get_path_to_asset_file(asset_source)?;
+
+            let file_name = PathBuf::from(&path).file_name().ok_or(
+                IoError::new(IoErrorKind::InvalidInput, "Unable to get file name from path")
+            )?.to_str().ok_or(
+                IoError::new(IoErrorKind::InvalidInput, "Unable to convert file name to str")
+            )?.to_owned();
+
+            let package_path = format!("package://{}/{}/{}", package_name, mesh_directory_name, file_name);
```
luca-della-vedova, line 88: Since the asset path already contains `package://`, this instruction seems to prepend another `package://` string, paths look like this on my machine: `package://package://package_name/path`, hence rviz is not able to resolve them.
```suggestion
            let package_path = format!("{}/{}/{}", package_name, mesh_directory_name, file_name);
```

demo_package_exporter/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    if args.len() != 3 {
+        eprintln!("Usage: {} <workcell_file> <output_directory>", args[0]);
+    }
+    let workcell_file_path = &args[1];
+    let output_directory = &args[2];
+
+    let package_context = PackageContext {
+        project_name: "output_package".to_string(),
```
luca-della-vedova, line 13: This hardcodes the package name to `output_package` while the rest of the executable uses `test_package`. As a consequence paths look like `package://test_package/[...]` which cannot be resolved.

```suggestion
        project_name: "test_package".to_string(),
```

luca-della-vedova (commented):
I found out why rviz is not visualizing them, it's mostly issues with the package.xml having the wrong package name and the mesh paths being in the wrong format.
I left comments in the specific lines with details on how I fixed them on my machine, I don't have `rviz` ready but I tried with [urdf-viz](https://github.com/openrr/urdf-viz) and no mesh is visualized with the current code but "something that mostly makes sense" is displayed after the suggested changes.

We should probably have this as an option in the site editor rather than a separate executable, after all the current "Export urdf" button is not too useful if the urdf contains a bunch of references to assets on fuel or on the local filesystem that will not be able to be resolved when loaded in Rviz.

demo_package_exporter/src/main.rs
audrow: Should be fixed in c822bd5.

package_exporter/src/generate_package.rs
audrow: Should be fixed in c822bd5.

package_exporter/src/generate_package.rs
audrow: Should be fixed in c822bd5.

audrow:
Closing in favor of #187.

PR #187, opened 2023-11-10
Export package instead of URDF

audrow:
This PR adds the business logic from #186 into `rmf_site_editor`.


rmf_site_editor/src/package_exporter/generate_package.rs
```diff
@@ -0,0 +1,236 @@
...
+                "Unable to convert asset_path to str",
+            ))?
+            .to_string())
+    } else if let AssetSource::Local(path) = asset_source {
+        Ok(path.clone())
+    } else {
+        Err(IoError::new(
+            IoErrorKind::Unsupported,
+            "Not a supported asset type for exporting a workcell to a package",
+        ))?
```
luca-della-vedova, line 187: This would be a bit more idiomatic written as a `match` statement rather than an `if let` chain, the unimplemented branch should probably also have all the variants written explicitly so we can make it a compile error if in the future a variant is added, i.e.:

```
match asset_source {
    AssetSource::Package(_) => [...],
    AssetSource::Remote(asset_name => [...],
    AssetSource::Local(path) => [...],
    AssetSource::Search(_) | AssetSource::OSMTile(_) | [...] => return the error
}
```

rmf_site_editor/src/workcell/save.rs
```diff
@@ -314,23 +302,71 @@ pub fn save_workcell(world: &mut World) {
...
+        .to_string();
+
+    let package_context = PackageContext {
+        license: "TODO".to_string(),
+        maintainers: vec![Person {
+            name: "TODO".to_string(),
+            email: "TODO".to_string(),
+        }],
+        project_name: package_name,
+        fixed_frame: "base_link".to_string(),
```
luca-della-vedova, line 363: We should probably change this to `world` since that's what we use as the root frame.

rmf_site_editor/src/workcell/save.rs
```diff
@@ -314,23 +302,71 @@ pub fn save_workcell(world: &mut World) {
...
+        .to_str()
+        .ok_or("Invalid file name")?
+        .to_string();
+
+    let package_context = PackageContext {
+        license: "TODO".to_string(),
+        maintainers: vec![Person {
+            name: "TODO".to_string(),
+            email: "TODO".to_string(),
+        }],
```
luca-della-vedova, line 361: Imho it would be nice if maintainer and version were prepopulated to something that at least makes the package compile (i.e. version 0.0.1, some placeholder name and email) so it's a bit easier to iterate on packages. Currently since these fields are not editable by the user testing a new package requires opening the `package.xml` and updating these values to make it buildable

luca-della-vedova (commented):
Very nice, just a few minor changes here and there but I think this is pretty much good to go

luca-della-vedova:
Another note, I'm a bit worried about the signoff since this repo requires all commits to be verified, can you check if there is any issue with your GPG key then force-push? It shows `Unverified` next to each of your commits.

rmf_site_editor/src/workcell/save.rs
audrow: Should be fixed in ee15eac.

rmf_site_editor/src/workcell/save.rs
audrow: Should be fixed in ee15eac.

rmf_site_editor/src/package_exporter/generate_package.rs
audrow: Should be fixed in ee15eac.

rmf_site_editor/src/lib.rs
```diff
@@ -19,6 +19,8 @@ pub mod save;
 use save::*;
 pub mod widgets;
 use widgets::{menu_bar::MenuPluginManager, *};
+pub mod package_exporter;
+use package_exporter::*;
```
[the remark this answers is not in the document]
audrow: Should be fixed in ee15eac.

rmf_site_editor/src/package_exporter/template.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+pub fn populate_and_save_templates(
+    templates: Vec<Template>,
+    context: &PackageContext,
+) -> Result<(), Box<dyn Error>> {
+    let context = tera::Context::from_serialize(context)?;
+    let mut tera = Tera::default();
+    for template in templates.iter() {
+        let content = std::fs::read_to_string(&template.path)?;
```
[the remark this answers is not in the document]
audrow: I would expect that this solves this. Any ideas?

https://github.com/open-rmf/rmf_site/blob/02b694a73103c7b07cd665c4ecc826d974d87b03/rmf_site_editor/src/workcell/urdf_package_exporter/generate_package.rs#L203

audrow:
> Another note, I'm a bit worried about the GPG signing since this repo requires all commits to be verified, can you check if there is any issue with your GPG key then force-push? It shows `Unverified` next to each of your commits.

All verified now!

luca-della-vedova:
I can't push changes to the branch so I'll just merge this in and push directly to the feature branch. For reference this is the commit that fixed the file issue (using the `include_bytes!` macro) as well as a new issue that popped up during refactor where `package://` paths were not being generated correctly.

PR #190, opened 2023-11-17
Add more state based menus

luca-della-vedova:
### Implemented feature

Followup from #189, add the logic to the rest of the menus, the most significant change is in the `View` menu.
Now all menu entries that are only relevant to site editor mode will only be visible in that mode, making the workcell editor mode a lot cleaner.
I also added a new entry, only for workcell editor, to toggle the origin reference arrow. I wanted to add an infinite grid spawning but it caused the anchors outline to be stuck in a displayed state.

Example of View menu in workcell editor mode:

Before, a lot of noise:
image

After, only what is relevant:
image

A new `ViewMenuPlugin` has been added and it populates all the buttons in the view menu, with their respective target states.

I was a bit on the fence on whether to group plugins by submenu (i.e. `ViewMenuPlugin` that includes all the buttons) or by state (i.e. adding a `WorkcellMenuPlugin` that adds all the workcell related buttons, a `SiteMenuPlugin` that adds all the site buttons and a generic `Common(?)MenuPlugin` that adds all the buttons that apply to all states).
I went for the first approach but happy to refactor if another way seems better.


arjo129 (approved):
Works for me.

PR #191, opened 2023-11-23
Make workspace save not blocking

luca-della-vedova:
### Implemented feature

This PR implements a long standing TODO that came back at us, file saving dialogs were blocking instead of async (like the loading ones).
This froze the background app in Linux / Windows which is a minor annoyance but completely froze the app in Mac which is a deal breaker.

This PR updates rfd as a precursor for `wasm` save support (a new `TODO` documents what would be necessary to get the feature in), makes file save dialogs run in a task pool, as well as remove dead code and refactor the `Save` plugin inside the `Workspace` plugin, so now the `Workspace` plugin includes all workspace related functions (saving, creating and loading).

Something similar will have to be done for the `Browse` window when creating new models, but I left that out of this implementation for now to avoid making the complexity explode.


Yadunund (approved):
Tested and works great!

PR #201, opened 2024-01-31
Add a few more properties to legacy navgraphs

luca-della-vedova:
### Implemented feature

While working on SDF export, I noticed that several properties are missing, this PR brings us "closer" to all the features adding doors, lifts and orientation constraints. It also updates the TODOs to bring them up to date with what is needed since we will need to revisit this section.

I split this into a standalone PR since the SDF export is already becoming a very large change and this is somewhat orthogonal so it should be easier to review (and probably it would be better to scrutinize this a bit closer, since mistakes in an exported SDF are a lot easier to spot than mistakes in an exported navgraph).
I'll only be able to do full end to end testing when this is integrated with the SDF branch and we can export a full demo world, so this is somewhat experimental code (and parts like the lift center calculation might use another pair of eyes)

### DIFF
--- a/rmf_site_format/src/lift.rs
+++ b/rmf_site_format/src/lift.rs
@@ -139,7 +139,7 @@ impl LiftProperties<u32> {
         let right_anchor = site.get_anchor(self.reference_anchors.right())?;
         let left_trans = left_anchor.translation_for_category(Category::Level);
         let right_trans = right_anchor.translation_for_category(Category::Level);
-        let yaw = (left_trans[1] - right_trans[1]).atan2(left_trans[0] - right_trans[0]);
+        let yaw = (left_trans[0] - right_trans[0]).atan2(left_trans[1] - right_trans[1]);
         let midpoint = [
             (left_trans[0] + right_trans[0]) / 2.0,
             (left_trans[1] + right_trans[1]) / 2.0,
--- a/rmf_site_format/src/lift.rs
+++ b/rmf_site_format/src/lift.rs
@@ -115,6 +115,60 @@ pub struct LiftProperties<T: RefTrait> {
     pub initial_level: InitialLevel<T>,
 }
 
+impl LiftProperties<u32> {
+    /// Returns the pose of the lift cabin center in global coordinates.
+    pub fn center(&self, site: &Site) -> Option<Pose> {
+        // Center of the aabb
+        let center = match &self.cabin {
+            LiftCabin::Rect(params) => {
+                let front_door_t = params
+                    .front_door
+                    .as_ref()
+                    .map(|d| d.thickness())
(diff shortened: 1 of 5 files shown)


luca-della-vedova:
I added two small commits https://github.com/open-rmf/rmf_site/pull/201/commits/d76f218b70ebdf9aa1c9bcf0259a9a1e6ce2652e and https://github.com/open-rmf/rmf_site/pull/201/commits/3e9497484ca3d0ec9d9098800ad7810d0456d753 that fix two minor issues I encountered with the office world that made it not work (multiple locations were created with the same `<Unnamed>` name, as well as lanes were missing).
With these changes and the export_sdf branch the office demo almost works.

luca-della-vedova:
With https://github.com/open-rmf/rmf_site/pull/201/commits/b4d0644cc2cf52c2e4aa9ab3d2d735c18843d2ac now the navgraph contains lift properties for vertices and lanes that contain lift anchors. On end to end demos robots can now use lifts.

rmf_site_format/src/lift.rs
```diff
@@ -115,6 +115,60 @@ pub struct LiftProperties<T: RefTrait> {
...
+                    -params.depth / 2.0 - params.thickness() - params.gap() - front_door_t / 2.0,
+                    params.shift(),
+                    DEFAULT_LEVEL_HEIGHT / 2.0,
+                ]
+            }
+        };
+        // Get the vector between the reference anchors
+        let left_anchor = site.get_anchor(self.reference_anchors.left())?;
+        let right_anchor = site.get_anchor(self.reference_anchors.right())?;
+        let left_trans = left_anchor.translation_for_category(Category::Level);
```
mxgrey, line 140: The category here should be `Category::Lift` since we're asking for the translation that a lift would have when defined by this anchor.

mxgrey (changes requested):

luca-della-vedova:
All feedback should be addressed [here](https://github.com/open-rmf/rmf_site/pull/201/commits/877694366834090af439093f722f39cba03c5e58).
I admit not really understanding the `translation_for_category` logic, I tried to export a few legacy buildings (i.e. hotel) to a `.site.ron` format and noticed that the only categorized anchors I could find where under the `General` category.

For the site anchors I was mostly thinking of the use case of a lane being between a site and a level anchor, for example if a user uses a fiducial anchor as the endpoint of a lane, and the second endpoint sits on a level anchor. But anyway since we don't want to allow site anchors because of the level ambiguity I just removed the logic.

mxgrey (approved):
I think the best way to fully validate the PR will be with e2e tests that show the properties are being used as intended by the fleet adapters, and we'll need to merge this PR before we can reach those e2e tests, so we should go ahead with this and apply corrections as needed later.

PR #203, opened 2024-02-05
Add rosdep key and remove dynamic linking by default

### LINKED ISSUE #202: Consider removing dynamic linking optimization
The dynamic linking optimization that we use [here](https://github.com/open-rmf/rmf_site/blob/e852b3465ac41fc56657ad26879b1dd0438c135c/rmf_site_editor/Cargo.toml#L53-L57) does reduce compile times quite significantly (on my machine it went from 55s to 46s), but as noted in the [docs](https://bevyengine.org/learn/book/getting-started/setup/#enable-fast-compiles-optional) it makes redistributing the app more tricky.

Specifically, because (probably?) colcon expects the binary to be statically linked, building the package with colcon and trying to run it results in the following error:

```
[...] error while loading shared libraries: libbevy_dylib-165be36195c0d4d6.so: cannot open shared object file: No such file or directory
```

On the other hand, if we remove the optimization we get longer compile times but `ros2 run` works like a charm.
It's a tradeoff but I believe for greater integration with ROS 2 we should probably drop the dynamic linking (maybe only use it on the local developer workstation if they really want?)

CC @Yadunund @mxgrey for feedback

luca-della-vedova:
### Fixed bug

Fix #202, as well as add rosdep keys.

### Fix applied

ROS 2 doesn't play well with dynamic linking so disabling it by default, while adding a note on the README on how to reeanble it at compile time to speed up development.
I also added rosdep keys to package.xml so users that use ROS to run the workcell editor don't need to install any system dependency manually.

These changes make the site editor fully compatible with rosdep, colcon and ros2 run utilities.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,13 @@ members = [
 # the workspace needs to explicitly specify this.
 # For more see: https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html
 resolver = "2"
+
+# Compile time optimizations as suggested in
+# https://bevyengine.org/learn/quick-start/getting-started/setup/#compile-with-performance-optimizations
+# Enable a small amount of optimization in debug mode
+[profile.dev]
+opt-level = 1
+
+# Enable high optimizations for dependencies (incl. Bevy), but not for our code:
+[profile.dev.package."*"]
+opt-level = 3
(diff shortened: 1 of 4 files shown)


arjo129 (approved):
LGTM

PR #209, opened 2024-03-21
Fix embedded asset path panic

### LINKED ISSUE #208: [Bug]: Bevy 0.12 migration broke colcon workflow

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

Since #206 there is now a panic on startup when running the site editor through `ros2 run`. The panic is internal in a bevy macro [here](https://github.com/open-rmf/rmf_site/blob/7c93231fcc6cb1207c24833469a35d08d4e504ab/rmf_site_editor/src/site/assets.rs#L22):

```
called `Option::unwrap()` on a `None` value
```

I'll be investigating but I believe this might be due to how the embedded_asset macros lookup the path and how colcon relocates the executable.
If all fails will just revert to the previous way of doing things when we implemented pretty much the same functionality ourselves.

luca-della-vedova:
### Fixed bug

Fixes #208.

### Fix applied

This is the upstream bug that tracks the reason for our panic https://github.com/bevyengine/bevy/issues/10377.
It has been fixed and released in 0.13 but we are on 0.12 as of now. As much as I'm always a big fan of updating we are cannot do it quite yet until `bevy_polyline` and `bevy_infinite_grid` are updated, and they seem to always lag about 1 version of bevy.

Luckily there seems to be a workaround in https://github.com/bevyengine/bevy/issues/10377#issuecomment-1858797002 that we can use until we migrate.

To test:

* Build the site editor in a colcon workspace (Note you might need to remove the root Cargo.toml file until https://github.com/colcon/colcon-cargo/pull/25 is solved.
* Try to run it with `ros2 run rmf_site_editor rmf_site_editor`.

Note that I could only test on Linux, if anyone could test on Windows that would also be great but since this is quite severe I wouldn't necessarily hold that as a blocker.

It should panic without this PR and work fine with it.

### DIFF
--- a/rmf_site_editor/src/site/assets.rs
+++ b/rmf_site_editor/src/site/assets.rs
@@ -19,9 +19,21 @@ use crate::{shapes::*, site::*};
 use bevy::{asset::embedded_asset, math::Affine3A, prelude::*};
 
 pub(crate) fn add_site_icons(app: &mut App) {
-    embedded_asset!(app, "icons/battery.png");
-    embedded_asset!(app, "icons/parking.png");
-    embedded_asset!(app, "icons/stopwatch.png");
+    // Taken from https://github.com/bevyengine/bevy/issues/10377#issuecomment-1858797002
+    // TODO(luca) remove once we migrate to Bevy 0.13 that includes the fix
+    #[cfg(any(not(target_family = "windows"), target_env = "gnu"))]
+    {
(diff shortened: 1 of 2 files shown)


arjo129 (approved):
LGTM.

Issue #98, opened 2022-11-27
WASD Controls


mxgrey:
Sometimes moving around using the mouse to navigate is clunky, especially on laptops. It might be good to have something like WASD controls for moving forward/right/back/left.

We just have to avoid conflicting keybindings that may be more useful.

mxgrey:
Done in #218

Issue #97, opened 2022-11-27
Touch Screen Controls


mxgrey:
It ought to be possible to use the site editor in some capacity on touch screens, especially tablets or laptops that have touch screen capabilities. Bevy supports such devices but we are currently not making good use of that ability.

Here are some suggestion for what the controls could be like:
* Single-finger tap (+ optionally drag): same as left click
* Double-finger drag: pan the camera, same as right click + drag
* One finger fixed, second finger drag: pitch/roll the camera, same as middle click + drag
* Double-finger pinch/expand: zoom
* Double-finger twirl: yaw

Issue #75, opened 2022-11-08
Improve pan/tilt/zoom controls


mxgrey:
The current pan and tilt controls for the user camera can get kind of wonky for certain angles or distances from the origin. We should make the behavior more intuitive. Here are a few suggestions:
* Panning should primarily be based on picking any object that happens to be under the cursor and then panning the camera in a way that keeps the same point on that same object under the cursor while dragging.
  * If there is no object under the cursor available to pick, then fall back to picking the point under the cursor that's in the x-y plane
  * If the x-y plane is not visible, then fall back to an imaginary point under the cursor that's some set distance in front of the camera
* Rotation should also primarily be based on picking, i.e. the point picked by the cursor should be dragged under the cursor by the rotation action.
  * If there is no object under the cursor then the dragged point should be an imagine point on a sphere around the camera
  * The center of the rotation should be the camera origin
* Zoom (wheel scroll) should translate the camera forwards or backwards
  * Shift+Scroll can be used to do a traditional scaling-based zoom
* Some egui panel should have a button to reset the camera view.
  * Maybe allow user camera positions to be saved in a list and reused

mxgrey:
Done by #218

PR #218, opened 2024-05-23
Improved camera controls

reuben-thomas:
### Implemented feature

Improving camera controls for the cursor to address https://github.com/open-rmf/rmf_site/issues/75 based on picking

I'm going to refactor the camera_controls module to distinguish commands from the mouse, keyboard (https://github.com/open-rmf/rmf_site/issues/98) and eventually touchscreen (https://github.com/open-rmf/rmf_site/issues/97). This should allow us to separately configure each input type and manage conflicts.


rmf_site_editor/src/interaction/camera_controls/cursor.rs
```diff
@@ -0,0 +1,385 @@
...
+            let denom = Vec3::Z.dot(cursor_ray.direction());
+            if denom.abs() > f32::EPSILON && pitch.abs() >= MIN_SELECTION_ANGLE {
+                let dist = (-1.0 * cursor_ray.origin()).dot(Vec3::Z) / denom;
+                if dist > f32::EPSILON {
+                    return cursor_ray.origin() + cursor_ray.direction() * dist;
+                }
+            }
+
+            // No groundplane intersection
+            // Pick a point of a virtual sphere around the camera, of same radius as its height
```
mxgrey, line 345: Using the height of the camera as the radius of the maximum sphere is an interesting choice. Is there a specific rationale behind that, or is it just based on experimenting with different ideas and seeing what feels best?

reuben-thomas: My intuition was that when the cursor is grabbing nothing, it should behave as if selecting the nearest object, which would be the ground. In hindsight, I think a better way to approximate the nearest object would be to grab the nearest [visible entity](https://docs.rs/bevy/latest/bevy/core_pipeline/core_3d/struct.Camera3dBundle.html#structfield.visible_entities) from the active camera

rmf_site_editor/src/interaction/camera_controls/cursor.rs
```diff
@@ -0,0 +1,452 @@
...
+    mouse_input: Res<Input<MouseButton>>,
+    keyboard_input: Res<Input<KeyCode>>,
+    raycast_sources: Query<&RaycastSource<SiteRaycastSet>>,
+    cameras: Query<(&Projection, &Transform, &GlobalTransform)>,
+    primary_windows: Query<&Window, With<PrimaryWindow>>,
+) {
+    if let Ok(window) = primary_windows.get_single() {
+        // Return if cursor not within window
+        if window.cursor_position().is_none() {
+            *cursor_command = CursorCommand::default();
```
mxgrey, line 72: This line causes a jarring drop out of the control mode if the cursor ever leaves the screen space.

I've opened a PR targeting your fork with a fix that I would recommend: https://github.com/reuben-thomas/rmf_site/pull/1

rmf_site_editor/src/interaction/camera_controls/keyboard.rs
```diff
@@ -0,0 +1,351 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
```
mxgrey, line 2: Nitpick: 2024

rmf_site_editor/src/interaction/camera_controls/keyboard.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        let mut target_zoom_motion = 0.0;
+        if keyboard_input.pressed(KeyCode::N) {
+            target_zoom_motion += -1.0;
+        }
+        if keyboard_input.pressed(KeyCode::M) {
+            target_zoom_motion += 1.0;
+        }
+        // Temporarily use M, N as zoom keys since +-, <>, [] cannot be detected with shift
+        // keyboard_input.get_pressed().for_each(|key| {
```
mxgrey, line 102: Looks like this is dead commented code, so it can be removed?

mxgrey (changes requested):
This is really awesome! The camera interaction is much smoother and more intuitive with these improvements.

I'm just leaving a few small comments, in addition to the things we've already discussed offline.

rmf_site_editor/src/interaction/camera_controls/keyboard.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+        if keyboard_input.pressed(KeyCode::Right) {
+            target_keyboard_motion.x += 1.0;
+        }
+        if target_keyboard_motion.length() > 0.0 {
+            target_keyboard_motion = target_keyboard_motion.normalize();
+        }
+
+        let mut target_zoom_motion = 0.0;
+        if keyboard_input.pressed(KeyCode::N) {
```
mxgrey, line 113: I think `M` and `N` are likely to be used for other hotkeys. I'd suggest changing this to `PgUp` and `PgDn`.

rmf_site_editor/src/interaction/camera_controls/mod.rs
```diff
@@ -481,73 +489,70 @@ fn camera_controls(
...
-
-            if changed {
-                // emulating parent/child to make the yaw/y-axis rotation behave like a turntable
-                // parent = x and y rotation
-                // child = z-offset
-                let rot_matrix = Mat3::from_quat(persp_transform.rotation);
-                persp_transform.translation = controls.orbit_center
-                    + rot_matrix.mul_vec3(Vec3::new(0.0, 0.0, controls.orbit_radius));
+        // Panning
+        } else if (cursor_command.command_type == CameraCommandType::Pan) {
```
mxgrey, line 547: The compiler is flagging this with a warning about unnecessary parentheses.

rmf_site_editor/src/interaction/camera_controls/keyboard.rs
arjo129: Knit: Smaller keyboards may not have `PgUp` and `PgDn`.

mxgrey: I think there's usually an equivalent shortcut, often using the `Fn` key.

In the long-term we should really have a hotkey register that lets people customize what keys are used for hotkeys, but in the meantime I think assigning the zoom in/out feature to keys that aren't always available is okay. If someone doesn't have `PgUp`/`PgDn` they will almost certainly still have some way to scroll, and they'll be able to use that for zooming.

mxgrey (approved):
Thanks for these huge UX improvements!

PR #219, opened 2024-06-03
Use site ID to ensure unique file names

mxgrey:
This uses Site ID values in file names for the following benefits:
1. The filenames will be the same even as the names of entities change
2. The correct filenames will be produced even if duplicate names exist among different models


luca-della-vedova:
This breaks the level transparency plugin since it uses model names, looking into it

luca-della-vedova:
Regression should be fixed by https://github.com/open-rmf/rmf_site/pull/219/commits/2872ab8c7bfebe7fbfa44b210aae2af89dcb5494, merging

luca-della-vedova (approved):

PR #220, opened 2024-06-06
Add main menu option for creating new site

mxgrey:
The main purpose of this PR is to support creating new sites from scratch from the main menu. However I'm sneaking in a bug fix and a quality of life improvement that I came up with while testing the process of building sites from scratch:
* If a list of groups (e.g. fiducials, textures) is being filtered in the selection drop-down menu, include a `more...` item in the drop down menu that clears the filter
* Fix the level spawning UI to add all the components that are needed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5833,3 +5833,7 @@ name = "zeroize"
 version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "525b4ec142c6b68a2d10f01f7bbf6755599ca3f81ea53b8431b7dd348f5fdb2d"
+
+[[patch.unused]]
+name = "rmf_site_format"
+version = "0.0.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5833,7 +5833,3 @@ name = "zeroize"
 version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "525b4ec142c6b68a2d10f01f7bbf6755599ca3f81ea53b8431b7dd348f5fdb2d"
-
-[[patch.unused]]
-name = "rmf_site_format"
-version = "0.0.1"
--- a/rmf_site_editor/src/workspace.rs
+++ b/rmf_site_editor/src/workspace.rs
@@ -285,9 +285,9 @@ pub fn dispatch_load_workspace_events(
                     .detach();
             }
             LoadWorkspace::BlankFromDialog => {
+                let sender = load_channels.sender.clone();
                 #[cfg(not(target_arch = "wasm32"))]
                 {
-                    let sender = load_channels.sender.clone();
                     AsyncComputeTaskPool::get()
                         .spawn(async move {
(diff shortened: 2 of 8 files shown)


Cargo.lock
```diff
@@ -5833,3 +5833,7 @@ name = "zeroize"
 version = "1.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "525b4ec142c6b68a2d10f01f7bbf6755599ca3f81ea53b8431b7dd348f5fdb2d"
+
+[[patch.unused]]
+name = "rmf_site_format"
+version = "0.0.1"
```
luca-della-vedova, line 5839: I suspect this is a spurious change result of building the package through `colcon build` that can be reverted

rmf_site_editor/src/workspace.rs
```diff
@@ -282,6 +284,35 @@ pub fn dispatch_load_workspace_events(
...
+                                let file = file.path().to_path_buf();
+                                let name = file
+                                    .file_stem()
+                                    .map(|s| s.to_str().map(|s| s.to_owned()))
+                                    .flatten()
+                                    .unwrap_or_else(|| "blank".to_owned());
+                                let data = WorkspaceData::LoadSite(LoadSite::blank_L1(
+                                    name,
+                                    Some(file.clone()),
+                                ));
```
luca-della-vedova, line 303: If this is the preferred default I wonder if we should also adopt it in the function that gets called when users click `File->New` or press `Ctrl-N`, since right now the behavior would be different (so basically having the same [here](https://github.com/open-rmf/rmf_site/blob/269efabaa7def79aa3d827606e2cd21d67960dc3/rmf_site_editor/src/workspace.rs#L237-L248))

luca-della-vedova (commented):
Just a few small comments

rmf_site_editor/src/workspace.rs
mxgrey: Good point, I've standardized that now by pushing the creation of a blank site upstream: [9e66538](https://github.com/open-rmf/rmf_site/pull/220/commits/9e6653876a49fa9a780597a210b64076f2e0425a)

Cargo.lock
mxgrey: Removed: [cd61932](https://github.com/open-rmf/rmf_site/pull/220/commits/cd61932f9eb949c6c7a93b8db01bd25443afe034)

mxgrey: Oh no, I reverted this and then it automatically changed back even though I wasn't using colcon. This is going to be a pain.

luca-della-vedova (approved):

Cargo.lock
mxgrey: Fixed again: [a4c19ab](https://github.com/open-rmf/rmf_site/pull/220/commits/a4c19ab62e2416a787f2d9c3767aaa29d13ca344)

luca-della-vedova: ah! yea I usually do `git add -p .` to go through all the changes before committing, but it is an extra step

PR #221, opened 2024-06-06
Making floor grid available to all edit modes

mxgrey:
This PR adds a floor grid that can be seen in any edit mode to give users a better idea of where the floor is and what the orientation of the camera is.

The view drop-down menu provides a way to toggle the visibility of the floor grid on and off.

This will be in a draft PR until I can figure out why these changes are causing all outlining to be turned on at all times.


mxgrey:
Closing this PR and opening an issue to restart this work #304

PR #226, opened 2024-06-12
Fix assigning site IDs to new lanes in lifts

### LINKED ISSUE #224: [Bug]: Not able to add the waypoint inside the lift and door of lift

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 22.04

### Open-RMF installation type

Source build

### Other Open-RMF installation methods

_No response_

### Open-RMF version or commit hash

2.3.0

### ROS distribution

Humble

### ROS installation type

Source build

### Other ROS installation methods

_No response_

### Package or library, if applicable

_No response_

### Description of the bug

ERROR librmf_site_editor::site::save: Unable to compile site: an object has a reference to an anchor that does not exist

As this error shows when adding the lift waypoint and not able to save the site and generate the nav graph

1. Add the lift 
2. Add the waypoint inside the lift and to save the anchor point for further nav graph generation.

### Expected behavior

_No response_

### Actual behavior

_No response_

Screenshot from 2024-06-08 11-28-41

luca-della-vedova:
### Fixed bug

Fixes #224 

### Fix applied

The function to assign site IDs to new entities was not being called for anchors that were children of lifts, hence the subsequent query to save the site would not find the anchor (its ID would be missing).
The fix is to make sure we assign a site ID to anchors that are children of the lift cabin.

To test:

* Create a new site.
* Create a lift.
* Add a lane that goes into the lift.
* Save the site.

Without this PR you should get this error:

```
ERROR librmf_site_editor::site::save: Unable to compile site: an object has a reference to an anchor that does not exist
```

With this PR it should save successfully.

### DIFF
--- a/rmf_site_editor/src/site/save.rs
+++ b/rmf_site_editor/src/site/save.rs
@@ -136,6 +136,7 @@ fn assign_site_ids(world: &mut World, site: Entity) -> Result<(), SiteGeneration
             ),
         >,
         Query<(), With<DrawingMarker>>,
+        Query<&ChildCabinAnchorGroup>,
         Query<&NextSiteID>,
         Query<&SiteID>,
         Query<&Children>,
@@ -148,6 +149,7 @@ fn assign_site_ids(world: &mut World, site: Entity) -> Result<(), SiteGeneration
         lifts,
         drawing_children,
         drawings,
+        cabin_anchor_groups,
         sites,
         site_ids,
         children,
@@ -209,6 +211,17 @@ fn assign_site_ids(world: &mut World, site: Entity) -> Result<(), SiteGeneration
         }
 
         if let Ok(lift) = lifts.get(*site_child) {
+            if let Ok(anchor_group) = cabin_anchor_groups.get(*site_child) {
+                if let Ok(anchor_children) = children.get(**anchor_group) {
+                    for anchor_child in anchor_children {
+                        if let Ok(e) = level_children.get(*anchor_child) {
+                            if !site_ids.contains(e) {
+                                new_entities.push(e);
+                            }
+                        }
+                    }
+                }
+            }
             if !site_ids.contains(lift) {
                 new_entities.push(lift);
             }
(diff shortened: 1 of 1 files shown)


rmf_site_editor/src/site/save.rs
```diff
@@ -209,6 +211,17 @@ fn assign_site_ids(world: &mut World, site: Entity) -> Result<(), SiteGeneration
         }
 
         if let Ok(lift) = lifts.get(*site_child) {
+            if let Ok(anchor_group) = cabin_anchor_groups.get(*site_child) {
+                if let Ok(anchor_children) = children.get(**anchor_group) {
+                    for anchor_child in anchor_children {
+                        if let Ok(e) = level_children.get(*anchor_child) {
```
luca-della-vedova, line 217: This is not the best naming, I reused the `level_children` query that already queries for anchors, but I can create a duplicated `anchors` query if we want the cleanliness

mxgrey: `level_children` queries for a lot more things than just anchors, so I definitely think we should not be using it for this. As an example of why.. It's plausible that at some point we'd implement animated lift doors by recycling the `DoorType` component inside a child of the lift. However we don't necessarily want that child entity to have its own SiteID since it's really part of the same "site element" as the lift. If we assign SiteIDs to all these different child types without considering whether it's really an element that gets saved, then we'll be contaminating the SiteIDs.

mxgrey (changes requested):
Thanks for figuring out the problem and getting it fixed so quickly. I'm just going to tweak one detail of the implementation and then approve.

rmf_site_editor/src/site/save.rs
mxgrey: Looking at the previously existing implementation, I'm realizing that we were already assigning SiteIDs to `level_children` entities within lifts. Now I'm conflicted and wondering if there was a reason we were doing this previously, e.g. to allow walls and models to be placed inside of lifts.

mxgrey (approved):

PR #230, opened 2024-07-01
Introducing model descriptions and instances

reuben-thomas:
This PR aims to separate model descriptions from instances. When a model description is changed, for instance its mesh or kinematic properties, all of its instances should also be changed. The user should also be able to change the description of an instance.


rmf_site_editor/src/site/group.rs
```diff
@@ -34,6 +34,10 @@ impl Members {
     pub fn iter(&self) -> impl Iterator<Item = &Entity> {
         self.0.iter()
     }
+
+    pub fn len(&self) -> usize {
```
mxgrey, line 38: If you add `DerefMut` to the derives, I don't think we'll need these member functions.

xiyuoh:
Hello! This PR looks really good so far, tested out on my machine and things are working well. To help with further review, could you update the merge conflicts from the recent changes on `main`?

mxgrey:
@xiyuoh Reuben is back to school full time, so it will be on us to make any changes needed to bring this PR to completion, including fixing merge conflicts.

xiyuoh:
@mxgrey Gotcha, thanks for letting me know

Issue #231, opened 2024-07-17
Bevy 0.14 update status


luca-della-vedova:
We currently are on 0.12, 0.14 was released recently and almost all crates are ready, so we can look into it when it won't disrupt existing PRs too much.

Among other things to look forward to, there are:

* Automatic flush point insertion (for more intuitive before / after behavior)
* Fair bit of performance improvements
* Gizmos and primitive shapes (could reduce our code size a fair bit if we went for them)
* Cleanup on the asset processing side since async trait functions were stabilized

Dependency state:

* `bevy_egui`, released, `0.28`
* `bevy_mod_raycast`, released, `0.18`
* `bevy_mod_outline`, released, `0.8`
* `bevy_mod_infinite_grid`, released, `0.13`.
* `bevy_gltf_export`, git dependency, created a `bevy_0.14` branch.
* `bevy_polyline`, released, `0.10`
* `bevy_stl`, PR pending https://github.com/nilclass/bevy_stl/pull/14, will be `0.14`, worst case use fork.
* `bevy_obj`, released, `0.14`.

luca-della-vedova:
Big blocker encountered where `Commands` are no longer `Send` and `Sync` https://discord.com/channels/691052431525675048/691052431974465548/1259464518539411570.

This breaks migration for the Widget systems implementation between 0.13 and 0.14. Investigating upstream

luca-della-vedova:
PR merged and released in bevy 0.14.2, if we force that version as a minimum migration should be unblocked.

PR #232, opened 2024-07-25
Update to Bevy 0.14

luca-della-vedova:
### Implemented feature

As per description. We technically can go up to 0.14 but it is currently stuck on https://github.com/open-rmf/rmf_site/issues/231#issuecomment-2237993393, so this gets us closer (the 0.13 -> 0.14 breakage is minimal, so not a lot of duplicated work).

Will comment on the lines themselves for migrations that were not trivial


rmf_site_editor/src/interaction/camera_controls/mod.rs
```diff
@@ -257,11 +271,12 @@ impl FromWorld for CameraControls {
             .spawn(DirectionalLightBundle {
                 directional_light: DirectionalLight {
                     shadows_enabled: false,
-                    illuminance: 20000.,
+                    illuminance: 50.,
```
luca-della-vedova, line 274: This is part of a large refactoring that they did for lights.
TLDR I adopted the lights to be more physically accurate, for example the new illuminance value is corresponding to a "family living room lights", compared to the previous that was "full daylight" [ref](https://docs.rs/bevy/0.13.2/bevy/pbr/struct.DirectionalLight.html).
In full daylight, a small lightbulb would have been pretty much invisible, with this setting we can more accurately model indoor environments.

As a result I had to increase the camera exposure to make sure that the scene wouldn't result too dim. The camera exposure value itself is hand-tuned to make the appearance after the migration "close" to the one before.

rmf_site_editor/src/interaction/camera_controls/mod.rs
```diff
@@ -382,6 +407,11 @@ impl FromWorld for CameraControls {
             .push_children(&orthographic_child_cameras)
             .id();
 
+        let mut ambient_light = world.get_resource_mut::<AmbientLight>().expect(
+            "Make sure bevy's PbrPlugin is initialized before the cameras");
+
+        ambient_light.brightness = 2.0;
```
luca-della-vedova, line 414: Since we made everything a lot darker than default bevy, I also had to reduce the ambient light otherwise its effect would be overwhelming the rendering and everything would look washed out.

rmf_site_editor/Cargo.toml
```diff
@@ -55,6 +53,9 @@ nalgebra = "0.32.5"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 clap = { version = "4.0.10", features = ["color", "derive", "help", "usage", "suggestions"] }
+bevy_egui = "0.27"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 console_error_panic_hook = "0.1.7"
+# Disable the manage_clipboard feature, depends on nightly web features
+bevy_egui = { version = "0.27", features = ["open_url", "default_fonts", "render"] }
```
luca-della-vedova, line 61: Ref https://github.com/mvlabat/bevy_egui/pull/270, `manage_clipboard` only works on nightly web APIs

luca-della-vedova (commented):
It is sadly a pickle. Ran into [this issue](https://github.com/bevyengine/bevy/issues/11236) while testing. Closing the camera preview window creates a panic in `winit` that seems to have been fixed with 0.14.

However, 0.14 migration is still stuck on https://github.com/open-rmf/rmf_site/issues/231#issuecomment-2237993393.

Will let this sit for a while, if 0.14 doesn't get unstuck I would propose temporarily commenting out the `Preview` button to remove the possible panic

luca-della-vedova:
Second block on web, https://github.com/bevyengine/bevy/pull/11278 is only available in 0.14 and the 0.13 behavior regressed with the workaround not really working, will just wait for 0.14 now.

luca-della-vedova:
Note https://github.com/open-rmf/rmf_site/pull/232/commits/da1f5747b294540cfb7683fb451a25b87e42a864 should be reverted, I commented out the line for now since I couldn't find a clean migration strategy. This [PR](https://github.com/emilk/egui/pull/4763) in `egui` added a new `LayerId` parameter to the function and we might need to do a bit of refactoring to get it from the UI element.

https://github.com/open-rmf/rmf_site/pull/232/commits/48d832d875fde1d047118cf0b597bc398136bf95 is an interesting one that should be investigated and probably reverted / reworked. Without it bevy currently panics, it seems that the system ordering we were enforcing was actually wrong, we were asking for systems to run between visibility propagation and transform propagation (I'm guessing so we can set the pose before transform is propagated). However, in Bevy the transform runs _after_ visibility, so when we ask Bevy to "please run visibility, then our system set then transform", while internally it has a constraint of "run transform then visibility" it creates an unresolvable condition and panics at startup time.
This system ordering, again opposite of what we expect, is present in:

* [0.14.2](https://github.com/bevyengine/bevy/blob/5c759a1be800209f537bea31d32b8ba7e966b0c1/crates/bevy_render/src/view/visibility/mod.rs#L254-L273) (Version being migrated to in this PR)
* [0.12.1](https://github.com/bevyengine/bevy/blob/v0.12.1/crates/bevy_render/src/view/visibility/mod.rs#L247-L254) (our current version).
* [0.10](https://github.com/bevyengine/bevy/blob/v0.10.0/crates/bevy_render/src/view/visibility/mod.rs#L255-L264) (our previous version)
* [0.8](https://github.com/bevyengine/bevy/blob/0149c4145f0f398e9fba85c2584d0481a260f57c/crates/bevy_render/src/view/visibility/mod.rs#L209-L216) Our previous previous version, I'll skip for brevity but I found it all the way going back.

So my current guess is that this condition has always been undefined but it only became a panic in the latest version of Bevy, which is actually great, since we can now fix it.
Again I don't think the fix I pushed is a good one, if I understand the intention behind the system set we only care that our systems run before transform propagation (so we can make sure to set all the `Pose` components and update bevy's `Transform` components before they are propagated), but I'm not 100% sure about that.

Furthermore https://github.com/open-rmf/rmf_site/pull/232/commits/0a6ec942fe88e481ba1c53005f553cce32164e8f should be reverted (specifically, bump to bevy impulse 0.2.1 once it's re-released).
We need a feature in `bevy_impulse` that was only merged in the `0.0.2` version but this branch will be using `0.2.x`. I opened an [upstream PR](https://github.com/open-rmf/bevy_impulse/pull/46) to forward merge `0.0.x` to `0.1.x`, once it is merged we should do a `0.1.x` to `0.2.x`, trigger a `0.2.1` release on crates.io and change the dependency to `bevy_impulse = 0.2.1`. The git dependency has my branch doing this forward port.

Finally, there are a lot of warnings because of deprecated APIs / unused variables. Those are not as critical so I left them until we solve all pending items here.

luca-della-vedova:
Now that we migrated to 0.16 this is no longer relevant

PR #233, opened 2024-08-02
Move workspace loading to workflows

luca-della-vedova:
### Implemented feature

Refactor workspace loading to use `bevy_impulse` workflows.

As noted in the title, first try at using the library so happy to hear ways I can do it better / more efficiently.
Previously we created channels and used a system that spawns a detached async task that sends messages to a channel (for file dialogs) and one that continuously polls that channel to see if any event has been received.
By contrast, now this is done in a workflow where the output of the file selecting is piped into the function that will use it.

I went for an API that removes events altogether and uses workflows instead.
This has, to my understanding, a few benefits:

* We can avoid the overhead of running systems that do nothing 99.99% of the time, note how this PR removes all workspace loading systems.
* We can just use a `Res<>` instead of an `EventWriter` (which is a `ResMut` behind the scenes), potentially increasing parallelism.
* The workflows can be used in a more flexible way. For example, we can trigger a workflow from any system and it will be ran in the next command flush, rather than having them in only one point of the schedule.
* More of a future point but this architecture should be more flexible, we can reuse workflows or part of them in different apps and don't necessarily have to do a single event reading processing kind of architecture, which is a bit harder to extend.

I was on the fence on how to save workflows in the app, I went for a single resource for "all workspace loading workflows" but we can play with modularity and, for example, have one resource per workflow, use marker components, split the services, etc. but that is a whole different discussion.

If this looks good I'll tackle workspace saving next, then go for the really complex one of model handling

### DIFF
--- a/rmf_site_editor/src/workspace.rs
+++ b/rmf_site_editor/src/workspace.rs
@@ -15,7 +15,7 @@
  *
 */
 
-use bevy::{prelude::*, tasks::AsyncComputeTaskPool};
+use bevy::{ecs::system::SystemParam, prelude::*, tasks::AsyncComputeTaskPool};
 use bevy_impulse::*;
 use rfd::AsyncFileDialog;
 use std::path::PathBuf;
@@ -468,34 +468,44 @@ impl FromWorld for WorkspaceLoadingServices {
     }
 }
 
-impl WorkspaceLoadingServices {
+impl<'w, 's> WorkspaceLoadingParams<'w, 's> {
     /// Request to spawn a dialog and load a workspace
-    pub fn load_from_dialog(&self, commands: &mut Commands) {
-        commands
-            .request((), self.load_workspace_from_dialog)
+    pub fn load_from_dialog(&mut self) {
+        self.commands
+            .request((), self.workspace_loading.load_workspace_from_dialog)
             .detach();
     }
 
     /// Request to spawn a dialog to select a file and create a new site with a blank level
-    pub fn create_empty_from_dialog(&self, commands: &mut Commands) {
-        commands
-            .request((), self.create_empty_workspace_from_dialog)
+    pub fn create_empty_from_dialog(&mut self) {
+        self.commands
+            .request(
+                (),
+                self.workspace_loading.create_empty_workspace_from_dialog,
+            )
             .detach();
     }
 
     /// Request to load a workspace from a path
-    pub fn load_from_path(&self, commands: &mut Commands, path: PathBuf) {
-        commands
-            .request(path, self.load_workspace_from_path)
+    pub fn load_from_path(&mut self, path: PathBuf) {
+        self.commands
+            .request(path, self.workspace_loading.load_workspace_from_path)
             .detach();
     }
 
     /// Request to load a workspace from data
-    pub fn load_from_data(&self, commands: &mut Commands, data: WorkspaceData) {
-        commands
-            .request(data, self.load_workspace_from_data)
+    pub fn load_from_data(&mut self, data: WorkspaceData) {
+        self.commands
+            .request(data, self.workspace_loading.load_workspace_from_data)
             .detach();
     }
+}
+
+/// `SystemParam` used to request for workspace loading operations
+#[derive(SystemParam)]
+pub struct WorkspaceLoadingParams<'w, 's> {
+    workspace_loading: Res<'w, WorkspaceLoadingServices>,
+    commands: Commands<'w, 's>,
 }
 
 // TODO(luca) implement this in wasm, it's supported since rfd version 0.12, however it requires
--- a/rmf_site_editor/src/workspace.rs
+++ b/rmf_site_editor/src/workspace.rs
@@ -468,7 +468,7 @@ impl FromWorld for WorkspaceLoadingServices {
     }
 }
 
-impl<'w, 's> WorkspaceLoadingParams<'w, 's> {
+impl<'w, 's> WorkspaceLoader<'w, 's> {
     /// Request to spawn a dialog and load a workspace
     pub fn load_from_dialog(&mut self) {
         self.commands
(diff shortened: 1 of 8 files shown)


luca-della-vedova:
Web CI failure is real, it seems bevy impulse cannot compile on web because of different types uses to represent tasks in wasm (where multithreading is not implemented).

mxgrey:
I was hoping the wasm situation would have had a fix upstream by now, but I'm not surprised that it isn't working. 

One solution that comes to mind is to implement a simple single-threaded async executor that polls all the ready tasks during `flush_impulses`. We would use feature flags to choose between the async task pool vs single threaded flusher.

I think the main issue is this would make async providers unsuitable for long-running tasks that don't `await` often when they're run in web assembly, even though they'd work fine in a native build. That kind of inconsistency may cause confusion and frustration for users.

In any case I'll take a stab at this and see how it goes.

mxgrey:
I've added a [`single_threaded_async`](https://github.com/open-rmf/bevy_impulse/commit/72e76388478a5bffca92e8baf7a5ea6396c84ebb) feature for `bevy_impulse` which seems to help the wasm issue.

However there's are some new compilation errors that show up in site editor when compiling for web assembly, all related to this `map_async` call:

```
error: future cannot be sent between threads safely
   --> rmf_site_editor/src/workspace.rs:386:18
    |
386 |                 .map_async(|_| async {
    |                  ^^^^^^^^^ future created by async block is not `Send`
```

I'm pretty baffled about why that would be `Send` when not targeting wasm but not be `Send` when targeting wasm. Maybe `AsyncFileDialog` doesn't support `Send` when it targets wasm..?

Theoretically when `single_threaded_async` is active we might not need the `Send` bound on tasks since they'll always be run on one thread..? I don't know if it's realistic to relax that bound, but it might be worth trying.

mxgrey:
I've figured out a way to relax the `Send` bound when the `single_threaded_async` feature is on. Testing the web build of `rmf_site_editor` locally, it seems to be working very well. The dialog selection workflow is working exactly as expected.

The only issue I noticed is that zooming in/out seems to have some probability of making the whole screen go blank for no apparent reason. I doubt that has anything to do with this PR, in fact it could easily be a weird issue with the version of Firefox on my laptop rather than any problem is the site editor itself, but it may be something to keep an eye on.

rmf_site_editor/src/workspace.rs
```diff
@@ -264,208 +233,268 @@ pub fn dispatch_new_workspace_events(
...
+            interaction_state.set(InteractionState::Enable);
+        }
+    }
+}
+
+#[derive(Resource)]
+/// Services that deal with workspace loading
+pub struct WorkspaceLoadingServices {
+    /// Service that spawns an open file dialog and loads a site accordingly.
+    pub load_workspace_from_dialog: Service<(), (), ()>,
```
mxgrey, line 369: Very minor remark: The third generic argument is for streams which are optional (the absence of streams is represented by an empty tuple). If you just leave it out it'll [fill in the empty tuple automatically](https://github.com/open-rmf/bevy_impulse/blob/3e681a91033427c5ae24995f4e53d19f656f278f/src/service.rs#L68).

So for example this could be `Service<(), ()>`. It's a very minor thing, but I think it's nicer to just see the input and output in the Service signature when there are no streams.

rmf_site_editor/src/workspace.rs
```diff
@@ -264,208 +233,268 @@ pub fn dispatch_new_workspace_events(
...
+            create_empty_workspace_from_dialog,
+            load_workspace_from_path,
+            load_workspace_from_data,
+        }
+    }
+}
+
+impl WorkspaceLoadingServices {
+    /// Request to spawn a dialog and load a workspace
+    pub fn load_from_dialog(&self, commands: &mut Commands) {
```
mxgrey, line 473: I don't have any objection to doing it this way, but I can suggest two alternatives that might be more ergonomic or idiomatic:

## 1. SystemParam

You could make a custom SystemParam struct that queries for both this resource and for `Commands`, then offers these `load_from_dialog`, etc methods and takes care of using Commands and the resource.

## 2. Extension trait for Commands

You could define a trait extension for Commands that implements these methods directly on Commands. However you'd also have to define one or more custom structs that implement the `Command` trait to get access to the resource from the `World` during the command flush, so this would be a bit more troublesome.

luca-della-vedova: +10 for the `SystemParam` approach, it's definitely a net gain so we can avoid this awkward `&mut commands` API, I forgot that it existed.
Added in https://github.com/open-rmf/rmf_site/pull/233/commits/5d08fbfb02681d3773ccd891b0667ac2146557a2

I do agree that the extension trait approach tends to be a bit boilerplate-y and harder to trace, but in any case we can mix and match and / or revisit when it's time to split the workcell editor mode

rmf_site_editor/src/workspace.rs
luca-della-vedova: Didn't know that! Done https://github.com/open-rmf/rmf_site/pull/233/commits/659da701cabb5f2f56ba45473bf7ec028866d0b1

rmf_site_editor/src/workspace.rs
```diff
@@ -264,211 +233,281 @@ pub fn dispatch_new_workspace_events(
...
+    pub fn load_from_data(&mut self, data: WorkspaceData) {
+        self.commands
+            .request(data, self.workspace_loading.load_workspace_from_data)
+            .detach();
+    }
+}
+
+/// `SystemParam` used to request for workspace loading operations
+#[derive(SystemParam)]
+pub struct WorkspaceLoadingParams<'w, 's> {
```
mxgrey, line 506: Super super minor nitpick but what if we call this `WorkspaceLoader` instead of `WorkspaceLoadingParams`? I think `WorkspaceLoader` would make its purpose more clear. Users shouldn't really have to care that it's a struct of system params.

luca-della-vedova: https://github.com/open-rmf/rmf_site/pull/233/commits/41f6233fad7df6db384880798c8b518a3d0d9577

mxgrey (approved):
This is an awesome introduction of `bevy_impulse` into actual use :+1:

PR #235, opened 2024-08-16
Automatic adjustment of log display range

xiyuoh:
This PR addresses the TODO to allow the slider range for displaying number of logs to be adjusted automatically. Currently we have a fixed range of [10, 1000]. With this PR, the range will start with [10, 100] at launch. As more logs are being printed, the upper limit will increase accordingly to the nearest multiple of hundred that is greater or equal to the current log history size.


luca-della-vedova (approved):

luca-della-vedova:
I can predict before the CI runs that this will fail web CI because of the usual `wasm-bindgen` bump

PR #236, opened 2024-08-21
Use workflows for mouse interactions

mxgrey:
This PR removes the `InteractionMode` state machine and replaces it with a high-level workflow for switching between mouse interaction services. This high-level workflow uses service injection to allow custom mouse interaction services to be plugged in by downstream users of the site editor.

We introduce several different mouse interaction services that can activated by the high-level workflow. Each of these mouse interaction services is implemented as its own workflow:
* Create edge
* Replace a side of an edge
* Create point
* Replace anchor for a point
* Create path
* Place an object in 2D
* Place an object in 3D

Some other mouse interaction workflows that we should consider adding in the future (left out of this PR since they were not already available or not well supported):
* Add a new anchor into an existing path
* Swap an anchor that a path is using with another anchor
* Re-parent a 3D object to another reference frame

This PR introduces some major changes to 3D object placement for the workcell editor. Placing a 3D object now involves an optional step of selecting a parent reference frame for the object. The workflow goes like this:
1. If the user had something selected when they clicked on the spawn model button, that something will automatically be used as the parent.
2. If nothing was selected when the spawn model button was clicked, we will not start out with any parent.
3. Pressing Esc while a parent is selected will clear out the parent, leaving the workflow state similar to (2).
4. Clicking on an object in the scene while no parent is selected will select that object as the parent.
5. Clicking on a spot on the selected parent will place the object at that location on the parent, with the z-axis aligned with the normal vector of the chosen spot.
6. Clicking somewhere that's not on the parent object will place the object in the x-y plane of the parent's frame.
7. Clicking somewhere that's not on any object will place the object there, whether or not a parent is selected.
8. Holding shift and clicking will ignore any hovering that is happening, placing the object in the x-y plane of the parent frame, or in the x-y plane of the world if no parent is selected.

All of this is communicate to the user through a combination of cursor-attached previewing and a new cursor-attached tooltip feature.

This PR also introduces some crucial fixes to model loading which had some reliability issues around potentially random system ordering and incomplete property propagation.


luca-della-vedova:
> Some other mouse interaction workflows that we should consider adding in the future (left out of this PR since they were not already available or not well supported):
> * Re-parent a 3D object to another reference frame

This feature is necessary for workcell calibration, the sequence is usually the following:

* Spawn a mesh in the desired position, its parent can be anything.
* Spawn a calibration frame for the mesh with the intended parent of the mesh. Snap it to a corner of the mesh where a calibration datum is placed.
* Reparent the mesh to be a child of the calibration frame. Note that reparenting calculates the `Transform` such that the `GlobalTransform` would be unchanged, so the position of the mesh will be the same.
* Run the calibration routine which readjusts the calibration frames based on external input.
* The mesh position will be automatically adjusted as well since it is a child of the calibration frame.

If we cannot reparent a mesh to a different anchor we can't do this calibration workflow, so the functionality is needed.

The most important point however, is in the redesign of the parenting UX.

It is a deliberate design choice in the current workcell editor that the only entity that can be a parent of another entity is a `Frame`. This is necessary for a few reasons:

* Allowing arbitrary hierarchies becomes messy very easily, for example a model that is a child of another model, then has a frame as a child, how do we translate this hierarchy to a robot with links and joints? Only having frames as parents makes this easier.
* URDF interoperability, one of the most important features of the workcell editor is to export it as a URDF, as well as being able to have a "tf tree" that is compatible with ROS 2. This works if we keep the concept of frames only being children of other frames. We simply translate `Frame` to TF frames and URDF links, meshes as `visual` and `collision` elements of links, `Inertia` children of `Frame`s as their `inertial` components. However once we let frames be children of meshes or meshes be children of other meshes this conversion becomes a lot harder (if even possible?).

I'm attaching a videos of a few of the rough edges that came up while allowing models to be parents of other entities:
[Screencast from 2024-08-22 12-06-54.webm](https://github.com/user-attachments/assets/b4ca105d-2674-4c6a-b3f5-eddf59ed443e)

For example when spawning a model as a child of a model the model disappears, frames that are children of models seem to be embedded in the model when saving and reloading. TBH I would just only allow frames as parents to avoid all of these and make the URDF interoperability easier.

Additionally, there was a check in place to make sure we only do the "mesh snapping" when hovering over [models](https://github.com/open-rmf/rmf_site/blob/47552eab5f7de5f056b9c76c11989af1a4b4ba53/rmf_site_editor/src/interaction/cursor.rs#L377-L380). This is necessary to avoid snapping to the anchor meshes as the video below (which is probably not something that a normal user would want):

[Screencast from 2024-08-22 12-19-04.webm](https://github.com/user-attachments/assets/6444fde3-19c3-40bd-9bbe-14c60fee3ff4)

I need to setup the rest of the infrastructure to test the rest of the workcell editor pipeline, but these are the points I could find immediately

mxgrey:
> If we cannot reparent a mesh to a different anchor we can't do this calibration workflow, so the functionality is needed.

I see, the only reason I didn't include it in this PR is because I wasn't sure how to preview it to the user. I thought we'd want to snap the model back to the cursor while selecting the new parent, and that would've been very difficult to propagate properties correctly.

What you've described, where the object remains at the same global transform as we reparent, will be very to implement, so I'll go ahead and add that in now.

> It is a deliberate design choice in the current workcell editor that the only entity that can be a parent of another entity is a Frame.

This should be very easy to fix. I just need to drill up to the first ancestor frame of the selected parent before inserting the object into the world. I think down the road we should consider whether it's possible to automatically consider every tangible object in the scene to be a frame, but that's obviously outside the scope of this PR.

mxgrey:
* [e9a36cf](https://github.com/open-rmf/rmf_site/pull/236/commits/e9a36cfe382d281b02b95605a01ab6d89a7211e4) makes sure that only frames can be parents to objects (anchors, models, visuals, and collisions)
* [fdcd8c0](https://github.com/open-rmf/rmf_site/pull/236/commits/fdcd8c0d612b688c23b07dc7938e4df4081e7443) prevents the intersection ray from hitting the round surface of anchors
* [49ab9b9](https://github.com/open-rmf/rmf_site/pull/236/commits/49ab9b92ad8135046b936216c5eab9a6c570300c) brings back the reparenting selection mode. Since there's no preview while using this mode, I found it difficult as a user to tell whether the app is in "select parent" mode versus being in regular inspector mode. To mitigate that problem, I added a tooltip that's always on during this mode to indicate that you're in parent selection mode.

The vanishing model issue was fixed with [e9a36cf](https://github.com/open-rmf/rmf_site/pull/236/commits/e9a36cfe382d281b02b95605a01ab6d89a7211e4). I also tested a round-trip of saving and reloading, and everything appeared to be working: everything loaded back up, and all the parent-child relationships appeared to be in tact.

I believe this addresses all the current feedback, but please let me know if I missed anything.

rmf_site_editor/src/interaction/select/select_anchor.rs
```diff
@@ -0,0 +1,650 @@
...
+    state_setup: Service<BufferKey<State>, SelectionNodeResult>,
+    update_preview: Service<(Hover, BufferKey<State>), SelectionNodeResult>,
+    update_current: Service<(SelectionCandidate, BufferKey<State>), SelectionNodeResult>,
+    handle_key_code: Service<(KeyCode, BufferKey<State>), SelectionNodeResult>,
+    cleanup_state: Service<BufferKey<State>, SelectionNodeResult>,
+    anchor_cursor_transform: Service<(), ()>,
+    anchor_select_stream: Service<(), (), (Hover, Select)>,
+    keyboard_just_pressed: Service<(), (), StreamOf<KeyCode>>,
+    cleanup_anchor_selection: Service<(), ()>,
+) -> impl FnOnce(Scope<Option<Entity>, ()>, &mut Builder) {
```
luca-della-vedova, line 306: This is also a fairly complex function, maybe a breakdown of what each service that users are expected to set is supposed to do would help?

rmf_site_editor/src/widgets/canvas_tooltips.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Push a `Cow::Borrowed("Your string literal here")` to push a `&'static str`,
+    /// which is the most efficient way to produce tooltip text.
+    ///
+    /// If your text is not a string literal then push `Cow::Owned(your_string)`.
+    tips: Vec<Cow<'static, str>>,
+    /// We keep track of previous tooltips to keep order as consistent as possible
+    /// to prevent flickering as it's possible for different systems to add in
+    /// tooltips at different times.
+    previous: Vec<Cow<'static, str>>,
+}
```
luca-della-vedova, line 38: A bit nitpicky I guess but what if the tooltips were just in an automatically ordered structure (i.e. `BTreeSet`)? It sounds like ther order of pushing is a bit erratic anyway so maybe a deterministic alphabetical order would just get rid of the complexity of keeping track of the old values and doing searches every iteration

luca-della-vedova (commented):
It's a lot to go through, I left a few minor feedback around and rough edges during my testing.

It's probably quite clear given that the only real code feedback I gave is on a small isolated module that this change is very complex to understand and I admittedly don't understand most of it. I think what is most valuable is trying to document as much as possible the complex workflows that are introduced, what their steps do, how users are supposed to configure them and a bit more explanation on the actual complex workflow APIs used.
Imo this is important especially since `bevy_impulse` is in very early stages with very little applications. This is pretty much the first complex application and will probably be referenced a lot by people when trying to design their own workflows.

I'll also have to understand this in quite a lot of detail when trying to work on the workcell editor mode but I suspect that will take some weeks and it's probably not worth it to keep this PR in limbo until then.

rmf_site_editor/src/widgets/canvas_tooltips.rs
mxgrey: The rationale for the current approach is that the oldest items in the tooltip will remain at the top, which I think gives better stability for a human who may be trying to read it. If we use a collection with built-in ordering like `BTreeSet` then new tips can, at any time, shift down a message that a user is paying attention to. Messages that are more fickle will remain towards the bottom, which I think is also helpful for the user to be able to focus on the bottom if they want to watch what tooltips are coming and going.

Personally I think this (admittedly subtle) quality of life improvement is worth the added complexity, but what do you think now that I've explained the rationale?

rmf_site_editor/src/widgets/inspector/inspect_fiducial.rs
```diff
@@ -92,6 +95,11 @@ impl<'w, 's> WidgetSystem<Inspect> for InspectFiducial<'w, 's> {
             return;
         };
         let Ok(tracker) = params.usage.get(parent.get()) else {
+            error!(
+                "Unable to find usage tracker for parent {:?} of fiducial {:?}",
+                parent.get(),
+                selection,
+            );
```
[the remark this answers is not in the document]
mxgrey: Great catch, it turns out the anchor selection filter wasn't properly filtering out existing anchors that belonged to the wrong type of scope. Cartesian (non-drawing) fiducials are only supposed to be attached to site anchors, so some of the logic breaks when they get attached to level anchors.

Fixed by this commit: [205d3ec](https://github.com/open-rmf/rmf_site/pull/236/commits/205d3ec38fbc22a2b3d002ba539846f1c3446ad3)

rmf_site_editor/src/widgets/canvas_tooltips.rs
luca-della-vedova: I see, I thought the main problem to be solved was that the tooltips were usually the same but system ordering might cause them to be pushed in different order, but if there is frequent adding / removing this sounds good

luca-della-vedova:
Last nit is that this PR introduces a bunch of extra warnings, like unused variables and imports that increase noise and make it more likely that we will miss important ones in the future. Should make sure to clean them up before merging

mxgrey:
Ah yeah actually quite a few warnings that have been lingering were kept on purpose as reminders for issues that need to be addressed in the old mouse selection state machine, so this is a good opportunity to clean up any remaining warnings. 

But I don't look forward to the day we try to apply clippy to this project...

rmf_site_editor/src/interaction/select.rs
```diff
@@ -15,10 +15,190 @@
...
+        app.world.command(|commands| {
+            commands.request((), select_workflow).detach();
+        });
+    }
+}
+
+pub fn build_select_workflow(
+    inspector_service: Service<(), ()>,
+    new_selector_service: Service<(), (), StreamOf<RunSelector>>,
+) -> impl FnOnce(Scope<(), ()>, &mut Builder) -> DeliverySettings {
```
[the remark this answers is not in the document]
mxgrey: Explanation of the workflow as well as a breakdown of the operations added [here](https://github.com/open-rmf/rmf_site/pull/236/commits/423239d9b7e7f42a2288d20d2949d723489b9714#diff-0aea2494ad26d1476e6ce7b9eaaa32959fa42b072fdab24ecde0816a2dbe5885R129). This one would probably be much easier to understand with a graphical representation. I might try to poke around for a suitable tool for creating and rendering [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) files that we might link to in the docs.

mxgrey:
I think all the latest feedback is addressed.

I also realized that I accidentally got rid of the ability to deselect the current inspected item by pressing the `Esc` key so I added that back in with [ca82bee](https://github.com/open-rmf/rmf_site/pull/236/commits/ca82bee9a4326a8d8bf1c194db4e7f479dfc14e5).

luca-della-vedova (approved):
Thanks for iterating!

PR #239, opened 2024-09-12
Split out workcell editor

luca-della-vedova:
### Implemented feature

Split workcell editor out into separate app, new app in https://github.com/luca-della-vedova/rmf_workcell/. Based on #238 so that should be merged first.


rmf_site_editor/src/interaction/select/place_object.rs
```diff
@@ -31,38 +34,28 @@ impl Plugin for ObjectPlacementPlugin {
...
+    current_level: Res<'w, CurrentLevel>,
 }
 
 impl<'w, 's> ObjectPlacement<'w, 's> {
-    pub fn place_object_2d(&mut self, object: Model, level: Entity) {
+    pub fn place_object_2d(&mut self, object: Model) {
+        let Some(level) = self.current_level.0 else {
+            warn!("Unble to create [object:?] outside a level");
+            return;
+        };
```
luca-della-vedova, line 58: This is part of my effort to reduce code duplication. I refactored both 2d and 3d object placement to take the same input (just a `Model`).
Specifically for the fuel widget, it was really the same between this app and the workcell editor, with the only different that the spawning function was different. I introduced a field in the plugin that allows users to pass a spawning function that takes as an input `Commands` and a `Model`, and will be called when the `Spawn model` button is clicked. This allowed reusing the code between the two applications but comes with the drawback that this minor API change is necessary.

https://github.com/open-rmf/rmf_site/blob/3496425691f9c43b46b6acb59bd257961e80a6f9/rmf_site_editor/src/widgets/fuel_asset_browser.rs#L28-L40

rmf_site_editor/src/lib.rs
```diff
@@ -15,6 +14,10 @@ pub use autoload::*;
 pub mod asset_loaders;
 use asset_loaders::*;
 
+// Bevy plugins that are public dependencies, mixing versions won't work for downstream users
+pub use bevy_egui;
+pub use bevy_mod_raycast;
```
luca-della-vedova, line 19: This is another interesting one that I'm considering pushing further.
If users need to use any of these plugins in their downstream applications they need to also include them in the exact same version in their `Cargo.toml` or might end up with very cryptic errors if two different versions are included in their application vs the site editor (since the two different versions will compile to types with the same name that are actually different and not interchangeable).

So my approach was to just make these dependencies public so users will just import them through the site editor. These two are the bare minimum needed for the workcell editor and probably most applications (`bevy_egui` is needed to create custom widgets, `bevy_mod_raycast` for raycasting for intersections with meshes / ground planes) but I'm happy to make more dependencies public if this approach sounds good.

rmf_site_editor/src/site/mod.rs
```diff
@@ -321,24 +304,21 @@ impl Plugin for SitePlugin {
...
+                .run_if(AppState::in_displaying_mode())
                 .in_set(SiteUpdateSet::AssignOrphans),
         )
         .add_systems(
             PostUpdate,
             (
                 add_wall_visual,
-                handle_update_fuel_cache_requests,
-                read_update_fuel_cache_results,
-                reload_failed_models_with_new_api_key,
```
luca-della-vedova, line 333: I just refactored these and the matching resources into a `FuelPlugin` to make it easier to add it to apps.

rmf_site_editor/src/widgets/inspector/mod.rs
```diff
@@ -294,10 +282,13 @@ impl<'w, 's> WidgetSystem<Tile> for Inspector<'w, 's> {
         state: &mut SystemState<Self>,
         world: &mut World,
     ) {
+        // TODO(luca) make sure this doesn't show in building preview mode
+        /*
         match world.resource::<State<AppState>>().get() {
             AppState::SiteEditor | AppState::SiteDrawingEditor | AppState::WorkcellEditor => {}
             _ => return,
         }
+        */
```
luca-della-vedova, line 290: This is the main point that I'm still trying to figure out.
The moment we access `AppState` from any of the widgets, that widget will be necessarily need the site editor and its `AppState`, making it extremely limiting.
If this block was present a downstream app that wanted to use this widget would need to either:
* Add `librmf_site_editor::AppState` to their app and make sure it is in the correct state, regardless of whether they need it or not or
* Copy the whole plugin just to remove this snippet (or customize it for their use case).

The drawback of this specific comment (and the reason it's still marked as a TODO) is that specifically for the site editor this will result in the inspection plugin showing up during the building preview mode, which might not be desirable.
I'm not sure if there is a better solution that doesn't involve a larger refactor.

rmf_site_format/src/misc.rs
```diff
@@ -353,6 +353,33 @@ impl Default for Pose {
...
+#[cfg(feature = "urdf")]
+impl From<&urdf_rs::Pose> for Pose {
+    fn from(pose: &urdf_rs::Pose) -> Self {
+        Pose {
+            trans: pose.xyz.map(|t| t as f32),
+            rot: Rotation::EulerExtrinsicXYZ(pose.rpy.map(|t| Angle::Rad(t as f32))),
+        }
+    }
+}
+
```
luca-della-vedova, line 382: I believe these have to be implemented here, I added a default off `urdf` feature to remove the dependency unless needed.

rmf_site_editor/src/site/change_plugin.rs
```diff
@@ -63,9 +62,7 @@ impl<T: Component + Clone + Debug> Plugin for ChangePlugin<T> {
     fn build(&self, app: &mut App) {
         app.add_event::<Change<T>>().add_systems(
             PreUpdate,
-            update_changed_values::<T>
-                .run_if(AppState::in_displaying_mode())
-                .in_set(SiteUpdateSet::ProcessChanges),
+            update_changed_values::<T>.in_set(SiteUpdateSet::ProcessChanges),
```
luca-della-vedova, line 65: This is a tricky one, similar change is in the [recall plugin](https://github.com/open-rmf/rmf_site/blob/3496425691f9c43b46b6acb59bd257961e80a6f9/rmf_site_editor/src/site/recall_plugin.rs#L35-L41).
The issue we are trying to tackle is that if users wanted to use this plugin in a third app they would need to include the rmf_site_editor `AppState` or the `run_if` function would panic.
The main drawback here is that I believe this changes the system to _always_ run. Now it could be argued that the net difference in this case (and others through this PR) is that they will additionally run in `AppState::MainMenu`, where there are very few entities, pretty much no work to be done and very little CPU usage, so I'd argue it's not a big issue, but still worth considering.

rmf_site_editor/src/widgets/workspace.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct WorkspaceMenuPlugin {}
+
+impl Plugin for WorkspaceMenuPlugin {
+    fn build(&self, app: &mut App) {
+        app.init_resource::<WorkspaceMenu>().add_systems(
+            Update,
+            handle_workspace_menu_events.run_if(AppState::in_displaying_mode()),
+        );
+    }
+}
```
luca-della-vedova, line 32: This is a new widget for workspace creation, now the top menu bar is fully modular and doesn't have any "prespawned" buttons.
The reason this is needed is again modularity and code duplication. The top menu bar buttons forced the widget to rely on internal workspace creation / file loading / file saving events, this meant that users that wanted custom workflows for these functions would need to duplicate the whole top menu bar.
Moving them to a classic child system means that users just need to add their children to the file menu instead.

This is necessary because these kind of file operations will most likely have to be customized in different apps. For example a file loading might need custom extension filters, a different loaded file handler etc.

rmf_site_editor/src/workspace.rs
```diff
@@ -184,31 +101,22 @@ impl Plugin for WorkspacePlugin {
...
                 Update,
-                (
-                    dispatch_new_workspace_events,
-                    sync_workspace_visibility,
-                    workspace_file_save_complete,
-                ),
+                (dispatch_new_workspace_events, sync_workspace_visibility),
             );
-        #[cfg(not(target_arch = "wasm32"))]
-        app.add_systems(Update, dispatch_save_workspace_events);
```
luca-della-vedova, line 203: There are several changes to this file, I'll just highlight them in this snippet that offers an overview.

Model saving now uses workflows, furthermore I separated the `rfd` file dialog spawning into a new `FileDialogServices` that provides workflows that allow users to pick files for loading / saving (optionally with filters for extensions), as well as picking folders for exporting (useful for SDF / URDF exporting).

Apps that need to handle workspaces will most likely need to create their own loading and saving services, however now they can reuse the `FileDialogServices`, as well as avoid having to worry about `rfd` so we can keep it as a private dependency.
Note that there are other parts in the codebase where we could use the new service rather than `rfd` (specifically three instances, saving lights, saving navgraphs and browsing local asset sources) however I left them out for now to try to limit the scope of this PR, I added TODOs in their place.

rmf_site_format/src/lib.rs
```diff
@@ -98,6 +95,9 @@ pub use physical_camera::*;
 pub mod point;
 pub use point::*;
 
+pub mod primitive_shape;
+pub use primitive_shape::*;
+
```
luca-della-vedova, line 100: I moved primitive shapes in the root of `rmf_site_format` from the workcell module. This is necessary since SDFs might contain primitive shapes (usually for collision meshes)

rmf_site_format/Cargo.toml
```diff
@@ -16,14 +16,17 @@ thiserror = "*"
...
+urdf-rs = { version = "0.7.3", optional = true }
 # Used for lazy initialization of static variable when they are non const
 once_cell = "1"
 pathdiff = "*"
 
 [dev-dependencies]
 float_eq = "1.0"
+
+[features]
+urdf = ["dep:urdf-rs"]
```
mxgrey, line 32: I think I would recommend making this a default feature. We just need to add
```
default = ["urdf"]
```

Considering the community that we're targeting, I think 
1. It's far more likely that the user will be using URDFs at least somewhat than not using URDFs at all
2. The small proportion of users that will absolutely not use URDF would probably not even notice the dependency
3. Since it's still an optional dependency, they can always disable it if they really don't want it

rmf_site_editor/src/sdf_loader.rs
```diff
@@ -35,7 +35,19 @@ pub struct SdfPlugin;
 
 impl Plugin for SdfPlugin {
     fn build(&self, app: &mut App) {
-        app.init_asset_loader::<SdfLoader>();
+        // Type registration is necessary to allow serializing the Scene that is loaded by this
+        // plugin
+        app.init_asset_loader::<SdfLoader>()
+            .register_type::<NameInSite>()
```
mxgrey, line 41: Some of these components are used pretty broadly outside of the SdfPlugin, so it's not obvious to me that they should be grouped here exclusively.

What would you think of having a `RegisterCommonTypesPlugin` that has all of this in it, and which gets loaded by `SdfPlugin`?

rmf_site_editor/src/site/change_plugin.rs
mxgrey: I certainly don't mind migrating away from a dependency on the `AppState`.

If anything maybe we should give this its own state tracker to explicitly toggle `ProcessChanges` on/off, but we can save that for later if we benchmark and find that it's eating up more resources than it should.

rmf_site_editor/src/widgets/view_nav_graphs.rs
```diff
@@ -215,6 +215,7 @@ impl<'w, 's> ViewNavGraphs<'w, 's> {
                         }
                         None => {
                             let future = AsyncComputeTaskPool::get().spawn(async move {
+                                // TODO(luca) change this to use FileDialogServices
```
mxgrey, line 218: Is this something that needs to be done before we merge, or is it a note for a future improvement?

rmf_site_editor/src/widgets/fuel_asset_browser.rs
```diff
@@ -16,24 +16,37 @@
...
+            model_spawner: |commands: &mut Commands, model: Model| {
+                commands.place_object_2d(model);
+            },
+        }
+    }
+}
+
+/// Stores the function that uses commands to place models for this application
+#[derive(Resource, Deref, DerefMut)]
+struct ModelPlacer(fn(&mut Commands, Model));
```
mxgrey, line 44: I have no problem using a function pointer for this, but in the future we could consider making this a `BoxedSystem` for added extensibility. The downside would be a lot of extra management of borrows, so we can definitely punt on that until we find a situation where it matters.

rmf_site_editor/src/widgets/inspector/mod.rs
mxgrey: One simple-ish option would be to have an AppState specifically for inspection that we use to toggle the inspector widget on/off explicitly.

I have a feeling that to get a satisfying level of granularity, we'll end up having dedicated state types for each different plugin. That wouldn't be a bad thing if we can additionally still have a high-level `AppState` that automatically manages all those different granular states types.

I don't mind punting that to a follow-up PR so we're not blocking this one in the meantime.

rmf_site_editor/src/lib.rs
mxgrey: This is a good idea for anything that we're exposing in the public API of the site editor :+1: 

I would just avoid re-exporting any dependencies that aren't explicitly being used in our public API, but I'm sure you know that already.

rmf_site_editor/src/interaction/mod.rs
```diff
@@ -165,14 +168,16 @@ impl Plugin for InteractionPlugin {
...
                 CategoryVisibilityPlugin::<CollisionMeshMarker>::visible(false),
                 CategoryVisibilityPlugin::<MeasurementMarker>::visible(true),
                 CategoryVisibilityPlugin::<WallMarker>::visible(true),
-                CategoryVisibilityPlugin::<WorkcellVisualizationMarker>::visible(true),
             ))
-            .add_plugins((CameraControlsPlugin, ModelPreviewPlugin));
+            .add_plugins((
+                CameraControlsPlugin,
+                ModelPreviewPlugin,
+                UserCameraDisplayPlugin::default(),
```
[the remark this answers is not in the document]
mxgrey: I'm concerned that this would cause problems if someone wants to use the interaction plugin but not the egui widgets.

I've opened a PR that allows this to be decoupled: https://github.com/open-rmf/rmf_site/pull/247

mxgrey (commented):
This is looking really good, I just have a few inline suggestions.

Also I noticed that there are a few comments of
```
// TODO(luca) change this to use FileDialogServices and be async
```
Are those meant to be addressed before merging this PR or some time afterwards?

rmf_site_editor/src/sdf_loader.rs
luca-della-vedova: The main reason is that component registration is only needed when we are adding a component to a `Scene` and we want to serialize / deserialize such `Scene`, which is currently only done in the `SdfLoader` https://github.com/open-rmf/rmf_site/blob/983cd3c943e47924c5cf436e30becb3d4cb357c3/rmf_site_editor/src/sdf_loader.rs#L308
Trying to load an sdf with the `SdfLoader` _without_ registering all the types that it will add to the scene will result in a panic. If in the future we added a new component to the loader (i.e. parsing a new SDF property) without registering that will trigger a panic at the bevy level.

```
thread 'main' panicked at [...]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bevy_scene-0.12.1/src/scene_spawner.rs:430:35:
scene contains the unregistered type `rmf_site_format::misc::NameInSite`. consider reflecting it with `#[derive(Reflect)]` and registering the type using `app.register_type::<T>()`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in exclusive system `bevy_scene::scene_spawner::scene_spawner_system`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

To me the strong coupling of registering the components in this plugin sends a message that these components _must_ be registered for this plugin to work, and if a new one was to be added in the future it also _must_ be registered. The moment the registration is done in another plugin somewhere else we might forget this and trigger panics when adding new components to the `SdfLoader`.

I think my recommendation here would be to keep it as is but refactor it into a `RegisterCommonTypesPlugin` as soon as we have another place where we serialize / deserialize scenes, at which point we might change the definition from "register everything that is needed by `SdfLoader`" to "register everything that the site editor needs" and make it a dependency of the `SitePlugin` instead.
That being said I noticed that all this coupling and potential panic info was all in my head and not in the code so I added a clearer note https://github.com/open-rmf/rmf_site/pull/239/commits/5275a35edce7c3351306e7550af6e144d4b1de09

rmf_site_format/Cargo.toml
luca-della-vedova: The "idea" behind it was that the site editor itself doesn't support anything `urdf` related at all so I could cut the dependency (in this case with your change I would have probably added a `default_features = off` to its dependency of `rmf_site_format`). But I noticed that we use one [tiny function](https://github.com/open-rmf/rmf_site/blob/983cd3c943e47924c5cf436e30becb3d4cb357c3/rmf_site_editor/src/site_asset_io.rs#L17) to parse `package://` paths and copying the code in doesn't bring any significant gain in dependency count.
TLDR added your suggestion in https://github.com/open-rmf/rmf_site/pull/239/commits/4637e1b6967d6c4d832a8daf6c4d561c9d3ddcdc

rmf_site_editor/src/widgets/view_nav_graphs.rs
luca-della-vedova: I would suggest leaving it for a future improvement, there is no regression and everything still works it's just that this existing code can probably be cleaned up.
I tried to keep the PR reasonably well scoped (I actually regret bundling in the `FileDialogServices` and workspace saving changes, it would have been a lot slimmer without). It was super low in the priority list because most of these workflows are fairly niche. There is a total of three cases where I added a TODO, in increasing level of "obscureness of the usecase":

* Browsing asset sources for local assets, local assets are not used too much and most of the assets we do are based on Fuel.
* Exporting nav graphs: Can be useful but from what I understand next gen would use directly `.site` files, current gen would probably generate navgraphs through the `Export SDF` feature at build time.
* Exporting lights: This is so niche that I almost wonder if we should get the button out of the UI altogether to declutter it. It exports all the light information into a text file and if I remember correctly was only used in a demo a long time ago.

rmf_site_editor/src/lib.rs
luca-della-vedova: I tried to push this a bit further, specifically I wanted to re-export `bevy` but I'm a bit worried it might not be a good idea.
The main rationale behind it is that `rmf_site_format` depends on `bevy`, if anyone was to write an application that depends on `rmf_site_format` but doesn't use the same version of `bevy` they will have a lot of issues (it actually happened to me in the past when updating to a newer version!).

The approach would have been something like adding in rmf_site_format's [lib.rs](https://github.com/open-rmf/rmf_site/blob/main/rmf_site_format/src/lib.rs) the following:

```
#[cfg(feature = "bevy")]
pub use bevy;
```

Then changing the site editor to do a:

```
// lib.rs
pub use rmf_site_format::bevy;

// Whenever bevy is used
use crate::bevy::prelude::*;
```

That would ensure that both the site editor and all downstream applications use the same version of bevy that `rmf_site_format` was compiled with. However I do not know how this would work with custom features. For example, `rmf_site_format` only depends on bare bevy: https://github.com/open-rmf/rmf_site/blob/983cd3c943e47924c5cf436e30becb3d4cb357c3/rmf_site_format/Cargo.toml#L20

While the site editor needs additional features: https://github.com/open-rmf/rmf_site/blob/983cd3c943e47924c5cf436e30becb3d4cb357c3/rmf_site_editor/Cargo.toml#L34

It is not clear to me if encouraging a pattern of using another crate's public dependency while downstream applications might need to redefine the dependency to add optional features introduces a lot of value, so I'll currently leave as this proposal where the public dependencies are just simple plugins that don't have any optional feature and can just reduce churn when migrating between versions of Bevy / creating a new application that depends on the site editor.

rmf_site_editor/src/widgets/view_nav_graphs.rs
mxgrey: This might be a good thing to open an issue ticket for, maybe with a "good first issue" tag, although that may be subjective since the contributor may have to have some understanding of async.

mxgrey (approved):

rmf_site_editor/src/widgets/view_nav_graphs.rs
mxgrey: Done with https://github.com/open-rmf/rmf_site/issues/248

PR #242, opened 2024-11-11
Introducing model descriptions and instances

xiyuoh:
This is a continuation of Reuben's PR https://github.com/open-rmf/rmf_site/pull/230, opening a new one to accommodate push restrictions.

This PR:
- Introduces the concept of model descriptions and instances, where a model description provides the information about a model (asset source, scale, etc.), and a model being spawned on site is an instance of a particular model description. This allows users to make changes to a group of model instances with the same model description at one go. This is helpful for sites with multiple identical models.
- Allows users to create different scenarios with a scenario tree, where they can edit and switch between variants of the same environment with the ability to undo or redo changes without affecting the root scenario.
- Provides a configurable property plugin that enables users to add new properties and configuration widgets, such as mobile robot tasks for instances and differential drive properties for descriptions.


xiyuoh:
Changes from Reuben's PR:

Major changes
- Model instance deletion: The logic for permanently deleting/temporarily removing a Model Instance in the original PR goes through the `Delete` Event twice - once to check whether the Model Instance should be temporarily removed only or permanently deleted (depends on the current scenario where this Model Instance is displayed), then a second time for actual deletion. The effort to streamline this process is in https://github.com/open-rmf/rmf_site/pull/243, and is implemented here with `filter_instance_deletion` ([057e186](https://github.com/open-rmf/rmf_site/pull/242/commits/057e1866a7f5db1d0695655dc4088917556622b4)).
- Refactor `Tasks` implementation - it started by wanting to implement the Display trait for Tasks so that we don't have to rely on a vector of Strings, but it led to quite a huge change in the way the Tasks enum is being written ([acb39b9](https://github.com/open-rmf/rmf_site/pull/242/commits/acb39b9a6f628f19b946780b410533d46b9cd9de), [94b54a9](https://github.com/open-rmf/rmf_site/pull/242/commits/94b54a94c021e1baee2d3a9913b44de0669dd8cf))

Minor changes
- Merge conflicts, a large portion of them in `creation.rs` ([42f25c2](https://github.com/open-rmf/rmf_site/pull/242/commits/42f25c2421cf943aa8f5a73317b014655bcf32bc))
- Model instance creation - modified further after clearing merge conflicts. Needs to be reviewed again after workcell editor PR is merged ([bb13ee0](https://github.com/open-rmf/rmf_site/pull/242/commits/bb13ee0db9be2ee38c79aced969c0d8dd2868e45), [e2b365c](https://github.com/open-rmf/rmf_site/pull/242/commits/e2b365cdcc74f368a17fd7428b23304c0a5e1ac0), [d450cd7](https://github.com/open-rmf/rmf_site/pull/242/commits/d450cd726450b54e76669c885a39a3261edcde08))
- If a description is no longer selected or hovered over, remove the description entity (if any) from the relevant model instance `support_selected`/`support_hovered` instead of clearing all entities that the model instance supports ([afddc82](https://github.com/open-rmf/rmf_site/pull/242/commits/afddc8240c3b4d7e278549366966aff97790744c))
- Simplify Model Instance creation - there are already a few ways to create a new Model Instance, limiting users to either spawn them from existing Model Descriptions under `Groups` or spawning one when loading a new Model Description ([8ffcfe6](https://github.com/open-rmf/rmf_site/pull/242/commits/8ffcfe66565123c681222ab67f2d90ae53589682))
- Fixed bug where scenarios from a previously loaded site still remains visible in the Scenarios widget ([5786bae](https://github.com/open-rmf/rmf_site/pull/242/commits/5786bae964da80f004c89759665a7d0bfb547115))
- Align delete button for scenarios so that it's visible within the widget ([ad5ba6e](https://github.com/open-rmf/rmf_site/pull/242/commits/ad5ba6ef95d6a4db63f7080b3ba863f21c5ec5b8))
- Send a `ChangeCurrentScenario` event when adding new levels, despite the scenario not changing, to trigger visibility updates for all model instances. Otherwise when creating a new level, the model instances from before still remains visible in the editor unless the user toggles between scenarios ([6ad0ea7](https://github.com/open-rmf/rmf_site/pull/242/commits/6ad0ea7c26fdf9bc068f81d5d13dc413c71a7f2c))
- Minor cleanup ([4729bef](https://github.com/open-rmf/rmf_site/pull/242/commits/4729befe340a57964f3b3a3d73f40a82f5d6dc91), [1d8aa7d](https://github.com/open-rmf/rmf_site/pull/242/commits/1d8aa7dc11b68c2350adfe8b9dcc8d4aa705befa), [af5a2d5](https://github.com/open-rmf/rmf_site/pull/242/commits/af5a2d5df97768603389dcc3e66c1a34a712c460), [c6eebf4](https://github.com/open-rmf/rmf_site/pull/242/commits/c6eebf44edc36ab6354bf0293724e70e2ed98482), [f0fa12c](https://github.com/open-rmf/rmf_site/pull/242/commits/f0fa12c18bf2fb6d4a71fd81294f617306126581), [d331a79](https://github.com/open-rmf/rmf_site/pull/242/commits/d331a79cf8d12ffc6c234c9be1833daf194ee98f), [7b01c91](https://github.com/open-rmf/rmf_site/pull/242/commits/7b01c911030384f23342515625ca52dcba4e1762), [83f3d85](https://github.com/open-rmf/rmf_site/pull/242/commits/83f3d85629e965dae0827c29e995828f7d73511a), [de9b60e](https://github.com/open-rmf/rmf_site/pull/242/commits/de9b60e1286c4a131e0536821bf37660a25bb4ba), [90a60ba](https://github.com/open-rmf/rmf_site/pull/242/commits/90a60ba7d4b62bfa9dd4cec3ea1a9934aa2f6aa4), [8c19eed](https://github.com/open-rmf/rmf_site/pull/242/commits/8c19eedce28a02b6bb8db4bcf91f0aa59e88ec87))

Notes:
- Known bug: Right now with the demo map, when expanding the `Added` dropdown in the Scenarios widget, there will be constant error messages being printed as the editor is unable to locate assets for the TeleportIngestor and TeleportDispenser. The editor will also crash if we try to delete the Default scenario, since these entities are invalid. This is only specific to the demo map.

xiyuoh:
I've found that when calculating grid occupancy, the physical entities used for calculation are mostly meshes instead of model instance entities. This meant if a user wishes to ignore a set of model instances, their meshes would still cause the space to be occupied. [5aed94f](https://github.com/open-rmf/rmf_site/pull/242/commits/5aed94fa759251ab28d228064a5177e4897f4b54) and [3a6ad1d](https://github.com/open-rmf/rmf_site/pull/242/commits/3a6ad1d9223c49554db92df455d7b491fe95f280) tackle that by ensuring that meshes that are children of an ignored entity would also be ignored.

xiyuoh:
Latest changes:
- Enabled saving to file for model properties by introducing the `OptionalModelProperties` component. It currently supports all available optional properties: `DifferentialDrive`, `MobileRobotMarker` for model descriptions, and `Tasks` created for model instances. ([4252c38](https://github.com/open-rmf/rmf_site/pull/242/commits/4252c3878bcf4ad675fb251584693a46d9cbebbb), [bbc373c](https://github.com/open-rmf/rmf_site/pull/242/commits/bbc373c22ed7531acd9618ba94fe478b7dabea56))
- Resolve merge conflicts from recent workflows PR ([766fd62](https://github.com/open-rmf/rmf_site/pull/242/commits/766fd62b0578944808f8a1b1cd98c615dbcc5b1d))
- Other minor cleanup ([cc29375](https://github.com/open-rmf/rmf_site/pull/242/commits/cc2937574fd367eb175acc4803968336209b26f6), [01fca32](https://github.com/open-rmf/rmf_site/pull/242/commits/01fca322526297ae674c8ab079137a9d470e87dc), [d97232d](https://github.com/open-rmf/rmf_site/pull/242/commits/d97232d57a9e49f7cef94f851fa7e0aea6b182ee))

TODO fix CI, though I foresee some refactoring through the review process

rmf_site_format/src/model.rs
```diff
@@ -57,3 +58,85 @@ impl Default for Model {
...
+/// Defines a property in a model description, that will be added to all instances
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq)]
+#[cfg_attr(feature = "bevy", derive(Component, Reflect))]
+pub struct ModelProperty<T: Default + Clone>(pub T);
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum OptionalModelProperty {
+    DifferentialDrive(DifferentialDrive),
+    MobileRobotMarker(MobileRobotMarker),
+    Tasks(Tasks<Entity>),
```
luca-della-vedova, line 71: I think this is making CI fail, the moment we compile `rmf_site_format` without the `bevy` feature `Entity` will not be defined, you can try this locally by just running `cargo build` in the `rmf_site_format` folder.
I don't know the details of the `Tasks` implementation but usually if we need to reference an entity in a "non bevy specific way" in `rmf_site_format` we just use `RefTrait`:
https://github.com/open-rmf/rmf_site/blob/7d21b4881ef4aeb251e93733388117c2a3275ab2/rmf_site_format/src/site.rs#L224

Which is a trait for references and is implemented for `Entity` (for usage in site editor) and `u32` (for usage in serialization / deserialization, as well as not having a bevy dependency).

Now the problem with this approach is that it will make `OptionalModelProperty` generic and that will propagate all the way up to the `ModelDescriptionBundle`, which might not be what you want

luca-della-vedova: But again, maybe having to do `ModelDescriptionBundle<u32>` instead of `ModelDescriptionBundle` wouldn't be too bad

xiyuoh: Ah thank you for pointing this out! I couldn't figure out the underlying issue behind using `Entity`. I can see another (and maybe better) way to store location information in tasks without using `Entity`, especially since the current method to save tasks involve exporting the entity ID that could change between runs. Let me make that change.

luca-della-vedova: Yap that's the classic use for `RefTrait`, to use both `Entity` and `u32`, usually the `u32` that is saved / loaded is the value that is contained in the `SiteID` component

luca-della-vedova:
Also the elephant in the room while I was giving this a spin, _where are the robots_?

### My expectation

What I would have expected from this PR is, when a world is loaded, models are converted into instances (and I see that) and robots are also correctly imported and converted into instances and that doesn't seem to be the case.
Specifically, there is currently a [LocationTag](https://github.com/open-rmf/rmf_site/blob/983cd3c943e47924c5cf436e30becb3d4cb357c3/rmf_site_format/src/location.rs#L29) for robot spawning which, similar to what used to be in the legacy traffic editor, contains information on what robot is spawned in the location.

image

My _expectation_ (not sure that is the intention of this PR) would have been to see that either:

* The tag is removed, since it's not relevant anymore.
* The robot is imported and spawned as an instance in the position of the anchor as a child of the level.

OR

* The tag is kept and used, maybe changed to point to a group rather than a `Model`.
* The model is spawned as an instance with some sort of affiliation to the location, so that when the location is moved / deleted the robot is also affected.

The first implementation followed the second approach, but only for historical reasons and I don't have strong feelings on what is right (it seems the current implementation would do the first instead).
One thing I personally like about connecting the robot to the location is that if the location was moved around the robot would also be moved around, which can be quite convenient when the location is a charger (so we can always start the simulation with a robot in its charger) but again I'm fairly neutral.

### What I actually see

What I see happening instead is that the location tag still exists and contains information about what robot is spawned in that location. The information is however ignored and the demo map by default contains no robot. Robots that are imported will use purely the model instance components creating a bit of a double source of truth conflict with the location tag

rmf_site_format/src/model.rs
xiyuoh: Fixed in [347d546](https://github.com/open-rmf/rmf_site/pull/242/commits/347d546e4b8e522622464d93a069665a7f76d5a6)

rmf_site_editor/src/interaction/model.rs
```diff
@@ -0,0 +1,98 @@
...
+use crate::{interaction::*, site::*};
+use bevy::prelude::*;
+
+pub fn update_model_instance_visual_cues(
+    model_descriptions: Query<
+        (Entity, &Selected, &Hovered),
+        (
+            With<ModelMarker>,
+            With<Group>,
+            Or<(Changed<Hovered>, Changed<Selected>)>,
```
mxgrey, line 27: In the current setup, we iterate over `&mut model_instances` which does not have this filter, meaning it will include **all** model instances in the world every update cycle. Moreover, this query will **only** include model descriptions whose `Hovered` or `Selected` status has changed during the latest cycle, so [this line](https://github.com/open-rmf/rmf_site/pull/242/files#diff-89f10a882e59e017aa0faa7062467dec447e56c2b827348b80b029fa3668243cR49) will not fetch anything if neither of these components changed for the model description.

I have a feeling this filter was meant to go on `model_instances`, not on `model_descriptions`.

rmf_site_format/src/sdf.rs
```diff
@@ -485,50 +490,78 @@ impl Site {
             });
             // TODO(luca) We need this because there is no concept of ingestor or dispenser in
             // rmf_site yet. Remove when there is
-            for (model_id, model) in &level.models {
+            for (model_instance_id, _) in &default_scenario.scenario.added_instances {
+                let model_instance = self
+                    .model_instances
+                    .get(model_instance_id)
+                    .expect("Model instance not found");
```
mxgrey, line 497: Don't use `.expect`.

mxgrey (changes requested):
Thanks @xiyuoh for this massive effort of refining this PR (and special thanks to @reuben-thomas for kicking this off months ago).

I think we're close to a good first draft of this feature set. I've left some inline comments for things that I believe should be changed before we merged.

I've also opened issue #254 to track bigger picture conceptual and architectural recommendations that I think we should address with follow-up PRs after we merge this one.

assets/demo_maps/office.building.yaml
```diff
@@ -181,7 +181,7 @@ levels:
       - [814.828, 239.25, 0, "", {is_parking_spot: [4, false]}]
       - [1210.544, 365.254, 0, ""]
       - [1191.442, 824.457, 0, ""]
-      - [1232.421, 658.567, 0, tinyRobot1_charger, {is_charger: [4, true], is_holding_point: [4, true], is_parking_spot: [4, true], spawn_robot_name: [1, tinyRobot1], spawn_robot_type: [1, TinyRobot]}]
+      - [1232.421, 658.567, 0, tinyRobot1_charger, {is_charger: [4, true], is_holding_point: [4, true], is_parking_spot: [4, true], spawn_robot_name: [1, tinyRobot1], spawn_robot_type: [1, Luca/TinyRobot]}]
```
luca-della-vedova, line 184: I'm not 100% convinced by this change.
At least for me, the idea of this demo map is that it is as close as possible to a legacy traffic editor world and shows how we can easily import models from the legacy format. Doing this change here will make this specific map work but every other traffic editor map still won't have any robots.
What I would propose instead is to add the fuel organization when the legacy map is converted to a site. Now the question will be "which organization"? The answer is "it's complicated" but for now I would just add an `OpenRobotics/` in front of every loaded model. There does seem to be an [OpenRobotics/TinyRobot](https://app.gazebosim.org/OpenRobotics/fuel/models/TinyRobot) model on fuel anyway. In the future if we can't upload models to `OpenRobotics` we can just change to a different organization

luca-della-vedova:
After #239 was merged there are (as predicted) a lot of conflicts.
My first question here is "how much was the workcell editor / 3d object placement tested?", depending on how close to working it was it might be good to port the changes to the now deleted files and open a draft PR to [rmf_workcell](https://github.com/luca-della-vedova/rmf_workcell/pull/1), but if it was only the bare minimum to make it compile without really any functionality it's probably not so necessary.
And the second question is also related, was there any functionality that was only kept specifically because the workcell editor was in so we wouldn't break it? I don't see much diff in the workcell editor mode so I suspect there is no concept of scenario there, so either spawning models is broken or there is some legacy path that we can clean up, perhaps as a followup PR.

mxgrey:
> depending on how close to working it was it might be good to port the changes to the now deleted files and open a draft PR to https://github.com/luca-della-vedova/rmf_workcell/pull/1

I think we treat these new concepts of instances and scenarios as experimental for right now and expect that there will be substantial iteration on them in the time to come. I've already made an [outline of next steps](https://github.com/open-rmf/rmf_site/issues/254), and I wouldn't be surprised if we end up with more changes than that as we get some hands-on experience with these capabilities.

With that in mind, I don't want to be in a position where we're awkwardly needing to keep the two repos in sync, so I'd suggest we make a tag for the current main branch, and bump the minor version before we merge these new features.

Once we've battle hardened and stabilized these capabilities it'll make more sense to think about the right way to introduce it to workcell editing.

assets/demo_maps/office.building.yaml
mxgrey: I was meaning to ask, how did we end up with a situation where [decorative models use `OpenRobotics/`](https://github.com/open-rmf/rmf_demos/blob/9bdd28ff2d326634c27536a2cd120290f15c4356/rmf_demos_maps/maps/office/office.building.yaml#L113) but [robots don't](https://github.com/open-rmf/rmf_demos/blob/9bdd28ff2d326634c27536a2cd120290f15c4356/rmf_demos_maps/maps/office/office.building.yaml#L113)?

I assume it has something to do with pit crew and/or fuel versus [`rmf_demos_assets`](https://github.com/open-rmf/rmf_demos/tree/main/rmf_demos_assets/models)? I'm going to guess the rationale is that decorative models are stable enough that fetching them from fuel will generally be fine, whereas robot models may need to have their plugin parameters tweaked, e.g. changing speeds or bidirectionality for the slotcar plugin?

If that's the case then I'd suggest we add the following as an action item for #254 :
1. Upload all the models from `rmf_demos_assets` to the `OpenRobotics` as @luca-della-vedova suggested
2. When loading a model, check:
    - Does the loaded SDF model contain a slotcar plugin?
      - If it does have a slotcar plugin, check if the entity already has a `Mobility` component
        - If it does already have a `Mobility` component, do nothing
        - If it doesn't already have a `Mobility` component, check if it has an `IsStatic` component
          - If it does have an `IsStatic` component, do nothing
          - If it doesn't have an `IsStatic` component, add a `Mobility` component with a `DifferentialDrive` configuration
      - If it does not have a slotcar plugin, do nothing

In the meantime for this PR I think it's enough to just change this to `OpenRobotics/TinyRobot` for the demo site.

luca-della-vedova: > In the meantime for this PR I think it's enough to just change this to `OpenRobotics/TinyRobot` for the demo site.

Fully agree with the analysis (including possible historical causes) and outlined solution except this teeny tiny point!
My proposal was to leave it as is on the `.building.yaml` level and add a conversion step that prepends `OpenRobotics/` to the value under `spawn_robot_type` so it would import a legacy `spawn_robot_type: [1, TinyRobot]` into a `AssetSource::Search(OpenRobotics/TinyRobot)`.
The main rationale is that this import strategy would make it possible to automatically import every legacy project out there. By contrast, if we make it necessary to prepend the organization name to the asset name (i.e. what this diff is doing) legacy projects will fail out of the box to fetch models and will have to be updated.

It's basically a tradeoff of:
* Updating our pipeline(s) so they work with existing maps (my proposal) or
* Force users to update existing maps so they work with our pipelines (this diff)

mxgrey: > My proposal was to leave it as is on the .building.yaml level and add a conversion step that prepends OpenRobotics/ to the value under spawn_robot_type

Ah yeah, that should give the smoothest outcome for the use cases that we're anticipating. I'd be a little concerned that it could be disruptive for anyone using their own custom robot models.

I don't love what I'm about to suggest, but what if we hard-code the names of the models that we know are hosted at the `OpenRobotics` org and only prepend `OpenRobotics/` to those?

mxgrey: Or a more complex but even more backwards compatible alternative:

Introduce a `spawn-robot` scheme that we apply during legacy imports which behaves similarly to `search` except that when the local search fails it will then add `OpenRobotics/` and search fuel.

luca-della-vedova: We could also add an additional searching iteration for `Search` assets, this is the current sequence:
https://github.com/open-rmf/rmf_site/blob/fe86373b94621b22c482bebd0b6333da2e8ff496/rmf_site_editor/src/site_asset_io.rs#L228-L232

We could extend the search relative to `MODEL_ENVIRONMENT_VARIABLE` (which is also what is used by Gazebo to look for models) to try the path without the organization, if an organization is present.
This means that if an `OpenRobotics/Table` is requested, the `Search` will look for both a `[GZ_SIM_PATH]/OpenRobotics/Table` as well as `[GZ_SIM_PATH]/Table`.
This could be undesirable if users have both but I would argue that at that point it is unclear to me how Gazebo itself would behave, since it would mean there are duplicated models in its path.
This search would still have priority and happen before the Fuel search and hopefully be fast enough (it's only a local file lookup) that it won't generally be noticeable.

mxgrey: A part of me is scared of adding more complexity to `search` because it could reach a point where the rules become very unclear to users and it becomes a cargo cult of people trying different weird things until they can get the search to work. But I'm open to that approach for the sake of helping existing users transition.

rmf_site_editor/src/interaction/model.rs
xiyuoh: My understanding of this implementation is to have the model instances highlighted if their affiliated description is selected/hovered over, rather than the other way round. The current app supports that, such that a change in selection/hovering for a description (queried with the `Changed<>` filter) will trigger an update to all affiliated instances being highlighted (query of all instances in each cycle).

From what I see this behavior also allows relevant task locations to be highlighted when we iterate over every model.

rmf_site_editor/src/site/model.rs
```diff
@@ -332,6 +337,32 @@ impl<'w, 's> ModelLoader<'w, 's> {
         })
     }
 
+    /// Spawn a new model instance and begin a workflow to load its asset source.
+    /// This is only for brand new models does not support reacting to the load finishing.
+    pub fn spawn_model_instance(
```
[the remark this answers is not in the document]
xiyuoh: Thanks for the suggestion. `ModelProperty<AssetSource>` not being available yet was exactly my struggle that led to this approach. I'll work on updating this to use the recommended alternative.

rmf_site_editor/src/interaction/model.rs
```diff
@@ -0,0 +1,98 @@
...
+            if tasks.is_changed() {
+                for mut location_selected in locations.iter_mut() {
+                    location_selected.support_selected.remove(&instance_entity);
+                }
+            }
+
+            if let Some(task_location) = tasks.0.first().and_then(|t| t.location()) {
+                if let Ok(mut location_selected) = locations.get_mut(task_location.0) {
+                    if instance_selected.cue() && !is_description_selected {
+                        location_selected.support_selected.insert(instance_entity);
```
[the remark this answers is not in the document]
xiyuoh: [57436ed](https://github.com/open-rmf/rmf_site/pull/242/commits/57436edc59a006c1c3e9595ff9bca056ff608fa2)

rmf_site_editor/src/occupancy.rs
```diff
@@ -180,6 +182,32 @@ fn calculate_grid(
...
+                .is_some_and(|ignored_set| ignored_set.contains(e))
+            {
+                continue;
+            } else if request.ignore.as_ref().is_some() {
+                let mut ignore: bool = false;
+                let mut queue = vec![*e];
+                // Some of these physical entities may be meshes of ignored entities.
+                // Here we check if their parent entities have been indicated to be
+                // ignored, and handle them accordingly.
+                while let Some(entity) = queue.pop() {
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_editor/src/occupancy.rs
```diff
@@ -180,6 +182,32 @@ fn calculate_grid(
         let physical_entities = collect_physical_entities(&bodies, &meta);
         info!("Checking {:?} physical entities", physical_entities.len());
         for e in &physical_entities {
+            if request
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_editor/src/site/model.rs
```diff
@@ -659,3 +705,61 @@ pub fn propagate_model_property<Property: Component + Clone + std::fmt::Debug>(
...
+        }
+    }
+
+    // Changes
+    for (instance_entity, affiliation) in model_instances.iter() {
+        if let Some(description_entity) = affiliation.0 {
+            if let Ok((_, _, property)) = model_properties.get(description_entity) {
+                if property.is_changed() || affiliation.is_changed() {
+                    let mut cmd = commands.entity(instance_entity);
+                    cmd.remove::<ModelProperty<T>>();
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_editor/src/widgets/inspector/mod.rs
```diff
@@ -207,9 +215,18 @@ impl Plugin for StandardInspectorPlugin {
                 InspectionPlugin::<InspectJoint>::new(),
                 InspectionPlugin::<InspectMeasurement>::new(),
                 InspectionPlugin::<InspectPhysicalCameraProperties>::new(),
-                InspectLiftPlugin::default(),
-                InspectionPlugin::<InspectPreview>::new(),
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_editor/src/site/site.rs
```diff
@@ -135,6 +148,30 @@ pub fn change_site(
...
+                if previous_scenario != new_scenario {
+                    change_current_scenario.send(ChangeCurrentScenario(new_scenario));
+                }
+            }
+        } else {
+            if let Ok(children) = children.get(cmd.site) {
+                let any_scenario = children
+                    .iter()
+                    .filter(|child| scenarios.get(**child).is_ok())
+                    .find_or_first(|_| true);
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_editor/src/site/scenario.rs
```diff
@@ -0,0 +1,303 @@
...
+    mut change_current_scenario: EventReader<ChangeCurrentScenario>,
+    mut current_scenario: ResMut<CurrentScenario>,
+    current_workspace: Res<CurrentWorkspace>,
+    scenarios: Query<&Scenario<Entity>>,
+    mut instances: Query<
+        (Entity, &mut Pose, &SiteParent<Entity>, &mut Visibility),
+        With<InstanceMarker>,
+    >,
+) {
+    for ChangeCurrentScenario(scenario_entity) in change_current_scenario.read() {
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_format/src/sdf.rs
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

rmf_site_format/src/sdf.rs
```diff
@@ -434,6 +434,11 @@ impl Site {
             filename: "toggle_floors".into(),
             ..Default::default()
         };
+        // Only export default scenario into SDF for now
+        let (_, default_scenario) = self
+            .scenarios
+            .first_key_value()
+            .expect("No scenarios found");
```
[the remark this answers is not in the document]
xiyuoh: [a2e2ef5](https://github.com/open-rmf/rmf_site/pull/242/commits/a2e2ef5061c6d2ea64a51136c6cf97dafda11c52)

I was wondering if it is necessary to also remove the `.expect()` found here. I took a stab at it but could not find an elegant way to raise errors without making multiple loops over each vector, mainly because each element is a tuple (member variables are `Vec<(T, Pose)>`), so I had trouble converting them into `Result<Vec<(T, Pose)>>` with `Pose` in the mix.

rmf_site_editor/src/site/save.rs
```diff
@@ -1181,6 +1197,216 @@ fn migrate_relative_paths(
...
+        instance_parent,
+        instance_affiliation,
+    ) in model_instances.iter()
+    {
+        if instance_parent
+            .0
+            .is_some_and(|p| !site_levels_ids.contains_key(&p))
+        {
+            continue;
+        }
```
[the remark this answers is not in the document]
xiyuoh: [eb49604](https://github.com/open-rmf/rmf_site/pull/242/commits/eb496040010b8da62d2c7b281a6ef72f926357cc)

rmf_site_editor/src/site/load.rs
```diff
@@ -338,6 +334,104 @@ fn generate_site_entities(
...
+            {
+                let model_instance_entity = loaded_instance
+                    .insert((Category::Model, SiteID(*model_instance_id)))
+                    .id();
+                id_to_entity.insert(*model_instance_id, model_instance_entity);
+                consider_id(*model_instance_id);
+
+                model_description_dependents
+                    .entry(model_description)
+                    .and_modify(|hset| {
```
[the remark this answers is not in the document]
xiyuoh: [4816d02](https://github.com/open-rmf/rmf_site/pull/242/commits/4816d02bc32d24f5421db878f542b00981a1b62c)

assets/demo_maps/office.building.yaml
luca-della-vedova: After some more thought I agree that all these solutions are a bit messy and probably unifying the robot type parameters going forward is the best solution.
I uploaded all the demo robots to a new [Open-RMF organization](https://app.gazebosim.org/Open-RMF), as well as opened an upstream [PR](https://github.com/open-rmf/rmf_demos/pull/272) in `rmf_demos` to change the demo maps going forward to use the same convention as models.
This will mean that pre-existing third party maps might have import issues but at least `rmf_demos` and `rmf_site` will be in a consistent state.
For this specific PR, as long as the `Luca/` changes to `Open-RMF/` it should all be good

mxgrey: I'm a little bit hesitant to remove `rmf_demos_assets` entirely...

1. Where are we actually version controlling those models? Will the copies on fuel be the only source? Is that the workflow recommended by the Gazebo team?
2. traffic-editor doesn't currently have a reasonable way to override the slotcar plugin properties, so traffic-editor users will be locked into the current slotcar parameters for the models on fuel

If it's possible for the `rmf_demos_assets` models to be installed locally in a way that they could be found via both searches for `Open-RMF/TinyRobot` and searches for `TinyRobot` then I think that would offer a cleaner transition.

This would be a good topic to discuss at the PMC session tomorrow morning.

rmf_site_format/src/sdf.rs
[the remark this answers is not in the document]
mxgrey: > I was wondering if it is necessary to also remove the .expect() found here.

Did you mean to link to somewhere else in the code? I see that you've removed the `.expect(~)` that my inline comment was referring to, so I don't think your question is related to this one.

rmf_site_format/src/scenario.rs
```diff
@@ -0,0 +1,154 @@
...
+            parent_scenario: Affiliation::default(),
+            added_instances: Vec::new(),
+            removed_instances: Vec::new(),
+            moved_instances: Vec::new(),
+        }
+    }
+}
+
+impl<T: RefTrait> Scenario<T> {
+    pub fn convert<U: RefTrait>(&self, id_map: &HashMap<T, U>) -> Result<Scenario<U>, T> {
```
mxgrey, line 67: I wonder if [this question](https://github.com/open-rmf/rmf_site/pull/242#discussion_r1898459969) was asking about the `.expect`s in this function?

If so, here's what I would recommend:

```rust
        Ok(Scenario {
            parent_scenario: self.parent_scenario.convert(id_map)?,
            added_instances: self
                .added_instances
                .clone()
                .into_iter()
                .map(|(id, pose)| {
                    let converted_id = id_map.get(&id).cloned().ok_or(id)?;
                    Ok((converted_id, pose))
                })
                .collect::<Result<_, _>>()?,
            removed_instances: self
                .removed_instances
                .clone()
                .into_iter()
                .map(|id| id_map.get(&id).cloned().ok_or(id))
                .collect::<Result<_, _>>()?,
            moved_instances: self
                .moved_instances
                .clone()
                .into_iter()
                .map(|(id, pose)| {
                    let converted_id = id_map.get(&id).cloned().ok_or(id)?;
                    Ok((converted_id, pose))
                })
                .collect::<Result<_, _>>()?,
        })
```

assets/demo_maps/office.building.yaml
luca-della-vedova: Yea I can see the controversy, let's discuss the options in the PMC. For this PR though @xiyuoh can you just remove `Luca` and change to `Open-RMF`? We should do that regardless of what we end up doing with `rmf_demos_assets`

rmf_site_format/src/scenario.rs
xiyuoh: Yes it is! Thanks for sharing.

rmf_site_format/src/sdf.rs
```diff
@@ -781,40 +821,6 @@ impl Site {
...
-                ..Default::default()
-            };
-            world.include.push(SdfWorldInclude {
-                uri: "model://".to_string() + robot_type,
-                name: Some(robot.name.0.clone()),
-                pose: Some(pose.to_sdf()),
-                r#static: Some(robot.is_static.0),
-                ..Default::default()
-            });
-        }
```
luca-della-vedova, line 817: Note, this will break SDF export since going forward maps will not have robots in them anymore, or to be more specific, robots will be there but they won't have robot properties (before we were just including the whole SDF tag).
This is a fairly complex topic though, I'll add it to the followup PR and not hold it as a blocker for this PR.

rmf_site_format/src/sdf.rs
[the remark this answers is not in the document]
xiyuoh: Ah I didn't realize I missed out on attaching the link, but I think you found what I was referring to. Thanks!

rmf_site_editor/src/site/model.rs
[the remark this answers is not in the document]
xiyuoh: [1bb13d0](https://github.com/open-rmf/rmf_site/pull/242/commits/1bb13d0ae0f29c9cbb1ed1caca37dbc1fc2f718b)

rmf_site_format/src/misc.rs
```diff
@@ -488,14 +522,14 @@ pub struct Pending;
 pub struct Original<T>(pub T);
 
 /// Marks that an entity represents a group
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
```
luca-della-vedova, line 525: What started as a "Why are we deriving `Eq` for a unit struct" ended up in an interesting journey. I feel this was done to please the compiler for the `skip_serializing_if` function in the [Site](https://github.com/open-rmf/rmf_site/blob/1bb13d0ae0f29c9cbb1ed1caca37dbc1fc2f718b/rmf_site_format/src/site.rs#L146-L147). If the function is changed to just be the same as the rest of the structs in the site `BTreeMap::is_empty` this can be reverted

tinyrobot.site.ron
luca-della-vedova, line 1: Is this committed by accident?

tinyrobot_new.site.ron
luca-della-vedova, line 1: Is this also committed by accident?

rmf_site_editor/src/site/model.rs
```diff
@@ -368,6 +432,23 @@ impl<'w, 's> ModelLoader<'w, 's> {
...
+                if entity == description_entity {
+                    instance_entities.insert(e);
+                }
+            }
+        }
+        for e in instance_entities.iter() {
+            self.update_asset_source_impulse(*e, source.clone())
+                .detach();
+        }
+    }
```
luca-della-vedova, line 451: Can this be simplified to remove the intermediate `HashSet` and just fire the workflow inside the first for loop? From a glance it doesn't seem there should be cases where duplicated entities are added (which would justify collecting into a hash set)

rmf_site_editor/src/site/model.rs
```diff
@@ -659,3 +778,60 @@ pub fn propagate_model_property<Property: Component + Clone + std::fmt::Debug>(
         }
     }
 }
+
+/// This system keeps model instances up to date with the properties of their affiliated descriptions
+pub fn update_model_instances<T: Component + Default + Clone>(
+    mut commands: Commands,
+    model_properties: Query<
+        (Entity, &NameInSite, Ref<ModelProperty<T>>),
```
luca-della-vedova, line 786: `Entity, &NameInsite` are unusued and can be removed from the query

tinyrobot_new.site.ron
xiyuoh: Yes, removed

rmf_site_editor/src/widgets/view_levels.rs
```diff
@@ -103,6 +105,12 @@ impl<'w, 's> ViewLevels<'w, 's> {
                         ))
                         .id();
                     self.current_level.0 = Some(new_level);
+                    // We trigger a ChangeCurrentScenario despite staying in the
+                    // same scenario to trigger visibility updates on all model instances
+                    if let Some(current_scenario) = self.current_scenario.0 {
+                        self.change_current_scenario
+                            .send(ChangeCurrentScenario(current_scenario));
```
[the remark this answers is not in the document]
xiyuoh: I played around with this and I agree it seems unnecessary, removed in [a850590](https://github.com/open-rmf/rmf_site/pull/242/commits/a8505907321595432a0cf4bb849024d79bcf138a)

rmf_site_format/src/misc.rs
```diff
@@ -441,10 +441,44 @@ pub struct PreviewableMarker;
...
-#[derive(Clone, Copy, Debug)]
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
 #[cfg_attr(feature = "bevy", derive(Component, Deref, DerefMut))]
 pub struct SiteID(pub u32);
 
+/// This component is applied to an entity as a reference to its parent entity.
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(transparent)]
+#[cfg_attr(feature = "bevy", derive(Component, Reflect))]
+pub struct SiteParent<T: RefTrait>(pub Option<T>);
```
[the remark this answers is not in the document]
xiyuoh: I might have found a case for having a `SiteParent` component, unsure if it was the original intention: when saving model instances to file, we need to export their parent entities _if_ they exist (i.e. which level the model instance is spawned on). This data is stored as an `Option`, which is not a valid component for the `ModelInstance` bundle. This is where having a `SiteParent` wrapping the `Option` could be useful.

Additionally, the `SiteParent` was never used when loading instances. When spawning instance entities, the parents are always set to the `Site` entity in `load.rs` first, but later updated by checking the `SiteParent` data in [this system](https://github.com/open-rmf/rmf_site/blob/xiyu/scenarios/rmf_site_editor/src/site/model.rs#L816). I have since modified it to use the imported `SiteParent` directly during loading in [a850590](https://github.com/open-rmf/rmf_site/pull/242/commits/a8505907321595432a0cf4bb849024d79bcf138a).

The follow-up for this could be to do away with the `Option` and make it mandatory for all instances to have a `Parent`? Doing away with orphan instances in files.

assets/demo_maps/office.building.yaml
xiyuoh: [5ce02a1](https://github.com/open-rmf/rmf_site/pull/242/commits/5ce02a131604dd8c4fe1eb0f5448310f85360d13)

rmf_site_format/src/scenario.rs
xiyuoh: [812c183](https://github.com/open-rmf/rmf_site/pull/242/commits/812c183d415550ffe9152cebc083bb1dbd4b570b)

rmf_site_editor/src/site/model.rs
xiyuoh: I was unable to borrow a mutable `self` for `self.update_asset_source_impulse` while iterating over `self.model_instances`

rmf_site_editor/src/site/model.rs
xiyuoh: [661ecb9](https://github.com/open-rmf/rmf_site/pull/242/commits/661ecb96697def25d13fe57e9c252a4ecd388c91)

rmf_site_editor/src/interaction/cursor.rs
```diff
@@ -120,6 +120,26 @@ impl Cursor {
...
+                    .id(),
+            )
+        } else {
+            None
+        };
+    }
+
     pub fn set_model_preview(
         &mut self,
         commands: &mut Commands,
```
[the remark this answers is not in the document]
xiyuoh: [f527790](https://github.com/open-rmf/rmf_site/pull/242/commits/f52779060abdfd2a4162bbc539d1f6c1d3212b09), I removed the now-unused `spawn_model` and `spawn_model_impulse` as well, let me know if that should be reverted for any reason

rmf_site_format/src/misc.rs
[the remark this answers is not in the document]
mxgrey: I like the sound of making it mandatory for all instances to have a `Parent`. The instance spawning workflow should generally be able to ensure a proper parent when an instance is created. In the strange case of an instance that doesn't have a `Parent`, the save function should just ignore it. Maybe we can also add a diagnostic for that case.

rmf_site_format/src/misc.rs
xiyuoh: [8c09224](https://github.com/open-rmf/rmf_site/pull/242/commits/8c09224bfc0b6b7aae1831a5a88cd3d1995e505c)

rmf_site_editor/src/site/model.rs
luca-della-vedova: Ah yes I see it now :facepalm: nevermind, thanks for clarifying!

rmf_site_editor/src/interaction/cursor.rs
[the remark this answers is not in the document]
luca-della-vedova: I think that's fair, the downstream workcell editor will probably need to be rethought a bit and it's unclear whether these function even work right now since they are not used anywhere.

rmf_site_format/src/misc.rs
luca-della-vedova: I should have brought up, the `model_instances` in Site has the same issue where it uses `is_default` instead of `BTreeMap::is_empty`

xiyuoh:
Thank you both @luca-della-vedova and @mxgrey for the extremely detailed and insightful review. I've addressed the comments and am ready for another round of lookthrough.

rmf_site_format/src/misc.rs
xiyuoh: Oh I attempted to modify that as well, but `model_instances` has a `ModelMarker` component which has the `PartialEq` trait as well. I saw that it was the original design of the struct so I left it as it is. Happy to update that if there wasn't a particular reason it is needed though

luca-della-vedova: Yea I wouldn't change existing code (at least to avoid blowing up the diff even more) but would just change https://github.com/open-rmf/rmf_site/blob/c8e1fb233c44dd96aa9e887513075f855a52b244/rmf_site_format/src/site.rs#L149

to also use `BTreeMap::is_empty`

xiyuoh: [19483c4](https://github.com/open-rmf/rmf_site/pull/242/commits/19483c4b7e7e83aa44ad3edfbeb37341f3919861)

luca-della-vedova (approved):

rmf_site_editor/src/interaction/cursor.rs
```diff
@@ -129,13 +129,12 @@ impl Cursor {
...
         self.remove_preview(commands);
         self.preview_model = if let Some(model_instance) = model_instance {
-            if let Some(mut spawn_instance) =
-                model_loader.spawn_model_instance(self.frame, model_instance.clone(), None)
-            {
-                Some(spawn_instance.insert(Pending).id())
-            } else {
-                None
-            }
+            Some(
```
mxgrey, line 131: Not a big deal but I think this can all be rewritten as

```rust
self.preview_model = model_instance
    .and_then(|model_instance| 
        model_loader.spawn_model_instance(self.frame, model_instance)
        .insert(Pending)
        .id()
    );
```

rmf_site_editor/src/site/load.rs
```diff
@@ -367,50 +367,37 @@ fn generate_site_entities(
...
+            .description
+            .0
+            .map(|e| model_description_dependents.get_mut(&e))
+            .flatten()
+        {
+            instances.insert(model_instance_entity);
+        } else {
+            error!(
+                "Model description missing for instance {}. This should \
                         not happen, please report this bug to the maintainers of \
```
mxgrey, line 390: Minor nitpick: Let's get this text aligned with the rest of the code.

rmf_site_editor/src/interaction/model.rs
mxgrey: You're absolutely right, I forgot that I had specifically asked Reuben to implement this so that hovering and selecting the group widget would have this effect.

I might suggest renaming the function to `update_visual_cues_for_model_group` to make that purpose more obvious, but it's a minor concern.

rmf_site_editor/src/site/model.rs
```diff
@@ -541,9 +553,25 @@ impl ModelLoadingServices {
...
+        let spawn_instance = app.world.spawn_workflow(|scope, builder| {
+            scope
+                .input
+                .chain(builder)
+                .then(instance_spawn_request_into_model_load_request.into_blocking_callback())
+                .connect_on_err(scope.terminate)
+                .then(load_model)
+                .map_block(|res| match res {
+                    Ok(success) => Ok(success),
+                    Err(err) => Err(InstanceSpawningError::ModelError(err)),
```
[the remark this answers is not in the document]
xiyuoh: [6067fef](https://github.com/open-rmf/rmf_site/pull/242/commits/6067feffabf66bd01dcb2e8c3537e8937bb041fb)

rmf_site_editor/src/site/load.rs
xiyuoh: [6067fef](https://github.com/open-rmf/rmf_site/pull/242/commits/6067feffabf66bd01dcb2e8c3537e8937bb041fb)

rmf_site_editor/src/interaction/cursor.rs
xiyuoh: [6067fef](https://github.com/open-rmf/rmf_site/pull/242/commits/6067feffabf66bd01dcb2e8c3537e8937bb041fb)

rmf_site_format/src/misc.rs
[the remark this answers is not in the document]
mxgrey: We can push this concern to https://github.com/open-rmf/rmf_site/pull/260 and tackle it in a follow-up PR.

mxgrey (approved):
After spending all this time reviewing the quality of the implementation in terms of its code, I finally spent some time playing around with the actual capabilities that are introduced.

It's honestly very cool. Seeing changes in parent scenarios effect all child scenarios simultaneously, and being able to make changes within child scenarios and revert them... this is a very powerful feature.

Thanks @reuben-thomas and @xiyuoh for all the tough work figuring out how to wrangle the entities and components to make this possible. And thanks @luca-della-vedova for kicking off the review with so much valuable feedback.

There's no shortage of things to follow up on from here, but this is an amazing start.

PR #247, opened 2024-12-20
Remove hard dependency on UserCameraDisplay for Interaction plugin

mxgrey:
This PR does a soft decoupling of the Interaction and Widget plugins. I can foresee downstream users not wanting the Widget plugin at all, for example if they want to embed the site editor into an Angular frame where Angular takes care of providing widgets.

I tested this by replacing [this block](https://github.com/open-rmf/rmf_site/blob/d05474aeb428f27c2ffe9fe02f026cff6c073b48/rmf_site_editor/src/lib.rs#L221-L223) with just

```
app.add_plugins( MainMenuPlugin);
```

and then running

```
cargo run --release -- assets/demo_maps/office.building.yaml
```

and using Shift + arrow keys to see that the rotation is handled correctly.


luca-della-vedova (approved):
Thanks!

PR #249, opened 2024-12-26
Don't export run_js to the library

luca-della-vedova:
## Bug fix

Depends on #253 

### Fixed bug

When developing a wasm application based on the site editor the following duplicate symbol error will be thrown by LLVM:

```
rust-lld: error: duplicate symbol: __wbindgen_describe_run_js
          >>> defined in /usr/local/google/home/lucadv/rmf_workcell/target/wasm32-unknown-unknown/release/deps/librmf_workcell_editor.librmf_workcell_editor.22daf35a81a732f8-cgu.11.rcgu.o
          >>> defined in /usr/local/google/home/lucadv/rmf_workcell/target/wasm32-unknown-unknown/release/deps/liblibrmf_site_editor-7511bce79e1a1870.rlib(librmf_site_editor-7511bce79e1a1870.librmf_site_editor.60371dd4f1e705d4-cgu.12.rcgu.o)
```

### Fix applied

I created a feature flag, enabled by default, that can be used to remove the main `run_js`. Downstream applications that want to redefine their UI can compile without the feature enabled


rmf_site_editor/Cargo.toml
```diff
@@ -51,6 +51,10 @@ ehttp = { version = "0.4", features = ["native-async"] }
 nalgebra = "0.32.5"
 anyhow = "*"
 
+[features]
+include-run-js = []
+default = ["include-run-js"]
```
mxgrey, line 56: As discussed in the PMC session, I'd suggest either
1. don't make this a default feature (meaning we need to explicitly enable the feature [here](https://github.com/open-rmf/rmf_site/blob/ae04fcbb5f5b1abfa27682ef15884afa9137ffd2/scripts/build-web.sh#L6)) or
2. make a new `rmf_site_[editor_]web` crate that's just a thin wrapper around `rmf_site_editor` with the addition of the `run_js` function.

luca-della-vedova: Made the feature default off https://github.com/open-rmf/rmf_site/pull/249/commits/7afe92f0f163c71d6c7df361e028ec7da4bf395c

mxgrey: After thinking about it some more, making it as a separate crate would have the added benefit of providing an explicit example to downstream users for how to make their own custom web app. Would that be worth the added effort?

luca-della-vedova: Done in https://github.com/open-rmf/rmf_site/pull/249/commits/320f362546a808c274d53a8c5cbe648a1fab5d4e, I was a bit on the fence on bikeshedding `rmf_site_web` vs `rmf_site_editor_web`, the second is more accurate but adds additional typing when trying to navigate the workspace with autocompletion, still went with `rmf_site_editor_web` in the end.
There is probably more splitting we could do to reduce the amount of `cfg` flags in the main app but this is the minimal "embed site editor" shim

mxgrey (approved):

PR #257, opened 2024-12-27
Early support for undo in `ChangePlugin`

arjo129:
### Implemented feature

This provides a POC version of undo for the `ChangePlugin`. We still have to implement undo for movement, adding and deleteing items. One thing to note is that the current architecture assumes a change at the individual component level. Also currently, we need `ResMut` access tot he undo buffer from all change plugins. There may come a time where we may want to change multiple components in a single operation.

#### TODO
- [ ] Deletion
- [ ] Addition of models

### DIFF
--- a/rmf_site_editor/src/site/mod.rs
+++ b/rmf_site_editor/src/site/mod.rs
@@ -111,6 +111,9 @@ pub use site_visualizer::*;
 pub mod texture;
 pub use texture::*;
 
+pub mod undo_plugin;
+pub use undo_plugin::*;
+
 pub mod util;
 pub use util::*;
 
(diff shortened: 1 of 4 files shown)


rmf_site_editor/src/site/mod.rs
```diff
@@ -212,6 +215,7 @@ impl Plugin for SitePlugin {
             ChangePlugin::<Scale>::default(),
             ChangePlugin::<Distance>::default(),
             ChangePlugin::<Texture>::default(),
+            UndoPlugin::default()
```
arjo129, line 218: TODO: Move to `lib.rs` in case we want to run in headless mode

Issue #254, opened 2024-12-26
Items to follw up on for Scenarios


mxgrey:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

This issue tracks follow-up action items for #242

- [ ] SiteParent should be replaced by querying for `Option<&Parent>`
- [ ] `OptionalModelProperties` should be refactored
  - It does not make sense to allow multiple instances of `DifferentialDrive` or `MobileRobotMarker` 
  - `MobileRobotMarker` should not need to be explicit since `DifferentialDrive` should always imply `MobileRobotMarker`
    - Consider having a single property of `Mobility { kind: String, config: serde_json::Value }`
    - Have a `Resource` containing a map from `String` (kind) to a Widget that can read and modify the `serde_json::Value`.
  - `Task` should not be a model property
    - Instead of being an enum, it should be similar to `Mobility` mentioned above, for example `Task { kind: String, config: serde_json::Value }`
    - Each task should be associated with an instance, not with a model description
      - We should have a schedule of tasks that include which instance performs it and at what time
      - This should be tied into scenarios so that each child scenario can introduce changes to the task schedule

luca-della-vedova:
> * [ ]  Automatic import/export of sdformat plugins as described [here](https://github.com/open-rmf/rmf_site/pull/242#discussion_r1898235148)
>   
>   * When loading a model, automatically convert its slotcar parameters to a `Mobility` component with a `DifferentialDrive` configuration
>   * Provide an extensible way to export instances with additional plugins
>     
>     * Add a `pub struct ExportWith(pub HashMap<String, serde_json::Value>)` component for model description entities
>     * Add a `pub struct ExportHandlers(pub HashMap<String, ExportHandler>)` resource that contains a map of export handlers
>     * When exporting an instance, apply the export handlers to it
>       
>       * Each `ExportHandler` is a `BoxedSystem<In = (Entity, serde_json::Value), Out = serde_json::Value>`
>       * For each entry in `ExportWith`, get the associated `ExportHandler` from `ExportHandlers` and pass in the `(Entity, serde_json::Value)` where the `Entity` is for the instance and the `serde_json::Value` is from the `ExportWith`.
>       * The output of the `ExportHandler` gets inserted as an element into the `<model>` for the instance
>     * Users can provide widgets to enable/disable and configure their export handlers
[...]

PR #260, opened 2025-01-03
Do not query SiteParent

xiyuoh:
This PR addresses one of the issues in https://github.com/open-rmf/rmf_site/issues/254 regarding the use of SiteParent in the site editor. Other than using it to store the parent entity on file, it should not be used in the codebase and instead we can query the built-in `Parent` directly.

This PR also makes it mandatory for each model instance to have a level entity set as its `Parent` at loading/creation, removing the need to assign it as a child to the site entity if there are no parents found. Model instances without a `Parent` will not be saved to file. Like before, any orphan instances found will be added as children to the current level entity. In the event a current level can't be found, we add a diagnostic issue.

### DIFF
--- a/rmf_site_editor/src/site/load.rs
+++ b/rmf_site_editor/src/site/load.rs
@@ -362,25 +362,26 @@ fn generate_site_entities(
         }
     }
 
-    for (model_instance_id, model_instance_data) in &site_data.model_instances {
-        let model_instance = model_instance_data
+    for (model_instance_id, parented_model_instance) in &site_data.model_instances {
+        let model_instance = parented_model_instance
+            .bundle
             .convert(&id_to_entity)
             .for_site(site_id)?;
 
-        // If there is no site parent, we do not spawn this model instance and generate
+        // The parent id is invalid, we do not spawn this model instance and generate
         // an error instead
-        let Some(parent) = model_instance.parent.0 else {
+        let Some(parent) = id_to_entity.get(&parented_model_instance.parent) else {
             error!(
-                "Site parent missing for instance {}. Unable to determine which \
+                "Parent id is invalid for instance {}. Unable to determine which \
                 level the instance should be spawned on. Please check the \
                 building or site file to ensure that every model instance's \
-                site parent is accounted for. Unable to load file.",
+                parent id is valid. Unable to load file.",
                 model_instance.name.0,
             );
             continue;
         };
         let model_instance_entity = model_loader
-            .spawn_model_instance(parent, model_instance.clone())
+            .spawn_model_instance(*parent, model_instance.clone())
             .insert((Category::Model, SiteID(*model_instance_id)))
             .id();
         id_to_entity.insert(*model_instance_id, model_instance_entity);
--- a/rmf_site_editor/src/site/load.rs
+++ b/rmf_site_editor/src/site/load.rs
@@ -367,11 +367,20 @@ fn generate_site_entities(
             .convert(&id_to_entity)
             .for_site(site_id)?;
 
+        // If there is no site parent, we do not spawn this model instance and generate
+        // an error instead
+        let Some(parent) = model_instance.parent.0 else {
+            error!(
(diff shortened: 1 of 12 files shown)


xiyuoh:
Updated to use `Parented` similar to rmf_workcell in [3077745](https://github.com/open-rmf/rmf_site/pull/260/commits/30777453acbb25f305dceb1f18d1ff84dae2e8bf) to store parent data to file, eliminating the need for `SiteParent` completely.

rmf_site_editor/src/site/model.rs
```diff
@@ -783,25 +786,42 @@ pub fn update_model_instances<T: Component + Default + Clone>(
     }
 }
 
+/// Unique UUID to identify issue of duplicated lift names
```
luca-della-vedova, line 789: Nit, update the comment after copying the snippet

rmf_site_editor/src/site/model.rs
```diff
@@ -783,25 +786,42 @@ pub fn update_model_instances<T: Component + Default + Clone>(
...
+pub const ORPHAN_MODEL_INSTANCE_ISSUE_UUID: Uuid =
+    Uuid::from_u128(0x4e98ce0bc28e4fe528cb0a028f4d5c08u128);
+
 pub fn assign_orphan_model_instances_to_level(
```
luca-della-vedova, line 793: It sounds like this function name is not accurate anymore after refactoring, it doesn't seem to assign instances to levels but instead report issues with orphan models

rmf_site_format/src/sdf.rs
```diff
@@ -497,36 +497,32 @@ impl Site {
...
+                    (
+                        model_description_id,
+                        self.model_descriptions.get(&model_description_id).ok_or(
+                            SdfConversionError::BrokenModelDescriptionReference(*model_instance_id),
+                        )?,
+                    )
+                } else {
                     continue;
-                }
+                };
```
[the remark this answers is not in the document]
xiyuoh: [a6d4139](https://github.com/open-rmf/rmf_site/pull/260/commits/a6d4139219d7eabae316a7946928e954c0163db0)

rmf_site_editor/src/site/model.rs
xiyuoh: [a6d4139](https://github.com/open-rmf/rmf_site/pull/260/commits/a6d4139219d7eabae316a7946928e954c0163db0)

rmf_site_editor/src/site/model.rs
xiyuoh: The function attempts to add any orphan model instances to the current selected level first, and only reports an issue if a current level is not available. I initially thought this would be easier for users to recover from an orphan model instance, but now that you pointed it out I think it'd be better to simply raise an issue and have them respawn the model instances on their desired level. Updated in [a6d4139](https://github.com/open-rmf/rmf_site/pull/260/commits/a6d4139219d7eabae316a7946928e954c0163db0)

rmf_site_editor/src/site/load.rs
```diff
@@ -362,16 +362,26 @@ fn generate_site_entities(
...
+        let Some(parent) = id_to_entity.get(&parented_model_instance.parent) else {
+            error!(
+                "Parent id is invalid for instance {}. Unable to determine which \
+                level the instance should be spawned on. Please check the \
+                building or site file to ensure that every model instance's \
+                parent id is valid. Unable to load file.",
+                model_instance.name.0,
+            );
+            continue;
+        };
```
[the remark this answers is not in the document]
xiyuoh: Good point. Seems like the `site` field is used for despawning entities in case of LoadSiteError, so I'll not convert it to `enum`. Providing the missing parent entity ID as broken reference should be sufficient for users to search their file for the invalid ID. Fixed in [a6d4139](https://github.com/open-rmf/rmf_site/pull/260/commits/a6d4139219d7eabae316a7946928e954c0163db0).

luca-della-vedova (approved):
Thanks for iterating!

PR #262, opened 2025-01-07
Update readme so bindgen installation matches what we need

mxgrey:
Following up on #253 this updates the README so users know which specific version of bindgen to install.


luca-della-vedova (approved):
Arr missed this, thanks!

PR #263, opened 2025-01-14
Refactor optional model properties

xiyuoh:
This PR targets https://github.com/open-rmf/rmf_site/issues/254 to refactor the current way of storing optional model description/instance data to file. The following changes aim to address the issues brought up in the ticket:

- A unified struct `Mobility` to represent mobile robots and accommodates mobility types other than the provided `DifferentialDrive`. `Mobility` is now an optional ModelProperty for model descriptions, and users can opt to select `DifferentialDrive` or add their own plugins to introduce other kinds of mobility to the editor. `Mobility` data (`kind` and `config`), along with any other optional ModelProperty we may introduce for model descriptions, are stored to file under `OptionalModelData`, which is a `serde_json::Value`.
- `Task` is no longer an enum that only supports default tasks (`GoToPlace` and `WaitFor`) in the site editor. It is similar to `Mobility` where users can create their own task kinds and add them to the editor using plugins. When creating a new `Task`, any registered Task type can be selected from the editor, and their own configuration widget will appear accordingly. `Tasks` are also stored to file under `OptionalModelData` as a `serde_json::Value`. To enable adding tasks to a model instance, users will need to toggle the `Robot` property button for its model description.
- Updated the way we store ModelPropertyData - it initially used `TypeId` to identify each ModelProperty, but it is not compatible with `serde_json`. I've since changed it to `ComponentId` instead.

Mobility configuration:

image

Task configuration:

image

Some considerations as I was working on the refactor:
- I still chose to store both `Mobility` and `Tasks` within the renamed `OptionalModelData` component (hopefully the naming is more suitable than `OptionalModelProperties`) instead of creating separate components for model descriptions/instances. Making the switching from a HashMap to `serde_json::Value` should help with increasing configurability for either case. `Mobility` is added for model descriptions and `Tasks` for model instances, like before.
- I intentionally didn't add `Mobility` or `Tasks` to the Model Description/Instance bundles respectively. To make sure that we don't have an unnecessary component required for non-mobile and non-robot models.
- I updated `MobileRobotMarker` to `RobotMarker` and retained it as a separate ModelProperty from `Mobility`. This is for future use cases where we may support non-mobile robots that are also capable of performing tasks. The `RobotMarker` property is used to allow task creation for model instances affiliated to the selected description.
- I didn't add in Task schedule / integration with scenarios as this PR is getting a little bloated already, happy to refine this refactor and bring in task schedules after.


rmf_site_format/src/mobility.rs
```diff
@@ -16,32 +16,72 @@
...
-#[cfg_attr(feature = "bevy", reflect(Component))]
-pub struct MobileRobotMarker;
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "bevy", derive(Component))]
+pub struct Mobility {
+    pub kind: String,
+    pub config: serde_json::Value,
+    pub bidirectional: bool,
+    pub collision_radius: f32,
+    pub rotation_center_offset: [f32; 2],
```
mxgrey, line 29: I think we should put all these fields (`bidirectional`, `collision_radius`, and `rotation_center_offset`) into `struct DifferentialDrive` since they might not make sense for other forms of mobility, e.g. omni wheel drive.

I'm not certain that `collision_radius` belongs inside of `Mobility` at all.. collision information probably deserves its own component. But let's keep it in `DifferentialDrive` for now until we can think it over more.

mxgrey (changes requested):
This PR is moving things in a really good direction with the extensibility it gives. But we need to make sure to find the right balance between flexibility and explicitness.

I've left some inline comments to recommend changes that I think will help lean the balance back towards more explicitness.

rmf_site_format/src/model.rs
```diff
@@ -163,7 +119,8 @@ pub struct ModelInstance<T: RefTrait> {
     pub marker: ModelMarker,
     #[serde(skip)]
     pub instance_marker: InstanceMarker,
-    pub optional_properties: OptionalModelProperties<T>,
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub optional_data: OptionalModelData,
```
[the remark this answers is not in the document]
mxgrey: I just noticed in your PR description

> I didn't add in Task schedule / integration with scenarios as this PR is getting a little bloated already, happy to refine this refactor and bring in task schedules after.

If you want to save this change for a follow-up PR that should be okay.

rmf_site_format/src/mobility.rs
xiyuoh: With reference to the comment above, if we were to simplify `RobotMarker`/`OptionalModelData` to `Robot`, `collision_radius` might go well as a field under `Robot` alongside `skills`.

mxgrey: Yeah I was thinking about what it should look like as its own thing parallel to `Mobility`.

Here's one way I can imagine it working:
* Instead of `Robot::skills` we can call it `Robot::properties`
* We introduce a `Collision`  struct similar to `Mobility` that has a `kind` and `config`
* We have a circle collision kind:

```rust
pub struct CircleCollision {
    pub radius: f32,
    pub offset: Vec2,
}
```

xiyuoh:
Thanks @mxgrey for the comments! After another round of iteration here are the main changes with the suggestions:
- `RobotMarker` + `OptionalModelData` combined as `Robot` struct with properties. `Robot` is also its own site data instead of being a field under Model Descriptions
- `Collision` separated out from `Mobility` as its own `Robot` property
- `Tasks` no longer stored as `OptionalModelData`, and instead are stored as their own site data as well
- Removed `OptionalModelData` entirely, brought in changes from https://github.com/open-rmf/rmf_site/pull/264 for a flattened hierarchy of Parented bundles. `serde(transparent)` no longer necessary.

rmf_site_editor/src/interaction/model.rs
```diff
@@ -72,34 +73,18 @@ pub fn update_model_instance_visual_cues(
         if let Some(tasks) = tasks {
```
mxgrey, line 73: Technically this isn't hurting anything, but I don't believe any of this block is needed any longer.

rmf_site_editor/src/interaction/model.rs
```diff
@@ -72,34 +73,18 @@ pub fn update_model_instance_visual_cues(
...
-                        location_hovered.support_hovering.remove(&instance_entity);
-                    }
-                }
-            }
+            // TODO(@xiyuoh) support task-based visual cues
         }
     }
 
     // When instances are removed, prevent any location from supporting them
     for removed in removed_components.read() {
```
mxgrey, line 86: I don't think this is needed any longer either.

rmf_site_editor/src/site/load.rs
```diff
@@ -348,18 +348,21 @@ fn generate_site_entities(
...
+        if let Some(model_description_entity) = id_to_entity
+            .get(robot_id)
+            .filter(|e| model_description_to_source.contains_key(*e))
+        {
+            commands
+                .entity(*model_description_entity)
+                .insert(ModelProperty(robot_data.clone()));
+        } else {
+            // TODO(@xiyuoh) consider creating an entity if not found
+            error!("Robot is pointing to a non-existent model description!");
```
mxgrey, line 364: Let's include the `robot_id` and `robot_data` values in this message.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_circle_collision.rs
```diff
@@ -0,0 +1,160 @@
...
+    else {
+        return;
+    };
+
+    let Some(circle_collision) = robot
+        .properties
+        .get(&Collision::label())
+        .and_then(|c| serde_json::from_value::<Collision>(c.clone()).ok())
+        .filter(|c| c.kind() == CircleCollision::label())
+        .and_then(|c| serde_json::from_value::<CircleCollision>(c.config().clone()).ok())
```
mxgrey, line 141: Since egui is an [immediate mode](https://en.wikipedia.org/wiki/Immediate_mode_(computer_graphics)) GUI framework, we're going to be parsing these values on every frame update. That's probably not an enormous burden given how lightweight these structs are, but I worry about the scalability in general.

We can move along with this PR as-is, but we should think about how to find the right balance with this extensible structure. For example, maybe `CircleCollision` should be a component which we add or remove from the entity as needed based on changes in the properties inside `Robot`. Then this widget can query for `CircleCollision` instead of doing this parsing.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_mobility.rs
```diff
@@ -0,0 +1,84 @@
...
+        };
+        let Ok(ModelProperty(robot)) = params.model_descriptions.get(description_entity) else {
+            return;
+        };
+        let mut new_robot = robot.clone();
+        let mobility_label = Mobility::label();
+        let mobility = robot
+            .properties
+            .get(&mobility_label)
+            .and_then(|m| serde_json::from_value::<Mobility>(m.clone()).ok());
```
mxgrey, line 68: Similar concerns about the high frequency deserialization here

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_robot_properties.rs
```diff
@@ -0,0 +1,255 @@
...
+use bevy_egui::egui::{ComboBox, Ui};
+use rmf_site_format::RobotProperty;
+use smallvec::SmallVec;
+use std::collections::HashMap;
+
+pub type ShowRobotPropertyWidgetFn = fn(&mut serde_json::Value, &mut Ui);
+
+/// This resource keeps track of all the properties that can be configured for a robot.
+#[derive(Resource)]
+pub struct RobotPropertyData(pub HashMap<String, HashMap<String, ShowRobotPropertyWidgetFn>>);
```
mxgrey, line 36: Nitpick: Maybe rename this to `RobotPropertyWidgets`. `Data` doesn't seem like the right descriptor to me.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_circle_collision.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Default)]
+pub struct InspectCircleCollisionPlugin {}
+
+impl Plugin for InspectCircleCollisionPlugin {
+    fn build(&self, app: &mut App) {
+        app.world
+            .resource_mut::<RobotPropertyData>()
+            .0
+            .get_mut(&Collision::label())
```
mxgrey, line 37: You could use

```rust
.entry(Collision::label())
.or_default()
```

here so that this doesn't fail if the entry isn't already present.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_differential_drive.rs
```diff
@@ -15,96 +15,48 @@
...
-    gizmos: Gizmos<'s>,
+#[derive(Default)]
+pub struct InspectDifferentialDrivePlugin {}
+
+impl Plugin for InspectDifferentialDrivePlugin {
+    fn build(&self, app: &mut App) {
+        app.world
+            .resource_mut::<RobotPropertyData>()
+            .0
+            .get_mut(&Mobility::label())
```
mxgrey, line 31: Could use `.entry` here as well.

mxgrey (commented):
This implementation is looking a lot more clean, very nice!

I just left a few more small nitpicks that should be easy to get through, and then a few longer term remarks to think about for a future PR.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_circle_collision.rs
xiyuoh: That's a good point. I'll follow up on this.

rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_collision.rs
```diff
@@ -0,0 +1,84 @@
...
+use bevy_egui::egui::Ui;
+
+#[derive(SystemParam)]
+pub struct InspectCollision<'w, 's> {
+    robot_property_data: ResMut<'w, RobotPropertyWidgets>,
+    model_instances: Query<
+        'w,
+        's,
+        &'static Affiliation<Entity>,
+        (With<ModelMarker>, Without<Group>, With<Robot>),
```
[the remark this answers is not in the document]
xiyuoh: This is neat! Added to https://github.com/open-rmf/rmf_site/pull/265 in [3c5ecf5](https://github.com/open-rmf/rmf_site/pull/265/commits/3c5ecf51731768c9208885feb1be6eae25433640) to avoid excessive merge conflicts.

xiyuoh:
Hi @mxgrey ! I've added the `RecallPlugin` implementation in [0e1a5c3](https://github.com/open-rmf/rmf_site/pull/263/commits/0e1a5c34616f090ee7b2644e2d0134f8e4241e13) since changes are pretty minimal. This approach makes it mandatory to implement the RecallPlugin for any RobotProperties added using `InspectRobotPropertyPlugin`. Another thing to note is that `Recall` does not need to be implemented for any RobotPropertyKinds, as changes are already tracked via the Property's serialized values, and recalled Kind components will be inserted automatically through [`update_robot_property_kind_components`](https://github.com/open-rmf/rmf_site/blob/xiyu/optional_model_properties/rmf_site_editor/src/widgets/inspector/inspect_model_description/inspect_robot_properties.rs#L364).

mxgrey:
Wow thanks for already incorporating the recall plugin!

While playing around with the recall feature I noticed some odd stuttering while toggling the property on/off. I'm not sure if it was being caused by race conditions with the recall system or simply by egui resizing the widgets, but I _think_ [2bcdfa3](https://github.com/open-rmf/rmf_site/pull/263/commits/2bcdfa3628dd5c9815327a0ea49353de6fe80274) reduced the severity of the stuttering. But even if that change has no effect, it's not bad to give a way to order systems around the recall updates.

> Another thing to note is that Recall does not need to be implemented for any RobotPropertyKinds, as changes are already tracked via the Property's serialized values

The catch here is that we won't be able to remember previous values for each kind as we switch between the kinds within one Property. In fact when I played with it just now, re-selecting `Differential Drive` from the drop-down menu when it was already selected will _reset_ the current values in the config to their default values, which I think is really not what we want.

I don't mind fixing that in a follow-up PR, but I do think this is something we should change eventually.

xiyuoh:
I see! Thanks for testing and catching that, I'll follow up on implementing RecallPlugin for Property Kinds.

mxgrey (approved):
Great improvements, thanks for the intense iteration on this!

PR #264, opened 2025-01-20
Make the inner bundle of Parented transparent

mxgrey:
A small tweak to how to things nested inside of `Parented` get serialized:

Before
```json
"166": {
    "parent": 1,
    "bundle": {
        "name": "tinyRobot2",
        "pose": {
            "trans": [
                20.423693,
                -5.312098,
                0.0
            ],
            "rot": {
                "yaw": {
                    "deg": 0.0
                }
            }
        },
        "description": 164,
        "optional_properties": []
    }
}
```

After
```json
"166": {
    "parent": 1,
    "name": "tinyRobot2",
    "pose": {
        "trans": [
            20.423693,
            -5.312098,
            0.0
        ],
        "rot": {
            "yaw": {
                "deg": 0.0
            }
        }
    },
    "description": 164,
    "optional_properties": []
}
```

This PR reduces some unnecessary struct nesting in the serialized output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -424,6 +424,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
 name = "bevy"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -503,7 +509,7 @@ dependencies = [
  "futures-lite 1.13.0",
  "js-sys",
  "parking_lot",
- "ron",
+ "ron 0.8.1",
  "serde",
  "thiserror",
  "wasm-bindgen",
@@ -1092,7 +1098,7 @@ dependencies = [
  "bevy_render",
  "bevy_transform",
  "bevy_utils",
- "ron",
+ "ron 0.8.1",
  "serde",
  "thiserror",
  "uuid",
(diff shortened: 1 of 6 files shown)


xiyuoh:
@mxgrey Thank you for taking a look at this! I experimented with it previously and the issue with flattening the Bundle content is that the fields of the bundle will be stored as strings with `""`. This results in errors when loading the same site file back in. Tested out this PR and here's the error I'm getting when loading a saved file with `serde(flatten)`:
```
2025-01-20T10:36:12.246615Z ERROR librmf_site_editor::workspace: Failed loading site SpannedError { code: ExpectedIdentifier, position: Position { line: 321, col: 7 } }
```

Adding a valid format (`model_descriptions`) vs. the currently error-prone format (`model_instances`) here for clarity:
```
model_descriptions: {
    86: (
      name: "OfficeChairBlack",
      source: (Search("OpenRobotics/OfficeChairBlack")),
      is_static: (true),
      optional_properties: ([]),
    ),
  },
  model_instances: {
    87: {
      "parent": 1,
      "name": "OfficeChairBlack",
      "pose": (
        trans: (6.562113, -6.721654, 0.0),
        rot: yaw(deg(243.52425)),
      ),
      "description": Some(86),
      "optional_properties": ([]),
    },
```

mxgrey:
Whoops, I hadn't thought of running all the unit tests before opening this PR. Thanks for helping me understand why the RON roundtrip test was failing.

I fell into a bit of a rabbit hole with https://github.com/ron-rs/ron/issues/548 and then https://github.com/ron-rs/ron/issues/555 but it seems everything is working now. For this to work we'll need to target the github repo of RON until its next release, but I think RON will probably have a new release well before we do.

xiyuoh (approved):
Thanks for hunting down the solution, tried out with the demo map and it works great! ✨  The error remains if we add anything to `optional_model_properties`, i.e. `Tasks` and `DifferentialDrive`, which would also require `serde(transparent)` for loading to work. However these struct are pending some refactor in https://github.com/open-rmf/rmf_site/pull/263 so I'm happy to give an approval first.

mxgrey:
@xiyuoh Considering the merge conflicts that will come up due to the changes in optional model properties due to the need to add `#[serde(transparent)]`, I'll close this PR and you can just copy/paste the relevant parts from here over to #263.

PR #266, opened 2025-02-05
Include/Hide model instances in scenarios

xiyuoh:
This PR addresses one of the issues in https://github.com/open-rmf/rmf_site/issues/254 to include/hide model instances from scenarios rather than adding/removing them entirely.

Major changes:
- No more `added_instances`, `moved_instances` and `removed_instances`; each `Scenario` stores `instances` with data about whether their poses have been moved from any parent scenario, and whether they are hidden in the current scenario.
- A collapsible widget listing all model instances (categorized by the affiliated model description), through which users can interact with each model instance:
   - Include or hide the model instance in the current scenario
   - Reset the model instance pose if it has been moved from any parent scenario
   - Remove the model instance from the site entirely (alternatively the Del key works on selected model instances too)
   - Displays the model instance's pose in the current scenario
- Removed deletion filter for model instances, now that we've moved to a include/hide approach rather than add/remove, so there's no longer a need to filter out deletion events based on whether the current scenario is a root or child.


rmf_site_editor/src/widgets/view_model_instances.rs
```diff
@@ -0,0 +1,241 @@
...
+                                                &mut self.reset_pose,
+                                                &mut scenario.instances,
+                                            );
+                                        } else {
+                                            non_affiliated_instances.insert(instance_entity);
+                                        }
+                                    }
+                                });
+                        }
+                        CollapsingHeader::new("Non-affiliated instances")
```
mxgrey, line 131:
```suggestion
                        CollapsingHeader::new("Unaffiliated instances")
```

rmf_site_editor/src/widgets/view_model_instances.rs
```diff
@@ -0,0 +1,241 @@
...
+            });
+    }
+}
+
+impl<'w, 's> ViewModelInstances<'w, 's> {
+    pub fn show_widget(&mut self, ui: &mut Ui) {
+        if let Some(current_scenario_entity) = self.current_scenario.0 {
+            if let Ok((_, _, mut scenario)) = self.scenarios.get_mut(current_scenario_entity) {
+                let current_scenario_instances = scenario.instances.clone();
+                let mut non_affiliated_instances = HashSet::<Entity>::new();
```
mxgrey, line 93:
```suggestion
                let mut unaffiliated_instances = HashSet::<Entity>::new();
```

rmf_site_editor/src/widgets/view_model_instances.rs
```diff
@@ -0,0 +1,241 @@
...
+                                // TODO(@xiyuoh) true if model is selected
+                                .default_open(false)
+                                .show(ui, |ui| {
+                                    for (
+                                        instance_entity,
+                                        instance_name,
+                                        category,
+                                        affiliation,
+                                        instance_site_id,
+                                    ) in self.model_instances.iter_mut()
```
mxgrey, line 110: I'm noticing instability in the ordering of this when I select or deselect instances.

I suspect a more efficient and more stable way of doing this iteration would be to iterate over the [`Members`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/site/group.rs#L31) of the description group.

mxgrey (changes requested):
I really like this refactoring. Being able to simplify the deletion logic and reducing some of the clutter in the UI are both big wins.

I left some comments on tweaks that I think could improve these changes further.

xiyuoh:
Thanks for the review @mxgrey ! The suggestion to use `SelectorWidget` and `Members` greatly improved the user experience.

I've implemented changes based on your comments with the exception of `Instance::Hidden` - I've created a `HiddenInstance` struct along with the other items in the enum as I thought it'd be good to store the last known `Pose` of the model instance before it was hidden, so that it'd appear at the same position when it is re-included later on. Since each entity's `Pose` keeps changing due to scenario switching it wouldn't be appropriate to use `Recall` here.

Additionally, I've tweaked `handle_reset_pose` and renamed it `handle_instance_updates` to manage any types of events where a change to `Instance` is involved in a specific scenario, including resetting to parent pose.

Other than those I also found that there were bugs (from this PR) that led to model instances being deleted if a scenario is removed, fixed that in [1176ba1](https://github.com/open-rmf/rmf_site/pull/266/commits/1176ba19b9ae4647711ad551e840024d2934f0df).

xiyuoh:
Apologies for making changes after requesting for a review, I realized I missed out on a sub-item in the follow up list to track the number of scenarios each instance shows up in and adding a diagnostic for instances that are not included in any scenarios. Added in [662e4aa](https://github.com/open-rmf/rmf_site/pull/266/commits/662e4aa42a8572874833e526840fb197cbe05083).

rmf_site_format/src/scenario.rs
```diff
@@ -35,18 +35,14 @@ pub struct ScenarioMarker;
 #[cfg_attr(feature = "bevy", derive(Component))]
 pub struct Scenario<T: RefTrait> {
     pub parent_scenario: Affiliation<T>,
-    pub added_instances: Vec<(T, Pose)>,
-    pub removed_instances: Vec<T>,
-    pub moved_instances: Vec<(T, Pose)>,
+    pub instances: HashMap<T, ((Pose, bool), bool)>,
```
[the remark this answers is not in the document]
mxgrey: I see that in the current implementation we have `pose` in every variant of `Instance`. To me that smells like we have the wrong layout for this information.
1. Do we really need any `pose` information if the object is hidden?
2. It seems `Added` is being used even if the instance was actually added in a parent scenario, is that right?

I think what I really had in mind was something like this:

```rust
pub enum Instance {
    Added(AddedInstance),
    Inherited(InheritedInstance),
    Hidden(HiddenInstance),
}

/// The instance was added by this scenario
pub struct AddedInstance {
    /// This is the default pose for the object
    pose: Pose,
}

/// The instance is being inherited from a parent scenario
pub struct InheritedInstance {
    /// The pose is modified relative to the parent scenario
    modified_pose: Option<Pose>,
}

/// The instance doesn't exist in this scenario
pub struct Hidden {
    // Do we actually need anything here?
}
```

So for example the [reset button](https://github.com/open-rmf/rmf_site/blob/213209cdff16caf63bd938f520356dce6a0aa56f/rmf_site_editor/src/widgets/view_model_instances.rs#L232) would be enabled if you have both `Instance::Inherited` and `inherited.modified_pose.is_some()`. When switching between scenarios, when we have `Instance::Inherited` we should crawl towards the root of the scenario tree until we get a `Some(Pose)` or until we reach `AddedInstance` to figure out where to place the instance. If we get past the root without encountering either then the instance would be implicitly hidden.

With the current system it's not actually behaving the way I would expect. Creating a new child scenario will essentially clone the state of the parent scenario at the time that you created the child scenario, but then any updates to the parent scenario will not carry over to the child scenario. Here are some steps to reproduce the unexpected behavior:

1. Launch the demo map
2. Set `Add Scenario:` to `Child`
3. Click `+` twice to create a lineage of three scenarios (the root scenario, its child, and its grandchild).
4. Select the child scenario (middle scenario)
5. Move an item in the child scenario
6. Select the grandchild scenario (deepest scenario)

You'll see that the item that was moved in the child scenario has reverted to its original position from the root scenario. The intended behavior is for the grandchild, which has not modified the position of the item, to have the same pose for the item as its direct parent (the direct child scenario of the root).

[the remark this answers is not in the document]
xiyuoh: Thanks for explaining!

> Do we really need any pose information if the object is hidden?

I opted to store the `Hidden` poses due to a possible case like this:
- A model instance A is added at pose `(x0, y0)` in the default root scenario
- A child scenario is created from the default root scenario, and model instance A is moved to `(x1, y1)`
- The user is toggling the include/hide button to experiment with the scenario. When model instance A is changed from `Inherited` to `Hidden`, the last known pose `(x1, y1)` of A in the child scenario would be lost. We would not be able to know the pose of A when toggling it back to be included in the child scenario, and we'd be forced to reset it to the unmodified parent pose `(x0, y0)`. Implementing `Recall` may not work as `Pose` would be changing with every scenario switch.

The same applies for toggling between include and hide in root scenarios - once the instance is `Hidden`, we'd lose any information about it's `Pose`.

> It seems Added is being used even if the instance was actually added in a parent scenario, is that right?

That's the current implementation. Thanks for clarifying and suggesting a less cluttered way to store and access parent scenario poses. I'll work on refactoring that.

> You'll see that the item that was moved in the child scenario has reverted to its original position from the root scenario. The intended behavior is for the grandchild, which has not modified the position of the item, to have the same pose for the item as its direct parent (the direct child scenario of the root).

I see now that I was approaching instance behavior very differently. I assumed we'd want the grandchild's pose to remain untouched if the child pose has changed, and the reset button would allow users to move the grandchild to the child pose if they'd like. Happy to change this so that inherited and unmodified pose changes are reflected automatically.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
+            update.scenario,
+            &children,
+            &recall_instance,
+            &scenarios,
+            &scenario_instances,
+        );
+        match update.update_type {
+            UpdateInstanceType::Add(new_pose) => {
+                commands
+                    .spawn(Instance::new_added(new_pose.clone()))
```
mxgrey, line 255: This feels suspicious to me. We're spawning a new instance modifier entity without checking if one already exists under this scenario for this instance. If a user repeatedly hid and added an instance to the scenario over and over again, I believe we'd be adding a new modifier entity each time.

I'm pretty sure the safer thing to do here is check whether an affiliated modifier entity already exists for this scenario before spawning a new one.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
+                            queue.push(*e);
+                        });
+                    }
+                }
+                // Only insert new instance in children scenarios. Other parent/root
+                // scenarios will not have access to this instance
+                for dependent in subtree_dependents.drain() {
+                    if let Ok((child_entity, _)) = scenarios.get(dependent) {
+                        commands
+                            .spawn(Instance::new_inherited(None))
```
mxgrey, line 274: Similarly I think we should check for an existing modifier in this scenario before spawning a new one.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
+                    .iter()
+                    .find(|(_, i)| *i == update.instance)
+                    .and_then(|(c_entity, _)| {
+                        scenario_instances
+                            .get_mut(*c_entity)
+                            .ok()
+                            .zip(Some(c_entity))
+                    })
+                else {
+                    continue;
```
mxgrey, line 293: It looks like this will short circuit if the instance modifier doesn't already exist for this scenario. That doesn't seem robust to me. I would think we should insert it if it doesn't already exist. I realize that's tricky since insertion commands don't mesh well with `&mut T`.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
-                    return;
+                        if parent_pose.is_some() {
+                            *instance = Instance::new_inherited(instance_pose)
+                        } else if let Some(instance_pose) = instance_pose {
+                            *instance = Instance::new_added(instance_pose);
+                        } else {
+                            let instance_id = model_instances
+                                .get(update.instance)
+                                .map(|n| n.0.clone())
+                                .unwrap_or(format!("{}", update.instance.index()).to_string());
```
mxgrey, line 312: You should use `.unwrap_or_else(|_| ...)` here, otherwise you're telling the program to format this string every single time.

Also I don't think `.to_string()` is needed since `format!` always outputs a String.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
+                            }
+                            _ => {}
+                        }
+                    }
+                    UpdateInstanceType::ResetPose => {
+                        if parent_pose.is_some() {
+                            *instance = Instance::new_inherited(None)
+                        }
+                    }
+                    _ => {}
```
mxgrey, line 339: In general I suspect it would be better to merge this whole match block with the `UpdateInstanceType::Added` match block that lives outside of this. I think splitting them based on whether you expect to spawn an entity is not sound since you can't **_really_** guarantee whether you really want to spawn an entity or modify a component in any of these cases.

One option that could simplify things is to insert a system which always runs before this one which adds an entity with `Instance::Inherited` and `Affiliated` components every time a new entity is created with a `Scenario` or `InstanceMarker` component. Essentially it would make sure the matrix of modifier entities is always filled in before this system is run.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -102,65 +135,214 @@ pub fn update_current_scenario(
...
                 }
             }
         }
+        if current_scenario.0.is_some_and(|e| e == update.scenario) {
+            change_current_scenario.send(ChangeCurrentScenario(update.scenario));
```
mxgrey, line 344: It's not totally clear to me if this is necessary. I could imagine a plugin that wants to make modifications to scenarios without changing what the current scenario is.

mxgrey (changes requested):
I think this is super close now, but there are a few more implementation details that I think we can make more robust.

Overall I think this new structure is definitely the right way to go, so just a few more tweaks and we should be good to merge.

rmf_site_editor/src/site/scenario.rs
xiyuoh: The `ChangeCurrentScenario` event reader updates the `CurrentScenario` resource to the contained scenario entity [here](https://github.com/open-rmf/rmf_site/blob/xiyu/include_instances/rmf_site_editor/src/site/scenario.rs#L121). This filter prevents unwanted scenario switching.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -16,82 +16,115 @@
...
     mut change_current_scenario: EventReader<ChangeCurrentScenario>,
     mut current_scenario: ResMut<CurrentScenario>,
-    current_workspace: Res<CurrentWorkspace>,
-    scenarios: Query<&Scenario<Entity>>,
-    mut instances: Query<(Entity, &mut Pose, &mut Visibility), With<InstanceMarker>>,
+    mut instances: Query<(Entity, &NameInSite, &mut Pose, &mut Visibility), With<InstanceMarker>>,
+    children: Query<&Children>,
+    recall_instance: Query<&RecallInstance>,
+    scenarios: Query<(Entity, &mut Scenario<Entity>)>,
+    scenario_entities: Query<(&mut Instance, &Affiliation<Entity>)>,
```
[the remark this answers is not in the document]
xiyuoh: Thanks for the suggestion. This definitely helps to simplify a lot of complicated variable naming throughout the PR.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -16,82 +16,115 @@
...
 
-        // Deselect if not in current scenario
-        if let Some(selected_entity) = selection.0.clone() {
-            if let Ok((instance_entity, ..)) = instances.get(selected_entity) {
-                if active_instances.get(&instance_entity).is_none() {
-                    selection.0 = None;
-                }
+            if deselect {
+                selection.0 = None;
+                selected.is_selected = false;
```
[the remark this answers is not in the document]
xiyuoh: I just remembered why I opted to do this: it was to prevent users from selecting a model instance if it was hidden in the current scenario. It didn't seem intuitive to me, but it's also not a deal-breaker. Added the `Select` event writer in [36d0ef8](https://github.com/open-rmf/rmf_site/pull/266/commits/36d0ef8d6ec9cfa938d329acf5ac7517de8c27e2)

rmf_site_editor/src/site/scenario.rs
xiyuoh: The intention is to trigger the `UpdateInstance::Add` event only when a new model instance has been spawned and placed in the site, so it shouldn't happen when existing model instances are being toggled or modified.

I've since moved this into the new system `insert_new_instance_modifiers` with the suggested checks in [a1e5719](https://github.com/open-rmf/rmf_site/pull/266/commits/a1e5719dc3dc2b533ae8ae0f036cb3f119ff4070) using `Added<T>` instead of `is_added()` as well.

rmf_site_format/src/scenario.rs
```diff
@@ -35,18 +108,14 @@ pub struct ScenarioMarker;
 #[cfg_attr(feature = "bevy", derive(Component))]
 pub struct Scenario<T: RefTrait> {
     pub parent_scenario: Affiliation<T>,
-    pub added_instances: Vec<(T, Pose)>,
-    pub removed_instances: Vec<T>,
-    pub moved_instances: Vec<(T, Pose)>,
+    pub instances: BTreeMap<T, Instance>,
```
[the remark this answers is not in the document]
xiyuoh: This is a great suggestion, I see how it improves the way we access scenario properties too. Updated in [77d4c72](https://github.com/open-rmf/rmf_site/pull/266/commits/77d4c72f8a6a70859be476c959e780536a7c84ed).

rmf_site_editor/src/site/scenario.rs
mxgrey: Ohhh I see now. Then I guess the purpose of the `ChangeCurrentScenario` event is to force an update of the visuals rather than actually changing the current scenario?

xiyuoh: That's right, it can be interpreted as changing elements in the current scenario

rmf_site_editor/src/site/scenario.rs
```diff
@@ -18,148 +18,381 @@
...
+            // If an instance modifier entity already exists for this scenario, update it
+            let instance_modifier = instance_modifier.as_mut();
+            match instance_modifier {
+                InstanceModifier::Added(_) => {
+                    *instance_modifier = InstanceModifier::new_added(new_pose.clone())
+                }
+                InstanceModifier::Inherited(_) => {
+                    *instance_modifier = InstanceModifier::new_inherited(Some(new_pose.clone()))
                 }
+                _ => {}
```
mxgrey, line 272: Let's explicitly use `InstanceModifier::Hidden => {}` here just in case we ever add another variant to this enum in the future.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -18,148 +18,381 @@
...
-                    existing_added_instance.1 = pose.clone();
-                    return;
-                } else if pose.is_added() {
-                    current_scenario
-                        .added_instances
-                        .push((entity, pose.clone()));
-                    return;
+        // Insert this new instance modifier into children scenarios as Inherited
+        let mut subtree_dependents = HashSet::<Entity>::new();
+        let mut queue = vec![current_scenario_entity];
```
mxgrey, line 284: We need to add modifiers to all scenarios, not just children of the current scenario.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -18,148 +18,381 @@
...
+    children: Query<&Children>,
+    current_scenario: Res<CurrentScenario>,
+    scenarios: Query<Entity, With<ScenarioMarker>>,
+) {
+    let Some(current_scenario_entity) = current_scenario.0 else {
+        return;
+    };
+    // Insert instance modifier entities when new scenarios are created
+    for (scenario_entity, parent_scenario) in added_scenarios.iter() {
+        if let Some(parent_scenario_entity) = parent_scenario.0 {
```
mxgrey, line 226: When a new scenario is added we should give it a modifier for every instance in the site, not just for every instance that its parent has a modifier for.

With the current implementation, when you add a new root scenario, the entire list of instances in the `view_model_instances` panel is blank.

rmf_site_format/src/scenario.rs
```diff
@@ -19,13 +19,78 @@ use crate::*;
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "bevy", derive(Component))]
+pub enum InstanceModifier {
+    Added(AddedInstance),
+    Inherited(InheritedInstance),
+    Hidden,
+}
+
+impl InstanceModifier {
+    pub fn new_added(pose: Pose) -> Self {
```
mxgrey, line 38: Let's change this to be more idiomatic:

```rust
pub fn added(pose: Pose) -> Self {
```

rmf_site_format/src/scenario.rs
```diff
@@ -19,13 +19,78 @@ use crate::*;
...
+        Self::Added(AddedInstance { pose: pose })
+    }
+
+    pub fn new_inherited(pose: Option<Pose>) -> Self {
```
mxgrey, line 42: Let's do

```rust
pub fn inherited() -> Self {
```

I don't think we're likely to see scenarios where we're creating a new `Inherited` from scratch and immediately have a modified pose for it.

rmf_site_format/src/scenario.rs
```diff
@@ -19,13 +19,78 @@ use crate::*;
...
+        Self::Inherited(InheritedInstance {
+            modified_pose: pose,
+        })
+    }
+
+    pub fn new_hidden() -> Self {
```
mxgrey, line 48: We could change this to `pub fn hidden()` to match the others, but I think we should just call `InstanceModifier::Hidden` directly.

mxgrey (changes requested):
Thanks for all the heavy iteration. We're definitely getting much closer, but there are still some parts that aren't working quite right and which I think could actually be simplified. I've left in-line comments about specific items, but I'll also describe how I think it should work at a high level:

The current implementation is manually shuffling around instance modifiers quite a bit. It leads to some incorrect/inconsistent behavior, and I think it even makes the code more complex than it needs to be. Here's how I would recommend we structure the modifier spawning:

* When a new scenario is added, spawn exactly one modifier for each instance in the site. If the scenario is a root scenario, make it a `Hidden` modifier. If it's a child scenario, use the `Inherited` modifier. If the instance is in the list of newly added instances and we're adding the current scenario, use `Added`.
* When a new instance is added, spawn exactly one modifier into each scenario in the site. If the scenario is the current scenario, make it an `Added` modifier. If it's a root scenario, `Hidden`. If it's a child scenario besides the current scenario, `Inherited`. If it's a scenario that was just added in the previous bullet point, skip it.

This will ensure that we have a fully filled out matrix of Scenarios x Instances. Every scenario has exactly one modifier for every instance, and every instance has exactly one modifier in every scenario.

I'm also thinking more and more that we should not use bevy's built-in parent/child mechanism to manage scenario parentage. Instead all scenarios are children of the site (from Bevy's perspective) and we just use `Affiliation` to keep track of what each scenario's parent is. If we can land on an implementation where there's no need for parent scenarios to know who their child scenarios are, then I think we can make that work. As far as I can figure if we implement this according to the recommendations I've made then I don't think parents ever need to know about their children, so this might be worth exploring.

rmf_site_editor/src/site/scenario.rs
```diff
@@ -18,148 +18,381 @@
...
+            // Inherit any instance modifiers from the parent scenario
+            if let Ok(children) = children.get(parent_scenario_entity) {
+                children.iter().for_each(|e| {
```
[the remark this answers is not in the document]
xiyuoh: Interesting, I assumed we'd want to copy `Hidden` into the new child (unless a user deliberately includes it). Reason being when creating children scenarios it's likely we'd want to work off from the current state of the parent scenario.

For example:
- Root Scenario A0 has 100 model instances to represent site obstacles (`1` to `20`) and furniture (`21` to `100`)
- Scenario B1, child of A0, includes furniture models + only obstacles represented by model instances `1` to `10`
- Scenario C1, child of A0, includes furniture models + only obstacles represented by model instances `11` to `20`
- If I want a new grandchild scenario B2 to include instances `1` to `10` and modify their poses, I would create a child off B1

If all instances in a new scenario are `Inherited`, and therefore included, for every new child scenario created from B1, I would have to manually find instances `11` to `20` and hide them one by one.

Currently when an instance modifier is categorized as `Inherited`, it means that the affiliated model instance is included in the scenario, with possibly modified pose. It does not inherit the visibility of the same model instance in the parent scenario. This is because we wanted the option of setting an instance in the child scenario to be visible even when it is hidden in the parent scenario.

If we'd like a new child scenario to inherited all modified poses and visibility, we can add a `modified_visibility` field to `InheritedInstance`. This way, with reference to the steps that you've listed above, if we toggle the root scenario's object back to visible, the object should become visible in the child scenario also, unless explicitly set otherwise.

rmf_site_format/src/scenario.rs
xiyuoh: We'd need to create new `Inherited` with modified poses when toggling instance modifiers from `Hidden` to `Inherited` in cases where the model instance has been moved and toggled off.

xiyuoh: I ended up following your suggestion to take out the arguments, and in cases where we need to instantiate with values I created `InheritedInstance { .. }` directly instead to accommodate future modifications.

rmf_site_editor/src/site/scenario.rs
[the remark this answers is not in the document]
xiyuoh: After some experimenting, I added an `explicit_inclusion` field in `InheritedInstance` to indicate when a modifier should be included in the scenario even when the instance is hidden in the parent scenario.

I believe it should be closer to what we want now; with the 3 modes suggested, users can choose whether to explicitly include/show an inherited modifier, or simply follow the parent scenario (unless root). Thanks for clarifying the desired behavior, these are updated in [a7102e7](https://github.com/open-rmf/rmf_site/pull/266/commits/a7102e7a9d44f2a284484937f2a775e9bf724b0e).

rmf_site_editor/src/site/scenario.rs
```diff
@@ -185,6 +209,34 @@ fn retrieve_parent_pose(
...
+) -> Option<bool> {
+    let mut parent_visibility: Option<bool> = None;
+    let mut entity = scenario_entity;
+    while parent_visibility.is_none() {
+        let Some(parent_entity) = scenarios.get(entity).ok().and_then(|(_, a)| a.0) else {
+            break;
+        };
+
+        let instance_modifier_entities =
+            get_instance_modifier_entities(parent_entity, children, instance_modifiers);
```
mxgrey, line 227: It seems like the current implementation uses this function to rebuild a hashmap N x M times where N is the number of scenarios and M is the number of instances with modifiers per scenario. It also seems like each time we build this hashmap we only do a single `.get(&Entity)` out of it.

In terms of performance the ideal thing to do here would be to cache this hashmap in a component for each scenario, but at that point we need to worry about cache invalidation, which means reinventing relations in Bevy, so we're better off waiting until we can use the relations feature.

The next best thing if we don't cache the hashmap would be to change this function so that you pass in `instance_entity` as an argument and the function just returns as soon as it finds the modifier for that instance rather than building the whole hashmap. You could call it

```rust
fn find_modifier_for_instance(
    instance: Entity,
    scenario: Entity,
    children: &Query<&Children>,
    instance_modifiers: &Query<(&mut InstanceModifier, &Affiliation<Entity>)>,
) -> Option<Entity> {
    ...
}
```

We do have some situations where we get multiple uses out of the hashmap, so we should make two separate functions: One to build the hashmap (i.e. keep the current function as-is) and one to return as soon as it finds the right modifier (i.e. introduce `find_modifier_for_instance`).

mxgrey (changes requested):
Awesome!!

I just have one last recommendation for a tweak to improve scalability, and then I think we'll be good to merge!

xiyuoh:
Thanks @mxgrey for the review, the suggestions have greatly improved the overall behavior of this feature. I've [a0fd749](https://github.com/open-rmf/rmf_site/pull/266/commits/a0fd749520fecf9b4d1574a7f2a5d66b5bf221df) the separate system for finding modifiers for a specific instance.

Additionally I found that when I add an instance in a child scenario (as `Added` modifier), the visibility toggling button doesn't work well. This is fixed in [2172ad8](https://github.com/open-rmf/rmf_site/pull/266/commits/2172ad8bae5888e1674b235930a3887490972fd8) by storing whether this was an `Added` or `Inherited` modifier in RecallInstance. Now, when an `Added` is hidden and toggled to be re-included, its modifier type won't be changed to `Inherited` just because a parent scenario exists.

mxgrey (approved):
I like the `explicit_inclusion` concept.

I have a feeling we're going to end up iterating on this one more time where we remove `Hidden` from `InstanceModifier` and instead just have a `visibility` field inside of `InstanceModifier::Inherited`. But for now this implementation is working well, and if we decide to make more changes we can do that in a follow up PR.

PR #267, opened 2025-02-10
Moved site object creation buttons to top menu bar

### LINKED ISSUE #228: Move creation widgets into the menu bar
I think it's more conventional to have creation buttons (e.g. create wall, create floor, add drawing, etc) as icon buttons in the menu bar at the top rather than text buttons on the side panel.

The biggest challenge for this feature request is creating the art for the icons.

sauk2:
### Implemented feature

Closes #228

- This PR moves the site object creation buttons into the top menu bar from the side panel. A separator separates the buttons from the rest of the menu items.
- The buttons appear according to `AppState::SiteEditor` and `AppState::SiteDrawingEditor`.
- The `Site Object` creation option in the side panel has been removed. The default selection now is the `Drawing` option.


mxgrey (changes requested):
Thanks for getting this improvement started! It's very interesting to see how the menu buttons and icon buttons can co-exist.

I have a few thoughts on how we can improve it from here:

* For the buttons I think a more modern style would be to just display the icon and then let the text show up as a tooltip when the user hovers over it. You can use `Button::on_hover_text("...")` to put that in.
* The current approach hard-codes the set of regular buttons that can be added to the menu bar. It would be better if we provide a way to
  1. Opt in/out of the creation widget buttons
  1. Allow users to add their own custom widget buttons to the menu bar

To make menu bar widgets extensible we should be able to draw inspiration from the [`PropertiesPanel`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/properties_panel.rs#L154) which uses [`show_panel_of_tiles`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/panel_of_tiles.rs#L38). I think the steps to make this work would look something like:
* Define a new panel widget called something like `HeaderPanel`.
* `HeaderPanel` would be implemented the same way as the [`PropertiesPanel` module](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/properties_panel.rs).
  * We would introduce `HeaderTilePlugin<W>` that works the same way as [`PropertiesTilePlugin<W>`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/properties_panel.rs#L87).
* [`MenuBarPlugin`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/menu_bar.rs#L28) would create a `HeaderTilePlugin` instead of [a `PanelWidget`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/menu_bar.rs#L30).
* [`top_menu_bar`](https://github.com/open-rmf/rmf_site/blob/a9ac2a4ad03903cf1b96730238b9bfddd430077e/rmf_site_editor/src/widgets/menu_bar.rs#L30) would change from a `PanelWidget` to a `WidgetSystem<Tile>`
* The creation buttons would move out of `top_menu_bar` and go into its own `WidgetSystem<Tile>` that can be put inside of a `HeaderTilePlugin<W>`.

With that, users would be able to pick and choose which plugins exist in the `HeaderPanel` and would be able to add their own custom buttons to the header.

Let me know if you have any questions or concerns about this outline.

rmf_site_editor/src/widgets/creation_site_object.rs
```diff
@@ -0,0 +1,254 @@
...
+            if button_clicked(ui, "📏", "Measurement") {
+                params.anchor_selection.create_measurements();
+            }
+        }
+    }
+}
+
+/// Helper funtion to display the button name on hover
+fn button_clicked(ui: &mut Ui, icon: &str, tooltip: &str) -> bool {
+    ui.add(Button::new(icon)).on_hover_text(tooltip).clicked()
```
mxgrey, line 253: I like this wrapper function 👍

rmf_site_editor/src/widgets/menu_bar.rs
```diff
@@ -27,13 +32,87 @@ pub struct MenuBarPlugin {}
...
+            HeaderPanelPlugin::default(),
+            MenuDropdownPlugin::default(),
+            LaneCreationPlugin::default(),
+            LocationCreationPlugin::default(),
+            WallCreationPlugin::default(),
+            DoorCreationPlugin::default(),
+            LiftCreationPlugin::default(),
+            FloorCreationPlugin::default(),
+            FiducialCreationPlugin::default(),
+            MeasurementCreationPlugin::default(),
```
mxgrey, line 45: Having each creation widget as its own plugin is more granularity than I was expecting, but I don't have any problem with it. It's plausible that someone would want to customize their app to only allow certain editing operations, e.g. users can create lanes but can't change the environment.

Still I would recommend making a plugin to bundle all these, e.g.

```rust
#[derive(Default)]
pub struct CreationPlugin {}

impl Plugin for CreationPlugin {
    fn build(&self, app: &mut App) {
        app.add_plugins((
            LaneCreationPlugin::default(),
            LocationCreationPlugin::default(),
            ...
        ));
    }
}
```

Then we can replace all of these with just `CreationPlugin::default()`.

mxgrey (changes requested):
Thanks for iterating on this @sauk2, I really like how this is turning out.

I left some recommendations for a few more tweaks in the inline comments. After those are addressed I think this will be ready to merge!

rmf_site_editor/src/widgets/menu_bar.rs
sauk2: > Having each creation widget as its own plugin is more granularity than I was expecting, but I don't have any problem with it. It's plausible that someone would want to customize their app to only allow certain editing operations, e.g. users can create lanes but can't change the environment.

That's what I thought too!

I have bundled all the plugins into on standard set.

mxgrey:
I like how this has turned out, but I became worried about fragmenting the creation widgets between the menu bar and the properties panel.

Thanks to how generalized this implementation is, I found that it wasn't too bad to finish migrating all the creation widgets to the menu bar, so I've [opened a PR](https://github.com/sauk2/rmf_site/pull/1) against this one to wrap that up. Take a look and let me know what you think.

I've removed the need to have a `creation_site_object.rs` module separate from `creation.rs`, so everything is in `creation.rs` now. Note that when reviewing the code of that PR it will be easier to follow by doing a diff of the old `creation_site_object.rs` and the new `creation.rs`.

mxgrey (approved):
Thanks for driving this quality of life improvement @sauk2 . It's remarkable how much better it feels to add new objects to the scene.

PR #268, opened 2025-02-17
Implement RecallPlugin for RobotPropertyKind

xiyuoh:
This PR implements Recall for RobotPropertyKinds such that if a Recall value is available in the world and `update_robot_property_kind_components` receives a default value of the RobotPropertyKind, the recalled value will be used to update property kind. Otherwise, we will insert the default value. The `RecallPropertyKind` trait is also introduced to enable access to the recalled value for each property kind via `assume()`.

This fixes the current behavior where the value resets to default if the property kind is re-selected or switched.


mxgrey (approved):
Excellent, this is working perfectly :+1:

PR #270, opened 2025-02-20
Scenario tasks

xiyuoh:
This PR targets one of the items in https://github.com/open-rmf/rmf_site/issues/254 and introduces `Task` to the site editor.
- Users may add `Task` in the editor and assign fleet/robots to them.
- The same pool of created `Task` exist across all scenarios, however users have the option to modify certain `TaskParams` in each scenario.
- Similar to the behavior of inherited properties in Model Instances (for `Visibility` and `Pose`), `TaskParams` and task inclusion can also be inherited, modified or reset back to match the properties from any parent scenarios. These are managed by the use of `TaskModifier` across scenarios and tasks.

This PR also includes the removal of the `agent` module.

Follow up to this PR includes a scenario-specific task ordering system based on the start time it should be dispatched/assigned.


mxgrey (approved):
These are some great improvements to the task system that we had before!

I've opened a ticket for a few quality of life items that we can follow up on: https://github.com/open-rmf/rmf_site/issues/312

But I there's a lot of good heavy lifting done in this PR, so let's get it in and then keep iterating.

Issue #255, opened 2024-12-26
Remove all uses of .expect(-) and .unwrap()


mxgrey:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

The site editor does not currently do any automatic saving or backups. Because of this it could be devastating if the application ever crashes on a user. An arbitrary amount of progress could be lost if they haven't saved in some amount of time. Because of this we should make great effort to minimize the risk of the application crashing.

Anywhere that we currently use `.expect(-)` or `.unwrap()` we should instead use `error!(...)` and either `return` or `continue` depending on the context.

mxgrey:
Note that we can allow use of `.unwrap()` or `.expect(...)` in certain code paths, like anything inside of a `#[test]` function or inside of the [`from_world`](https://github.com/open-rmf/rmf_site/blob/809f44c92502e26214fd8c5da41db0937ac9d9b3/rmf_site_editor/src/site/assets.rs#L110) implementation of a `Resource` that only gets created while the application is starting up.

Tests can have `.unwrap()` and `.expect(...)` because we want them to exit with an error code as soon as anything undesirable occurs.

Resources can have `.unwrap()` in their `from_world` implementations because the failure would happen immediately during startup which allows us to identify the problem quickly, and there is no risk of a user losing progress on their work.

PR #273, opened 2025-03-31
replaced a few expects

Nielsencu:
## Bug fix
Replaced non-critical .expect()s, for instance, the ones that only prints out a failure if fails. I'm still not sure what's the best case to handle the expect()s that is inside a function returning an expected type.

An example:
https://github.com/open-rmf/rmf_site/blob/809f44c92502e26214fd8c5da41db0937ac9d9b3/rmf_site_editor/src/interaction/cursor.rs#L149C1-L153C11

Here, the trait FromWorld's implementation of from_world always expects the function to return the Self object, meaning that it cannot fail. However, in the case that there are errors in creating interaction or site assets, I think it should return an Err object, and handled properly by callers of the from_world function. The other option is to keep the return Self type, but provide default value in case of such errors.

### Fixed bug
Part of bugfix for https://github.com/open-rmf/rmf_site/issues/255
### Fix applied
Replaced a few of the expect()s with error! macros.

### DIFF
--- a/rmf_site_format/src/anchor.rs
+++ b/rmf_site_format/src/anchor.rs
@@ -16,6 +16,7 @@
 */
 
 use crate::{Categorized, Category, Pose};
+use bevy::utils::tracing::error;
 #[cfg(feature = "bevy")]
 use bevy::{
     ecs::{query::QueryEntityError, system::SystemParam},
@@ -23,7 +24,6 @@ use bevy::{
 };
 use glam::{Vec2, Vec3};
 use serde::{Deserialize, Serialize};
-use bevy::utils::tracing::error;
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 // TODO(MXG): Change this to untagged for a cleaner looking format once this
@@ -42,7 +42,9 @@ impl From<[f32; 2]> for Anchor {
     }
 }
 
-fn to_slice(p: &[f32; 3]) -> [f32; 2] { [p[0], p[1]] }
+fn to_slice(p: &[f32; 3]) -> [f32; 2] {
+    [p[0], p[1]]
+}
 
 impl Anchor {
     pub fn translation_for_category(&self, category: Category) -> [f32; 2] {
--- a/rmf_site_format/src/anchor.rs
+++ b/rmf_site_format/src/anchor.rs
@@ -16,7 +16,6 @@
 */
 
 use crate::{Categorized, Category, Pose};
-use bevy::utils::tracing::error;
 #[cfg(feature = "bevy")]
 use bevy::{
     ecs::{query::QueryEntityError, system::SystemParam},
@@ -42,7 +41,7 @@ impl From<[f32; 2]> for Anchor {
     }
 }
 
-fn to_slice(p: &[f32; 3]) -> [f32; 2] {
+fn to_array2(p: &[f32; 3]) -> [f32; 2] {
     [p[0], p[1]]
 }
 
@@ -51,7 +50,7 @@ impl Anchor {
         match self {
             Self::Translate2D(v) => *v,
             Self::CategorizedTranslate2D(v) => *v.for_category(category),
-            Self::Pose3D(p) => to_slice(&p.trans),
+            Self::Pose3D(p) => to_array2(&p.trans),
         }
     }
 
@@ -74,7 +73,7 @@ impl Anchor {
                         return true;
                     }
                     Self::Pose3D(p) => {
-                        let p_right = Vec2::from_array(to_slice(&p.trans));
+                        let p_right = Vec2::from_array(to_array2(&p.trans));
                         return (p_left - p_right).length() <= dist;
                     }
                 }
@@ -97,7 +96,7 @@ impl Anchor {
                 }
                 Self::Pose3D(p) => {
                     let p_left = Vec2::from_array(*left_categories.for_general());
-                    let p_right = Vec2::from_array(to_slice(&p.trans));
+                    let p_right = Vec2::from_array(to_array2(&p.trans));
                     return (p_left - p_right).length() <= dist;
                 }
             },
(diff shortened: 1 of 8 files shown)


rmf_site_format/src/anchor.rs
```diff
@@ -42,7 +43,14 @@ impl From<[f32; 2]> for Anchor {
 }
 
 fn to_slice(p: &[f32]) -> &[f32; 2] {
-    p.try_into().expect("Wrong array size")
+    match p.try_into() {
```
mxgrey, line 46: I think this is a case where we're taking a fallible code path when we don't really need to.

It seems the entire purpose of `to_slice` is to rework the translation inside of an anchor's pose into a `&[f32; 2]` so that the `[x, y]` position of an anchor can be _borrowed_ no matter which variant is being used for the anchor.

Realistically a `&[f32; 2]` has almost no benefit over a `[f32; 2]` (value instead of borrow), and we can always produce a `[f32; 2]` value without any possibility of an `Err`.

I think the right fix for this is to have `translation_for_category` return a `[f32; 2]` and eliminate the use of `to_slice` and/or replace it with something that produces a `[f32; 2]` for a given `&[f32; 3]`.

rmf_site_format/src/sdf.rs
```diff
@@ -23,12 +23,19 @@ use once_cell::sync::Lazy;
...
 use thiserror::Error;
+use bevy::{prelude::*, utils::tracing};
+use tracing::error;
 
 const DEFAULT_CABIN_MASS: f64 = 1200.0;
 
 static WORLD_TEMPLATE: Lazy<SdfRoot> = Lazy::new(|| {
-    yaserde::de::from_str(include_str!("templates/gz_world.sdf"))
-        .expect("Failed deserializing template")
+    match yaserde::de::from_str(include_str!("templates/gz_world.sdf")) {
```
mxgrey, line 32: Since this can have cascading ramifications if it fails, I think the right thing to do here is

```rust
static WORLD_TEMPLATE: Lazy<Result<SdfRoot, String>> = ...
```

Then in `Site::to_sdf` where this gets used, we return an `Err` result if `WORLD_TEMPLATE` has an `Err` value.

We can add a variant to `SdfConversionError` like `CorruptedWorldTemplate` to cover this error case.

rmf_site_format/src/sdf.rs
```diff
@@ -870,6 +877,8 @@ mod tests {
             ..Default::default()
         };
         let s = yaserde::ser::to_string_with_config(&sdf, &config).unwrap();
-        std::fs::write("test.sdf", s).unwrap();
+        if let Err(e) = std::fs::write("test.sdf", s) {
+            error!("Unable to write test.sdf {:?}", e);
```
mxgrey, line 881: This is inside of a test so we actually do want to use `.unwrap()` here. We want the tests to exit with an error code right away whenever something violates our expectations.

mxgrey (changes requested):
Thanks for the PR! This will be a nice step forward in reliability.

There are just a few places where I think your changes can be improved. I've left inline comments for each.

mxgrey:
> I built the repo on my comp and the app ran fine, but the github CI reported build failure.

If you're seeing `style / style (pull_request)` fail, that means you should run

```bash
cargo fmt
```

in the root directory of the repo and then commit and push the changes that produces.

rmf_site_format/src/sdf.rs
```diff
@@ -49,6 +50,8 @@ pub enum SdfConversionError {
     BrokenModelInstanceReference(u32),
     #[error("Entity [{0}] referenced a non existing model description")]
     BrokenModelDescriptionReference(u32),
+    #[error("Failed deserializing world template")]
+    CorruptedWorldTemplate,
```
mxgrey, line 53: Let's make this

```rust
CorruptedWorldTemplate(String)
```

and put the `Err(String)` from `WORLD_TEMPLATE` into it when an error occurs. You should also change the error message to something like

```rust
#[error("Failed to deserialize world template: {0}")]
```

so the details of the error will be included in the `SdfConversionError` message.

rmf_site_format/src/anchor.rs
```diff
@@ -41,16 +42,16 @@ impl From<[f32; 2]> for Anchor {
     }
 }
 
-fn to_slice(p: &[f32]) -> &[f32; 2] {
-    p.try_into().expect("Wrong array size")
+fn to_slice(p: &[f32; 3]) -> [f32; 2] {
```
mxgrey, line 45: It would be inaccurate to keep calling this `to_slice` since the term "slice" refers to something of the form `&[T]`, and we no longer have the `&`. A better function name might be `to_array2` or `truncate`.

mxgrey (changes requested):
We're very close now, just two small tweaks and then we can merge!

Nielsencu:
made the required changes!

rmf_site_format/src/sdf.rs
```diff
@@ -18,18 +18,18 @@
 use crate::{
     Anchor, Angle, AssetSource, Category, DoorType, Level, LiftCabin, Pose, Rotation, Site, Swing,
 };
+use bevy::{prelude::*, utils::tracing};
```
mxgrey, line 21: Why was `bevy` introduced here? Was that done to bring in `error!` from `tracing`?

I don't think anything from `bevy` is being used anymore, so we can remove this. It's currently causing compilation errors because we want `rmf_site_format` to be able to build without `bevy`.

mxgrey (changes requested):
It seems `cargo fmt` needs to be run again, and we need to remote an unnecessary case of `use bevy::...`.

rmf_site_format/src/sdf.rs
mxgrey: FYI to reproduce the build error locally you should be able to `cd` into the `rmf_site_format` directory and run `cargo test`.

Nielsencu:
Sorry for the build issues! Applied the changes, ran cargo fmt and ran cargo test successfully (thought running cargo run in root was sufficient)

PR #274, opened 2025-04-03
replaced unwraps

Nielsencu:
Bug fix
Replaced unwrap()s, where necessary.

Fixed bug
Part of bugfix for https://github.com/open-rmf/rmf_site/issues/255

Fix applied
Replaced with error! macros or silent.


mxgrey (approved):
Sorry for the long delay in reviewing this. We got very bogged down in the migration to bevy 0.16.

Thanks for the cleanup! I fixed the post-migration merge conflicts and tweaked a few things for simplicity, and now I think this is ready to go.

Issue #248, opened 2024-12-24
Replace all uses of FileDialog with the FileDialogServices


mxgrey:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

Tracking [this recommendation](https://github.com/open-rmf/rmf_site/pull/239#discussion_r1895164797) so we remember to follow up on it.

* We should replace the use of `rfd::FileDialog` for browsing local assets and exporting nav graphs with a workflow based on `FileDialogServices`.
* Exporting lights should be removed since it was put in for a one-off demo and is no longer relevant.

PR #275, opened 2025-04-06
replace with fds

Nielsencu:
Part of https://github.com/open-rmf/rmf_site/issues/248.

Attempt to replace fds on view_nav_graphs.rs file:

Initializes the File Dialog Services (fds) as a resource during build(), and define fds under systemparam as a ResMut (mutable resource). Then, create file dialog filters and pass it as request input to file dialog service, and take the response when promise is completed.

Question:
- As for file dialog filters, what should the name of the file requested be and the extension?
- Is the usage of ResMut correct here? I was thinking of using it since it's a unique borrow, meaning only one person can borrow it at a time

### DIFF
--- a/rmf_site_editor/src/widgets/mod.rs
+++ b/rmf_site_editor/src/widgets/mod.rs
@@ -187,11 +187,7 @@ impl Plugin for StandardUiPlugin {
             )
             .add_systems(
                 PostUpdate,
-                (
-                    resolve_light_export_file,
-                    resolve_nav_graph_import_export_files,
-                )
-                    .run_if(AppState::in_displaying_mode()),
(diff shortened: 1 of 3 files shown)


Nielsencu:
Removed export lights and added + sign next to add button
image

Import Nav Graph with FileDialogServices  (nav.yaml file extension)
image

mxgrey:
Thanks for working on this, but I've started down a [different path](https://github.com/open-rmf/rmf_site/commit/69e3380a6d19673aa0f54fb967af130fc4db64fd) for refactoring the nav graph import/export that uses the menu bar instead of the side panel widget.

I'll be closing this PR in favor of that other branch since it will also be introducing headless export of nav graphs.

Issue #225, opened 2024-06-11
[Feature request]: Warn before discarding unsaved changes


mxgrey:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

If a user is closing the application or performing any other action that would discard any changes which have not been saved, the editor should block that action and prompt the user to make sure they are really okay with losing the unsaved changes.

kscottz:
Heads up, I am going to open this one up to the community.

pedroazeredo04:
Hello! I am a new contributor and I am interested in helping with this

kscottz:
@pedroazeredo04 go ahead, give it a try. Please keep us posted.

kscottz:
@pedroazeredo04 how is it going?

pedroazeredo04:
Hi @kscottz !! I have added the progress until now to [my fork](https://github.com/pedroazeredo04/rmf_site/tree/exit_confirmation). I wrote a plugin that prompts the user with a popup screen, asking if they are sure they want to quit, when they try to close the site. What’s still missing is detecting when the site has been modified, so that the prompt only appears when there are unsaved changes.

However, I am not sure about the best way to keep track of unsaved changes. I thought about making a `SiteChanged: bool` resource, and edit it to `true` in every function that changes the state of the site. But I don't think that's a good idea, since that's hardly maintainable. Any suggestions?

mxgrey:
Thanks for the effort, Pedro! I'd suggest [opening a PR](https://github.com/open-rmf/rmf_site/compare) to discuss the next steps in more detail.

I think having a resource to track whether there are unsaved changes is a good idea, and you're right that it would be unreasonable to put that resource in every part of the code. Instead what I would recommend is toggling the resource to true just in the [`update_changed_values`](https://github.com/open-rmf/rmf_site/blob/c5f5a7a223cec8184e49580be9f762394861611d/rmf_site_editor/src/site/change_plugin.rs#L70) system since that catches most of the changes that happen during editing. It won't be perfect because it doesn't catch _every_ unsaved change, but I think it will get us very good coverage with very little effort.

[...]

PR #279, opened 2025-04-25
Exit confirmation

pedroazeredo04:
### Implemented feature

This Pull Request targets #225, by adding a prompt that asks the user if they really want to exit the app, whenever there are unsaved changes.

In order to do that, I created an `ExitConfirmationPlugin`, that is responsible for handling the `WindowCloseRequested` events, by knowing when the user should be stopped or not. This plugin is also responsible for creating the window with the exit confirmation.

In order to keep track of the unsaved changes, as suggest by @mxgrey, I created a `SiteChanged` resource, and toggled it to `true` in the [update_changed_values](https://github.com/open-rmf/rmf_site/blob/c5f5a7a223cec8184e49580be9f762394861611d/rmf_site_editor/src/site/change_plugin.rs#L70) function.

As this is my first pull request and I am quite new to Bevy and to this project, I'll be glad to change anything needed.

### DIFF
--- a/rmf_site_editor/src/exit_confirmation.rs
+++ b/rmf_site_editor/src/exit_confirmation.rs
@@ -1,38 +1,52 @@
+use crate::AppState;
 use bevy::app::AppExit;
 use bevy::prelude::*;
 use bevy::window::WindowCloseRequested;
 use bevy_egui::{egui, EguiContexts};
 
 #[derive(Resource, Default)]
-pub struct QuitDialog {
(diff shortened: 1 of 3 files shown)


mxgrey (approved):
Thanks for this excellent contribution, @pedroazeredo04 !

I've pushed a few commits to your branch to fix some small but important details. Feel free to ask me about them if any are unclear.

I think this feature is looking great now! I'll merge as soon as CI passes.

Issue #277, opened 2025-04-10
Bug: Sometimes models are stuck as hidden


mxgrey:
I've been seeing an intermittent issue where sometimes no models show up at all, ~~and cannot be added~~. There are no errors reported in the log and no indication that anything else is wrong. My best guess is that something in the model loading workflow is getting hung.

The behavior seems random, and usually restarting the application will allow all the models to come back. It is known to occur when loading the demo world.

Update: This seems to be related to the scenarios feature. The instances are not visible because they are stuck in the hidden mode. Trying to toggle the visibility of an instance doesn't work, it remains stuck on hidden.

mxgrey:
This issue is fixed by this commit: https://github.com/open-rmf/rmf_site/commit/c31381fd4ca8ec0f89535a25fc8e81a02a36ece2

We should get that into its own PR and merge it into main.

mxgrey:
Fixed by #282

PR #282, opened 2025-05-20
Fix accidental hiding of models

mxgrey:
This addresses #277


xiyuoh (approved):
Works great, thanks for the fix! Also tested out on known cases where modifying systems in the `Update` set would reliably cause models to be hidden at launch, this PR corrects the behavior.

PR #285, opened 2025-05-27
Update bevy_gltf_export branch and add Bevy patch

xiyuoh:
Following https://github.com/open-rmf/rmf_site/pull/281 we've found a bug in Bevy that panics when entities are spawned in `PostUpdate`. This is evident in the site editor as we currently experience crashes when creating lanes and locations. The fix is already available (https://github.com/bevyengine/bevy/pull/19064), but it may take some time before they're merged and released. Tested and verified that it stops the crashes.

This PR temporarily and manually adds a patch for Bevy `v0.16.0` with the fix cherry picked. This allows us to keep `main` functional. We can remove the patch when `v0.16.1` is released.

I've also updated the branch for `bevy_gltf_export` back to Luca's repo for general housekeeping.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -498,8 +498,7 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 [[package]]
 name = "bevy"
 version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a5cd3b24a5adb7c7378da7b3eea47639877643d11b6b087fc8a8094f2528615"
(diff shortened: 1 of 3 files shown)


luca-della-vedova (commented):
Thanks for the fix! There are a few other Cargo.toml fixes that I spotted, my proposal would be to bundle all the nice to have cleanups in one PR and have this just for the workaround (so we can easily revert the commit later).

Specifically:
[bevy_mod_outline](https://github.com/open-rmf/rmf_site/blob/main/rmf_site_editor/Cargo.toml#L19https://github.com/open-rmf/rmf_site/blob/main/rmf_site_editor/Cargo.toml#L19) was released so we can just use the released version.
[ron](https://github.com/open-rmf/rmf_site/blob/4930c0320fff41b4779d8666a563b97429084279/rmf_site_format/Cargo.toml#L15) was also released, so same comment there.

luca-della-vedova (approved):

PR #287, opened 2025-06-02
Migration follow-up: version bumps and bug fixes

xiyuoh:
Since https://github.com/open-rmf/rmf_site/pull/281 we've had a couple of bugs and housekeeping items to follow up:
- Use updated release / branches for `bevy_mod_outline`, `bevy_gltf_export`, and `ron`
- Extra `AssetSource` ComboBox in inspector, causing unnecessary widget for models/floors/walls and errors for duplicate ComboBox id
  - In the widget implementation we fix this by returning early if the selected entity is a non-group entity that does not have any affiliated group entity (e.g. models). This ensures that the `AssetSource` widget won't be duplicated for group/non-group affiliations.
- Fix wall and floor meshes missing during creation
  - We were doing mutable borrows of `Mesh3d` to update wall/floor meshes during creation when moving anchors around. With the migration it somehow caused the meshes to only render when the cursor is stationary.
  - Changing the implementation to modify the `Mesh` asset directly instead of the handle fixed this issue.

### DIFF
--- a/rmf_site_editor/Cargo.toml
+++ b/rmf_site_editor/Cargo.toml
@@ -17,7 +17,7 @@ path = "examples/extending_menu.rs"
 
 [dependencies]
 bevy_mod_outline = "0.10"
-bevy_gltf_export = { git = "https://github.com/luca-della-vedova/bevy_gltf_export" }
+bevy_gltf_export = { git = "https://github.com/luca-della-vedova/bevy_gltf_export", rev = "098fc7c" }
 bevy_stl = "0.16"
 bevy_obj = { version = "0.16", features = ["scene"] }
 smallvec = "*"
--- a/rmf_site_editor/Cargo.toml
+++ b/rmf_site_editor/Cargo.toml
@@ -16,8 +16,8 @@ name = "extending_site_editor"
 path = "examples/extending_menu.rs"
 
 [dependencies]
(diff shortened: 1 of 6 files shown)


rmf_site_editor/Cargo.toml
```diff
@@ -16,8 +16,8 @@ name = "extending_site_editor"
 path = "examples/extending_menu.rs"
 
 [dependencies]
-bevy_mod_outline = { git = "https://github.com/komadori/bevy_mod_outline", branch = "bevy-0.16"}
-bevy_gltf_export = { git = "https://github.com/xiyuoh/bevy_gltf_export", branch = "xiyu/bevy_0.16"}
+bevy_mod_outline = "0.10"
+bevy_gltf_export = { git = "https://github.com/luca-della-vedova/bevy_gltf_export" }
```
luca-della-vedova, line 20: Can you set a commit hash to the latest main here? To make sure it doesn't break unexpectedly

xiyuoh: [c73b477](https://github.com/open-rmf/rmf_site/pull/287/commits/c73b477a8f5f4c031dae0093d2f3779eff1f7db8)

luca-della-vedova (approved):

PR #293, opened 2025-06-03
moved crates to /crates (#290)

rydb:
### Implementation description

I moved all of the crates into a new /crates folder. to address #290.

I fixed all of the asset imports/CI paths to account for the new location.

Issue #288, opened 2025-06-02
[Other issue]: move crate dependencies to workspace dependencies.


rydb:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

For the sake of streamlining updating dependencies across crates, and to reduce the risk of duplicate dependencies, crate dependencies should be confederated under workspace dependencies, so imports can just be reduced to just: 

```toml
# dependent Cargo.toml
[dependencies]
bevy = {workspace = true}
dependency2 = {workspace = true, features = ["feature1", "feature2"]}
# etc etc


# root Cargo.toml
[workspace.dependencies]
bevy = {version = "0.16"}
dependency2 = {version = "???"}
```

PR #294, opened 2025-06-03
Workspace deps

rydb:
 
This pr builds on: https://github.com/open-rmf/rmf_site/pull/293
to convert all crate imports in sub-crates into workspace imports as described in #288

This will reduce the risk of duplicate crate imports from minor version differences and will make updating crate versions down the line much more straight forward.


rydb:
> I need to admit not being very familiar with this strategy but how does it work for downstream users that need the crate itself?

downstream users of the crate are not affected. They don't compile all workspace dependencies, only the ones that the crates that they import use. 

E.G: if there were two crates

robot_physics
robot_format

lets say robot_physics used `rapier`, and `robot_format` used `serde`. If an end user only imports `robot_format`, they won't need to compile `rapier`, only `serde`.

> Let's say we specify `bevy` as a workspace dependency so we don't need to specify it in both `rmf_site_editor` and `rmf_site_format`, if we were to publish `rmf_site_format` would a downstream user of the library be aware of its workspace dependency or not?

A crate that uses workspace dependencies and one that doesn't look the same to an end user as far as I'm aware. The only difference is if they'll need to look at [workspace.dependencies] to know the exact versions of things.

> Also this PR diff is massive because it includes the crate move as well

This pr is stacked ontop of a different pr. I need both of these pr's QOL improvements for the upcoming ones I'm writing.

https://github.com/open-rmf/rmf_site/pull/293

rydb:
Closed in favor of:
#313

PR #300, opened 2025-06-16
Add Battery component

xiyuoh:
Following [this discussion](https://github.com/open-rmf/rmf_site/pull/272#issuecomment-2966127138), this PR adds another robot property `Power Source` and property kind `Battery` to the site editor, and updates slotcar's SDF import/export workflow to include these battery-related parameters.

### DIFF
--- a/rmf_site_editor/src/site/slotcar.rs
+++ b/rmf_site_editor/src/site/slotcar.rs
@@ -16,11 +16,14 @@
 */
 
 use crate::site::{
-    robot_properties::*, Change, DifferentialDrive, ExportHandler, ExportHandlers, ExportWith,
-    Group, IsStatic, Mobility, ModelMarker, ModelProperty, ModelPropertyQuery, Robot,
+    Battery, Change, DifferentialDrive, ExportHandler, ExportHandlers, ExportWith, Group, IsStatic,
+    Mobility, ModelMarker, ModelProperty, ModelPropertyQuery, PowerSource, Robot, RobotProperty,
+    RobotPropertyKind,
 };
 use bevy::prelude::*;
+use rmf_site_format::robot_properties::*;
 use sdformat_rs::{ElementData, ElementMap, XmlElement};
+use serde_json::{Map, Value};
 
 pub struct SlotcarSdfPlugin;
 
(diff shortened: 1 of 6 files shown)


rmf_site_format/src/robot_properties.rs
```diff
@@ -399,3 +437,153 @@ impl Recall for RecallCircleCollision {
...
+            .and_then(|current| f64::try_from(current.data.clone()).ok())
+        {
+            battery.charging_current = current as f32;
+        }
+        if let Some(power) = elements
+            .get("nominal_power")
+            .and_then(|power| f64::try_from(power.data.clone()).ok())
+        {
+            battery.power = power as f32;
+        }
```
luca-della-vedova, line 551: Probably similar to my other comment on power, but what does this `nominal_power` actually mean? I'm starting to wonder if we should split the battery parameters into pure battery (voltage, capacity, _most likely_ charging current) and leave all the power parameters such as idle power, tool power, friction coefficient to another component

rmf_site_editor/src/widgets/inspector/mod.rs
```diff
@@ -212,30 +212,42 @@ impl Plugin for StandardInspectorPlugin {
...
+                    InspectModelPropertyPlugin::<InspectModelScale, Scale>::new(
+                        "Scale".to_string(),
+                    ),
+                    InspectModelPropertyPlugin::<InspectModelAssetSource, AssetSource>::new(
+                        "Asset Source".to_string(),
+                    ),
+                    InspectRobotPropertiesPlugin::default(),
+                    InspectRobotPropertyPlugin::<InspectMobility, Mobility, RecallMobility>::new(),
+                    InspectRobotPropertyPlugin::<InspectCollision, Collision, RecallCollision>::new(
+                    ),
```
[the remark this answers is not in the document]
xiyuoh: I can reduce it - it was automatically formatted due to an editor extension.

rmf_site_format/src/robot_properties.rs
xiyuoh: [7fc974f](https://github.com/open-rmf/rmf_site/pull/300/commits/7fc974f79a58c9b45b77fb8fc02d84a4674ab80b). Thanks for pointing this out, I think it's a good call to separate them. I added `PowerDissipation`, `MechanicalSystem` and `AmbientSystem` components to support slotcar parameters, and moved `nominal_power` to `AmbientSystem` as "idle power".

I didn't add in `ToolSystem` this time round since (a) they are not represented in Slotcar params and (b) we might want to think more about what fields to add in if and when we do want to include tool power drain as a robot property kind.

rmf_site_editor/src/site/slotcar.rs
```diff
@@ -147,31 +193,58 @@ impl Default for SlotcarParams {
...
+            Value::Number(speed) => speed.as_f64(),
+            _ => None,
+        }) {
+            self.nominal_turn_speed = rotational_speed as f32;
+        }
+
+        self
+    }
+
+    fn from_battery(&mut self, battery: &Value) -> &mut Self {
```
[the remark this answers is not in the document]
xiyuoh: [0db687e](https://github.com/open-rmf/rmf_site/pull/300/commits/0db687e9989b45267b05cd68883885c91b9db933)

rmf_site_format/src/robot_properties.rs
mxgrey: I have nothing against using physical battery properties to configure things, but one observation I've made over the years is that system integrators often don't know or care about having accurate parameters for battery properties. They often just think in terms of battery drain rate while moving vs while idle.

We can stick with the physical parameters for now, but down the road we might want to consider just operating in terms of simple "average rate of drain while moving", "average rate of drain while idle", and "average rate of charging" parameters.

luca-della-vedova: Would the rate in this case be in Wh or even one higher level and just be a battery percentage per unit time?
But yea I agree that the parameters are a bit hard to get right and something simpler would likely be preferred, we can figure out in a followup how to support both and migrate them

luca-della-vedova (approved):
Thanks! There are some merge conflicts but LGTM

mxgrey:
As a heads up, I'm about to take a stab at the merge conflicts.

mxgrey:
> I'm personally not a huge fan of exporting SDF with tags such as sphere, box, etc. The main reason is that it makes it impossible to interoperate with any other software (i.e. blender).

I can certainly understand this for the visual meshes, but I don't think collision shapes are important from blender's point of view, unless blender has a plugin for importing the SDF itself. But at that point I would expect an SDF importer should be able to handle the SDF primitive shapes.

> Find a way to parse the SDF and keep all the various geometries when exporting (including spheres, boxes, etc).

This is what I think might work best. We can store the original collision primitive information in a component and then do an `Option<&_>` query for it. Then whatever that primitive is will override what we export.

> since we don't even use wheels to apply torque anymore. ... I suppose we could just get the model's bounding box and export the model's collision as a cube.

I forgot that we aren't using the wheel physics anymore. I suppose that means the robot is essentially sliding across the ground all the time. Using a bounding box collision geometry will likely run into some problems since box collision geometries are not actually very stable (volume-volume collisions are most stable with 3 contact points rather than 4) and the box would be larger than the visual robot so it may get snagged on nearby objects.

I'll try testing this hypothesis about the physics by exporting the world and then manually replacing the robot model with an `<include>` of the original. If that gets it working correctly then I'll merge this an open a ticket for using the primitive geometries.

mxgrey:
For reference, this is what I'm seeing from the robot right now:

https://github.com/user-attachments/assets/46268235-d5f9-41ee-bf8f-f78d7a4fe691

It slides around a lot and then eventually gets stuck in a loop of overshooting its target.

luca-della-vedova:
> For reference, this is what I'm seeing from the robot right now:
> 
>  back-and-forth.mp4 
> It slides around a lot and then eventually gets stuck in a loop of overshooting its target.
> 
> One thing I'm noticing is the relative transforms of the nested parts look very wrong. E.g. the visual wheels are way off, the box collision geometry of the body is too low, and the sphere collision geometries are in the wrong place.

Ah that makes sense, when this was first implemented there was no model with a complex hierarchy so there was no proper support for nested hierarchies of links. This will probably also need support on the GLB exporter side. I'll have a look now that there is a model to test it with.

mxgrey (approved):
Let's go ahead and merge the heavy lifting this PR does and follow up with a PR to fix the transforms.

PR #302, opened 2025-06-17
Fix camera controls when the cursor is over a void

mxgrey:
During the migration to 0.16 we broke the logic for the camera controls when the cursor is hovering over a blank void. As a result the camera suddenly halts when the cursor tries to cross a void, and the user cannot do any panning or orbiting if they start over a void.

This PR fixes the logic for how we handle cases where there's nothing under the cursor to pick.


xiyuoh (approved):
Thanks for catching this and for the quick fix!

Issue #292, opened 2025-06-02
[Other issue]: Improve iterative compile times


rydb:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

The project's iterative compile times could definitely be improved.

running the below for a fresh build:
`
cargo run --example extending_site_editor --timings
`

my compile time came out to:

![Image](https://github.com/user-attachments/assets/d8e364db-38d5-4042-bee5-384186448e5d)

and on a second compile, appending rmf_site_editor/workspace/lib.rs/fn run()  with 

```rust
println!("hello!");
```

resulted in an iterative compile time of:

![Image](https://github.com/user-attachments/assets/25ecd92b-fd38-4034-b5b7-e116cf43f5e4)"

These compile times are with an ideal environment. I am running these tests on a high end gaming pc, on linux, and with the wild 0.4 linker.

Aside from compiling the project in nightly with cranelift, these are all of the optimizations that I am aware of that you can get when compiling the project as an end user in stable rust.

For lower end machines, and for machines that use windows(and so don't get to benefit from the mold/wild linker and for some reason, also have higher base-line compile times?), their times are likely much larger.

I've opened a few issues for things that can be resolved to get some compile time improvements.

#288, 
#289, 
#291

The biggest gains can be made by splitting up the project into smaller crates. Smaller gains being made by being more selective with imports in individual crates. 

Once all of the major improvements have been merged, this project's compile times could potentially shrink down to around 2-7 seconds on iterative compiles.

PR #306, opened 2025-06-19
rmf_site_editor meshes -> rmf_site_mesh split off

rydb:
### Why

rmf_site_editor's meshing is used throughout the project without needing anything within it self. Its also somewhat hard to read, being in a 1200 line long shapes.rs file. Mesh generation dependent features (I.E cursor.rs), also can't be split off until this is split off.

This pr splits that file into a /primitives folder for 3d shapes, and /shapes folder for 2d shapes/3d shape helper functions, it also adds a blank example to be expanded on for testing the mesh maker functions in.

This pr will contribute to:
#292
#292

and builds off of 
#303

- mesh generation is now split across several files instead of just one. 

- rmf_site_format's Angle has been severed off, instead, a minor mirror of `Angle` is used for rotation.

- Radians and Degrees are now inside of a new-type, and calculations for meshes take `impl Into<Radian>` instead of `Angle` to reduce enum matching.

### GenAI Use

- [ X] I did not use GenAI


rydb:
closed in favor of
#316

PR #314, opened 2025-06-25
Apply root inverse transform to exported SDF models

luca-della-vedova:
### Fixed bug

As noted in https://github.com/open-rmf/rmf_site/pull/300#issuecomment-2999497903, nested models were exported to SDF with the wrong transform.

### Fix applied

This PR changes the behavior of SDF exporting for non static models to apply the inverse transform of the model root to all the submeshes.
The rationale behind this change is that static models are all combined into a single "level mesh", so their transform can be embedded in the GLB node itself.
For non static models however, we want to have the pose embedded in the SDF's `<pose>` tag. The previous logic ignored transform components for all non static submeshes, however what we should really be doing is compute the inverse of the root pose of the model and apply it to all submeshes global transforms.

With this change this is what the exported office world looks like:

image

Note that there is still an issue when running actual simulations that seems to be related to robots overshooting their destination, I suspect it might be due to friction parameters that are not imported or exported in the SDF, I'll have a look


luca-della-vedova:
Note that #319 is a simpler approach to exporting models, but this PR is still a bugfix for models that include hierarchy of links, it just happens that for now only robots do

PR #319, opened 2025-06-27
Use SDF experimental parameters to export robots

luca-della-vedova:
### Fixed bug

This PR brings the "old" approach of integrating robots to the site editor of using `<include>` tags rather than building the SDF manually.
Specifically, experimental [parameter passing in SDF](http://sdformat.org/tutorials?tut=param_passing_proposal) is used to update the slotcar parameters of the robot, while the rest of the plugin is included as is.

### Fix applied

With this update, exported non-static models look the same:

```
    <model name="Coke">
      <static>false</static>
      <pose>17.70439 -5.2915354 0 0 0 1.5708001</pose>
      <link name="link">
        <collision name="collision">
          <geometry>
            <mesh>
              <uri>meshes/model_158_collision.glb</uri>
            </mesh>
          </geometry>
        </collision>
        <visual name="visual">
          <geometry>
            <mesh>
              <uri>meshes/model_158_visual.glb</uri>
            </mesh>
          </geometry>
        </visual>
      </link>
    </model>
```

Non static models that include a slotcar plugin instead, will look like this:

```
    <include>
      <uri>model://TinyRobot</uri>
      <name>tinyRobot1</name>
      <pose>10.433054 -5.5750957 0 0 0 0</pose>
      <experimental:params>
        <plugin action="replace" element_id="slotcar" filename="libslotcar.so" name="slotcar">
          <nominal_drive_speed>0.5</nominal_drive_speed>
          <nominal_drive_acceleration>0.25</nominal_drive_acceleration>
          <nominal_turn_speed>0.6</nominal_turn_speed>
          <nominal_turn_acceleration>1.5</nominal_turn_acceleration>
          <reversible>true</reversible>
          <nominal_voltage>12</nominal_voltage>
          <nominal_capacity>24</nominal_capacity>
          <charging_current>5</charging_current>
          <mass>20</mass>
          <inertia>10</inertia>
          <friction_coefficient>0.22</friction_coefficient>
          <nominal_power>20</nominal_power>
        </plugin>
      </experimental:params>
    </include>
```

Note that to get this to work I needed to update the `yaserde` dependency on the `sdf_rust_experimental` level (the old version didn't allow for `:` in XML field names). Among other things they removed the `Default` implementation / requirement for all structs that need to be serialized or deserialized.
This will cause errors whenever models that actually don't follow the specifications are used. For example, the [Coke](https://app.gazebosim.org/OpenRobotics/fuel/models/Coke) model had  the following `model.sdf`:

```
<sdf version="1.6">
  <model name='Coke'>
    <static>true</static>
    <link name="body">
      <pose>0 0 0 0 0 0</pose>

        <visual name="visual">
          <geometry>
            <mesh><uri>model://Coke/meshes/coke.obj</uri></mesh>
          </geometry>
        </visual>

        <collision name="collision">
          <geometry>
            <mesh><uri>model://Coke/meshes/coke.obj</uri></mesh>
          </geometry>
        </collision>

        <!-- TODO: Replace with accurate values of mass and inertia -->
        <inertial>
          <mass>1.0</mass>
          <inertia>
            <ixx>0.003366666666666667</ixx>
            <iyy>0.0008666666666666667</iyy>
            <izz>0.0041666666666666675</izz>
          </inertia>
        </inertial>
    </link>
  </model>
</sdf>
```

According to the SDF specification however, [all](http://sdformat.org/spec?ver=1.6&elem=link#inertia_iyz) elements of `inertia` are required, which means this model doesn't follow the specification and will fail to be deserialized.
Previously, becase `yaserde` used `Default` behind the scenes, it would just silently fallback to the default of `0.0`, which _happens to be_ what we need.
Since the model itself doesn't follow the specification my proposal is to fix this at the asset level and add tags for the missing inertia values set to their default of `0.0`. This is the only broken model in the demo world but I can't rule out that more models will have issues.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,8 @@ ron = "0.10"
 thiserror = "*"
 glam = { version = "0.29" }
 uuid = { version = "1.13"}
-yaserde = "0.7"
-sdformat_rs = { git = "https://github.com/open-rmf/sdf_rust_experimental", rev = "9fc35f2"}
+yaserde = "0.12"
+sdformat_rs = { git = "https://github.com/open-rmf/sdf_rust_experimental", branch = "luca/experimental_params_new_yaserde" }
 urdf-rs = { version = "0.7.3"}
 once_cell = "1"
 pathdiff = "*"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5201,7 +5201,7 @@ dependencies = [
  "urdf-rs",
  "utm",
  "uuid",
- "yaserde",
+ "yaserde 0.12.0",
 ]
 
 [[package]]
@@ -5231,7 +5231,7 @@ dependencies = [
  "thiserror 2.0.12",
  "urdf-rs",
  "uuid",
- "yaserde",
+ "yaserde 0.12.0",
 ]
 
 [[package]]
@@ -5458,13 +5458,13 @@ dependencies = [
 [[package]]
 name = "sdformat_rs"
 version = "0.1.0"
-source = "git+https://github.com/open-rmf/sdf_rust_experimental?rev=9fc35f2#9fc35f2b5f55afab37a4c1906c395fc22d069322"
+source = "git+https://github.com/open-rmf/sdf_rust_experimental?branch=luca%2Fexperimental_params_new_yaserde#0bd26acbb29f1585348ebd44e5ec4a1a56a91d31"
 dependencies = [
  "convert_case",
  "nalgebra",
+ "xml-rs",
  "xmltree",
- "yaserde",
- "yaserde_derive",
+ "yaserde 0.12.0",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -33,8 +33,8 @@ ron = "0.10"
 thiserror = "*"
 glam = { version = "0.29" }
 uuid = { version = "1.13"}
-yaserde = "0.7"
-sdformat_rs = { git = "https://github.com/open-rmf/sdf_rust_experimental", rev = "9fc35f2"}
+yaserde = "0.12"
+sdformat_rs = { git = "https://github.com/open-rmf/sdf_rust_experimental", branch = "luca/experimental_params_new_yaserde" }
```
arjo129, line 37: This is probably me being "Kay-Poh" but it might be worth also simply re-exporting `ya-serde` upstream inside `sdformat_rs` so we don't need to explicitly import it here.

luca-della-vedova: That's a good call, https://github.com/open-rmf/rmf_site/pull/319/commits/da7a35a05f63707965b01381702f3a8a5ee91f14

mxgrey (approved):
Thanks for working out this nice solution to the exporting issues we were having. I think this is definitely the right direction for us to go in, and with this we can finally use the site editor to generate working simulations for typical use cases.

I've opened #328 to track a few follow-up items that we can consider working on to generalize some of the ideas introduced by this PR.

PR #321, opened 2025-07-01
Add support for Chargers plugin

luca-della-vedova:
### Implemented feature

There was no support for exporting chargers at the SDF exporter level and, even if there was, legacy charger behavior has been broken for some time.
This PR adds charger information to the exported SDF world following the new approach in https://github.com/open-rmf/rmf_simulation/pull/146. Specifically, a new world plugin is added that is in charge of creating entities with a `Charger` component that `slotcar` will use to check for charger proximity.

I noticed that the `site.get_anchor()` API I introduced with the first prototype was allowing potential footguns since it allows fetching either a site or level anchor but actually all its uses needed either a site anchor (lifts) or a level anchor (doors), so I made it more limited.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/crates/rmf_site_format/src/site.rs
+++ b/crates/rmf_site_format/src/site.rs
@@ -233,11 +233,23 @@ impl Site {
         ron::de::from_bytes(s)
     }
 
-    /// Returns the anchor and its level, given the id
-    pub fn get_level_anchor(&self, id: u32) -> Option<(&Anchor, &Level)> {
-        self.levels
-            .values()
-            .find_map(|l| l.anchors.get(&id).map(|a| (a, l)))
+    /// Returns an anchor and its level (if it's a level anchor), given the id
+    pub fn get_anchor_and_level(&self, id: u32) -> Option<(&Anchor, Option<&Level>)> {
+        self.anchors
+            .get(&id)
+            .map(|site_anchor| (site_anchor, None))
+            .or_else(|| {
+                self.levels.values().find_map(|l| {
+                    l.anchors
+                        .get(&id)
+                        .map(|level_anchor| (level_anchor, Some(l)))
+                })
+            })
+    }
+
+    /// Returns an anchor given the id
+    pub fn get_anchor(&self, id: u32) -> Option<&Anchor> {
+        self.get_anchor_and_level(id).map(|(a, _)| a)
     }
 
     #[allow(non_snake_case)]
--- a/crates/rmf_site_format/src/legacy/nav_graph.rs
+++ b/crates/rmf_site_format/src/legacy/nav_graph.rs
@@ -140,10 +140,10 @@ impl NavGraph {
                 for (_, door) in &level.doors {
                     let door_name = &door.name.0;
                     let (v0, v1) = match (
-                        site.get_level_anchor(door.anchors.start()),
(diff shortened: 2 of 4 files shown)


crates/rmf_site_format/src/site.rs
```diff
@@ -233,10 +233,11 @@ impl Site {
         ron::de::from_bytes(s)
     }
 
-    pub fn get_anchor(&self, id: u32) -> Option<&Anchor> {
-        self.anchors
-            .get(&id)
-            .or_else(|| self.levels.values().find_map(|l| l.anchors.get(&id)))
+    /// Returns the anchor and its level, given the id
+    pub fn get_level_anchor(&self, id: u32) -> Option<(&Anchor, &Level)> {
```
mxgrey, line 237: This change is problematic in two ways:
* It will prevent level-specific anchors from being used as reference points for lifts. By default we use site-wide anchors for lifts, but there may be valid reasons to use a level-specific anchor, e.g. if some artifact on a certain level (such as a wall vertex) is useful as a reference point for how the lift should be positioned.
* It will prevent site-wide anchors from being used for elements like doors and walls. Being able to reference a site-wide anchor for level-specific elements is useful if multiple levels follow a certain pattern and users would like to keep that pattern synced across all levels.

It looks like the motivation behind this change is to figure out what level (and elevation) to place a charger. What I would recommend is to return an `Option<&Level>` from this function and if it has a `None` value then we place the charger on every level. I think that would be the most logical outcome to infer if a user has tied a location to a site-wide anchor.

I might also suggest we tweak the API like this:

```rust
pub fn get_anchor_and_level(&self, id: u32) -> Option<(&Anchor, Option<&Level>)> { ... }
pub fn get_anchor(&self, id: u32) -> Option<&Anchor> {
    self.get_anchor_and_level(id).map(|(a, _)| a)
}
```

mxgrey (changes requested):
Thanks for porting the charger plugin over!

I think the exporting itself looks good, but I left an in-line comment with a recommendation regarding the change from `get_anchor` to `get_level_anchor`.

crates/rmf_site_format/src/site.rs
luca-della-vedova: > This change is problematic in two ways:
> 
> * It will prevent level-specific anchors from being used as reference points for lifts. By default we use site-wide anchors for lifts, but there may be valid reasons to use a level-specific anchor, e.g. if some artifact on a certain level (such as a wall vertex) is useful as a reference point for how the lift should be positioned.
> * It will prevent site-wide anchors from being used for elements like doors and walls. Being able to reference a site-wide anchor for level-specific elements is useful if multiple levels follow a certain pattern and users would like to keep that pattern synced across all levels.
> 

Ah I was doing some testing and noticed that I couldn't assign any of these elements to a "diamond shaped anchor" and mistakenly thought those were the site anchors. I just realized that actually site and level anchors have the same visual mesh and the diamond anchors are lift anchors instead! I'll revert this.

> What I would recommend is to return an `Option<&Level>` from this function and if it has a `None` value then we place the charger on every level.

Sounds good

mxgrey (approved):
Looks great!

Issue #315, opened 2025-06-25
Each robot model instance is spawned on every level


mxgrey:
After loading the hotel world from `rmf_demos_maps` I noticed that all the robot model instances are spawned on every level in the site instead of only being spawned at the `location` that there meant to be.

When we load the robot model instances based on their `location`, they should be assigned to the same floor that the `location` is on.

While we're at it, we should probably make the level information a `Property` with a `Modifier` so users are able to move robots between levels, and assign the robot to different levels across different scenarios.

xiyuoh:
Closed as completed https://github.com/open-rmf/rmf_site/pull/327

PR #324, opened 2025-07-04
Spawn robots on their respective levels only

xiyuoh:
Fixes duplicate robots spawned on every level reported in https://github.com/open-rmf/rmf_site/issues/315. Remaining of the ticket will be addressed in a separate PR.


luca-della-vedova (approved):
Yap this seems to address it, thank you!

Issue #311, opened 2025-06-24
Reduce duplicate implementation across Property/Modifier code


xiyuoh:
The current modifier Property (Pose, Visibility, TaskParams) implementations are very similar and repetitive. Specifically for `Property::insert` (triggered when a Property is newly inserted into a scenario element entity) and `Property::insert_on_new_scenario` (triggered when a new scenario has been added to the site). The idea was to keep implementations flexible for users to customize, but since we have the `StandardProperty` trait available, we can consider moving commonly implemented logic in there.

Another area with huge overlap is the system in charge of handling modifier updates, i.e. `handle_instance_modifier_updates` and `handle_task_modifier_updates`. They each make use of their own `UpdateModifier` variant (`UpdateModifier<UpdateInstance>` and `UpdateModifier<UpdateTaskModifier>`) to read user input and process changes to the respective Modifiers. However, these enums can be reduced, e.g. from

```
pub enum UpdateInstance {
    Include,
    Hide,
    Modify(Pose),
    ResetPose,
    ResetVisibility,
}
```
to 
```
pub enum UpdateModifier<T: Property> {
    Modify(T),
    Reset,
}
```

With this, instead of exposing one `handle_<>_modifier_updates` system per scenario element (Instance, Task), we can have just a single `handle_modifier_updates<T: Property>` implemented in the PropertyPlugin. This helps us shield users from unnecessary interaction with modifier logic.

mxgrey:
Closed by #325

Issue #310, opened 2025-06-24
Investigate whether Modifier feature can be improved with observers


mxgrey:
Currently property modifiers are being updated with a scheduled `World` system `update_property_value`. We can consider whether this might be better to implement as an observer that immediately reacts to events.

We may also want to consider adding `#[require(ScenarioModifiers)]` to the `ScenarioMarker` component or otherwise adding `ScenarioModifiers` to `ScenarioBundle` and just skipping it for serialization. Maybe we would remove `ScenarioMarker` altogether at that point.

xiyuoh:
I'm currently working on this ticket on this [branch](https://github.com/open-rmf/rmf_site/tree/xiyu/improve_modifiers), regarding `ScenarioModifiers` I ended up adding it to `ScenarioBundle` and serialized it. I previously thought it'd duplicate data stored to file, but it would just add existing modifier entities as `SiteID`s. This helps us not have to manually insert modifier data in `ScenarioModifiers` in `load.rs`; without serializing we would have to do our own assignment of element Entity to modifier Entity from scratch.

mxgrey:
Closed by #325

PR #325, opened 2025-07-04
Improve Modifiers implementation

xiyuoh:
This PR targets https://github.com/open-rmf/rmf_site/issues/310 and https://github.com/open-rmf/rmf_site/issues/311 to achieve the following improvements:
- Make `ScenarioModifiers` a required component as part of `ScenarioBundle`
  - `ScenarioModifiers` is not meant to be manually edited by users, but we were storing them in a way where they have to be inserted with new data everytime a file is being loaded.
  - ~[88f0845](https://github.com/open-rmf/rmf_site/commit/88f0845a17e8032d44220f655b1b8dd75c6373d3) serializes `ScenarioModifiers` data to file during save, and loads them accordingly without need for manual intervention. The only time entities are inserted into `ScenarioModifiers` outside of `AddModifier` is when loading a legacy building map.~ **EDIT**: [64750ad](https://github.com/open-rmf/rmf_site/pull/325/commits/64750ada244b83d2d10cd6d2fee32ed906a00864) reverts the serialization of ScenarioModifiers. When loading, modifier entities are being created and sent to be inserted via the `AddModifier` event.
  - Additionally with `ScenarioModifiers` being required, there is no longer use for `ScenarioMarker` so that has been removed.
- Reduce code overlap: Combine Modifier implementations
  - Currently whenever we add in a new `PropertyPlugin` with a fresh set of Property-Modifier combination, we are implementing a new `handle_<>_modifier_updates()`, leading to a lot of duplicate code. 
  - There is also an `Update<>` enum for each of these combination to clarify what type of change has been made to what type of Modifier. This is however redundant as we can separate all changes for all Modifiers to `Reset` or `Modify`.
  - [8c902b3](https://github.com/open-rmf/rmf_site/commit/8c902b39b24b8c18140bdf2ea1a7fb960cedf65f) includes efforts to eliminate repeated implementations such that:
    - `handle_instance_modifier_updates`, `handle_task_modifier_updates` &rarr; `handle_modifier_updates<T>`
    - `UpdateInstance`, `UpdateTaskModifier` &rarr; `UpdateModifier<T>` where
      ```
      pub enum UpdateModifier<T: Property> {
          Modify(T),
          Reset,
      }
      ```
- Reduce code overlap: Helper functions in Property trait
  - Similar to above, whenever a new `PropertyPlugin` is added, the trait systems are implemented very similarly. [4ec4aa2](https://github.com/open-rmf/rmf_site/commit/4ec4aa272d9b4ea519171239eb88d9e0f38f60d4) adds in some helper functions to the trait impl so that we don't have to repeat the same logic over and over again.
- Moved some scheduled systems or observers from `site/mod.rs` into `PropertyPlugin`
  - When we only had Modifiers for Instances it wasn't much of an issue, but as more types of Property are added in, it's easy to miss out on individual systems that have to be manually inserted into the schedule to ensure the Property updates are functioning properly. By having them inside the `PropertyPlugin`, integrators don't have to worry about forgetting stranded systems.
  - Specifically `handle_cleanup_modifiers<T>` is now `on_remove_element` in the `PropertyPlugin`.
- Turn more systems into observers for immediate change
  - I previously opted to use regularly scheduled systems for these updates because it seemed like when changes are effective immediately, switching between scenarios looked spotty (changes happen one by one rather than all at one go). After more experimenting I observed that's no longer the case; changes for different properties happen one by one regardless, but e.g. when multiple poses change from one scenario to another, it still updates all at once.
  - The following systems are now observers:
    - `handle_scenario_modifiers` &rarr; `add_scenario_modifiers` + `remove_scenario_modifiers` (not typed so outside of PropertyPlugin)
    - `handle_modifier_updates<T, E>` &rarr; `on_update_modifier_event<T, E>` (in PropertyPlugin)
    - `update_property_value<T, E>` &rarr; `on_update_property<T, E>` (in PropertyPlugin)
- Introduced `Element` trait to help identify and constrain which components can be used in the `PropertyPlugin`. An `Element`, e.g. model instances and tasks, does not currently have any trait implementation, but it serves to differentiate which components may have certain properties vary across different scenarios.

Hopefully with these changes the way scenarios/modifiers/properties are being implemented becomes less cluttered and easier to use.


luca-della-vedova (dismissed):
As you noticed this breaks the serde_json round trip, the reason seems to be related to [this upstream issue](https://github.com/serde-rs/json/issues/989) that will likely not be fixed anytime soon (according to the maintainer, a Rust feature that is not stable yet will be needed for a proper fix).
There seem to be two fixes:

1) Remove `flatten` from the struct that contains the modifiers, [here](https://github.com/open-rmf/rmf_site/blob/3b718fe1768fbeabd50f8783528e1e76b274f55e/crates/rmf_site_format/src/scenario.rs#L92). I just tested this and it works.
2) The [solution mentioned in the issue](https://github.com/serde-rs/json/issues/989#issuecomment-1445215664) of using an external crate. I haven't given this a try.

My recommendation would be to just remove `flatten`, it is true that it saves us one layer of indentation but I would argue it can make it slightly trickier tot figure out which field belongs where by inspecting the format (there will be a flattened "name" that by visual inspection does not belong to the `Scenario` itself but one of its child structs).

crates/rmf_site_format/src/legacy/building_map.rs
```diff
@@ -562,6 +562,16 @@ impl BuildingMap {
...
+            // Insert scenario modifiers
+            let scenario = scenarios.get_mut(&default_scenario_id).unwrap();
+            scenario.instances.iter().for_each(|(id, _)| {
+                let modifier_id = site_id.next().unwrap();
+                scenario
+                    .properties
+                    .scenario_modifiers
+                    .0
+                    .insert(*id, modifier_id);
+            });
```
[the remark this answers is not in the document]
mxgrey: On that note, off the top of my head I can't think of any reason that we need to assign site IDs to modifiers.

I believe the key of [this map](https://github.com/open-rmf/rmf_site/blob/bffbb443c2208fb197b43e0001e9aff6bf1db5c5/crates/rmf_site_format/src/scenario.rs#L72) should be the site IDs of the instances, so there's no use assigning site IDs for the modifiers.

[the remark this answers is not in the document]
xiyuoh: The SiteID is being inserted here because we do not have existing ScenarioModifiers for legacy building maps, yet  it is a required component for scenarios, and when we import them we want to have it ready with data.

Previously, in `load.rs` we create the modifier entities in from InstanceModifiers/TaskModifiers and insert them into the ScenarioModifiers map. There was an offline discussion about not having to manually insert/remove entity pairs into ScenarioModifiers in `load.rs` (or anywhere else really aside from inside [AddModifier](https://github.com/open-rmf/rmf_site/blob/xiyu/improve_modifiers/crates/rmf_site_editor/src/site/modifier.rs#L169) and [RemoveModifier](https://github.com/open-rmf/rmf_site/blob/xiyu/improve_modifiers/crates/rmf_site_editor/src/site/modifier.rs#L218)). In order to achieve that, we would need the file to provide information about any scenario modifiers that exist in the site. That led to the choice to assign SiteID to modifier entities and storing them to file.

As I'm typing this I realize I misunderstood the point of the previous discussion, and should probably just trigger AddModifier events inside `load.rs` instead of serializing ScenarioModifiers. I'm gonna do that now.

crates/rmf_site_editor/src/site/load.rs
```diff
@@ -542,77 +543,58 @@ fn generate_site_entities(
         consider_id(*scenario_id);
 
         // Spawn instance modifier entities
-        let mut scenario_modifiers: ScenarioModifiers<Entity> = ScenarioModifiers::default();
-        for (instance_id, instance) in scenario_data.instances.iter() {
+        for (instance_id, instance_modifier) in scenario_data.instances.iter() {
             if let Some(instance_entity) = id_to_entity.get(&instance_id) {
```
mxgrey, line 547: It seems that this `if let Some(_)` will just quietly move on if the instance is missing. If the instance is missing then the file is malformed, and we should bring that up as an error.

To easily produce a site loading error for a missing instance, you can do
```rust
let instance_entity = id_to_entity.get(&instance_id).ok_or(instance_id).for_site(site_id)?;
```

crates/rmf_site_editor/src/site/modifier.rs
```diff
@@ -86,19 +83,41 @@ impl RemoveModifier {
     }
 }
 
+#[derive(Clone, Debug, Copy)]
+pub enum UpdateModifier<T: Property> {
+    Modify(T),
+    Reset,
+}
+
```
mxgrey, line 91: Minor quality of life suggestions:

```rust
impl<T: Property> UpdateModifier<T> {
    pub fn modify(
        scenario: Entity,
        element: Entity,
        value: T,
    ) -> UpdateModifierEvent<T> {
        UpdateModifierEvent {
            scenario,
            element,
            update_mode: Self::Modify(value),
            trigger_update_property: true,
        }
    }

    pub fn reset(scenario: Entity, element: Entity) -> UpdateModifierEvent<T> {
        UpdateModifierEvent {
            scenario,
            element,
            update_mode: Self::Reset,
            trigger_update_property: true,
        }
    }
}
```

crates/rmf_site_editor/src/site/load.rs
```diff
@@ -542,77 +543,58 @@ fn generate_site_entities(
...
+                    let visibility = if vis {
+                        Visibility::Inherited
+                    } else {
+                        Visibility::Hidden
+                    };
+                    commands
+                        .entity(modifier_entity)
+                        .insert(Modifier::<Visibility>::new(visibility));
+                }
+                commands.trigger(AddModifier::new(
```
mxgrey, line 567: I think this is now very very close to the kind of API that makes sense to me, but I would still change one thing.

Right now you're manually spawning an entity, adding `Modifier<T>` components to it, and then triggering the `AddModifier` event.

What we should be doing here instead is use the `UpdateModifier` event:

```rust
for (instance_id, instance_modifier) in scenario_data.instances.iter() {
    let instance_entity = id_to_entity.get(&instance_id).ok_or(instance_id).for_site(site_id)?;
    if let Some(pose) = instance_modifier.pose {
        commands.trigger(UpdateModifier::modify(scenario_entity, instance_entity, pose));
    }

    if let Some(vis) = instance_modifier.visibility {
        let visibility = if vis {
            Visibility::Inherited
        } else {
            Visibility::Hidden
        };
        commands.trigger(UpdateModifier::modify(scenario_entity, instance_entity, visibility));
    }
}
```

I think we should be able to absorb all the `add_scenario_modifiers` logic into `on_update_modifier_event`. Then we can get rid of the `add_scenario_modifiers` observer and `AddModifier` event, which should simplify the overall implementation.

Let me know if there's something I'm overlooking that makes it necessary to split `on_update_modifier_event` and `add_scenario_modifiers` into separate observers. As far as I can tell, `on_update_modifier_event` is just a superior observer to `add_scenario_modifiers`, as long as we move the logic for updating affiliations and parentage into `on_update_modifier_event`.

crates/rmf_site_editor/src/site/property_plugin.rs
```diff
@@ -200,3 +326,93 @@ fn on_add_root_scenario<T: Property, F: QueryFilter + 'static + Send + Sync>(
...
+    mut commands: Commands,
+) {
+    for scenario_entity in scenarios.iter() {
+        commands.trigger(RemoveModifier::new(trigger.target(), scenario_entity));
+    }
+}
+
+/// Whenever a Modifier<T> component has been removed from a modifier entit`y, this
+/// observer checks that the entity is not empty (has no other Modiifers). If
+/// empty, the modifier entity will be sent for removal and despawn.
```
mxgrey, line 343:
```suggestion
/// Whenever a Modifier<T> component has been removed from a modifier entity, this
/// observer checks that the entity is not empty (has no other Modifiers). If
/// empty, the modifier entity will be sent for removal and despawn.
```

crates/rmf_site_editor/src/site/property_plugin.rs
```diff
@@ -200,3 +326,93 @@ fn on_add_root_scenario<T: Property, F: QueryFilter + 'static + Send + Sync>(
...
+fn on_remove_modifier<T: Property>(
```
mxgrey, line 344: Is there a particular need to remove an empty modifier entity? If we just leave it in place, then it may get reused later rather than needing to respawn the modifier.

mxgrey (changes requested):
This PR has a lot of great improvements! I think it's very very close to what I believe the scenario modifier architecture should look like. I've left some inline comments about a few points that I think can be improved further. Most of it should actually involve deleting code.

Let me know what you think, especially if any of my suggestions aren't feasible for some reason. I'll keep reviewing and testing out this branch in parallel.

crates/rmf_site_editor/src/site/scenario.rs
```diff
@@ -244,9 +201,9 @@ pub fn check_selected_is_visible(
 
 /// Tracks pose changes for instances in the current scenario to update its properties
 pub fn update_model_instance_poses(
```
mxgrey, line 203: Right now this is a loose system specifically for synchronizing changes in `Pose` with changes in the current scenario, but this is something that should be happening for properties in general. The `PropertyPlugin` should add an observer that triggers on [`OnInsert`](https://docs.rs/bevy/latest/bevy/prelude/struct.OnInsert.html) which triggers an `UpdateModifierEvent` for the current scenario and the modified entity.

It seems like currently the visibility property is relying on the widget directly calling `UpdateModifierEvent`, which means if users introduced a plugin that directly changes the visibility of a model then those changes wouldn't be tracked in the scenarios.

In general the current approach won't scale well to covering other property types, so we should add a generic value tracking system into `PropertyPlugin`.

crates/rmf_site_editor/src/site/property_plugin.rs
```diff
@@ -79,104 +148,158 @@ impl<T: Property> LastSetValue<T> {
...
+
+    if event.trigger_update_property {
+        commands.trigger(UpdateProperty::new(event.element, event.scenario));
+    }
 }
 
-fn on_add_property<T: Property, F: QueryFilter + 'static + Send + Sync>(
+/// Updates the current scenario's property values based on changes to the property modifiers
+fn on_update_property<T: Property, E: Element>(
+    trigger: Trigger<UpdateProperty>,
```
mxgrey, line 236: I think the names of `UpdateProperty` and `UpdateModifier` are too similar.

The docs for `UpdateProperty` say
```rust
/// This event is triggered when changes have been made to an element in the
/// current scenario, and requires the element property value to be updated.
```
which I don't think is accurate anymore since this event seems to be mostly about switching the property to the modifier of a specific scenario. Maybe a name like `UseModifierForScenario` would be more accurate?

crates/rmf_site_editor/src/site/load.rs
xiyuoh: c2e5169

crates/rmf_site_editor/src/site/property_plugin.rs
xiyuoh: Yeah I think we can leave empty modifier entities alone. I would still remove modifier entities if the affiliated instances (or elements in general) are deleted from the site.

[2ba7ea3](https://github.com/open-rmf/rmf_site/pull/325/commits/2ba7ea3538df3eea80f24e4e15e9192ce7d30ec1)

crates/rmf_site_editor/src/site/scenario.rs
```diff
@@ -96,7 +67,7 @@ impl Property for Pose {
                 break;
             }
         }
-        // Insert visibility modifier into all root scenarios outside of the current tree as hidden
+        let mut new_visibility_modifiers = Vec::<(Modifier<Visibility>, Entity)>::new();
```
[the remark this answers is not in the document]
xiyuoh: Yeah this is a good call out, I've updated it in [ce33d1d](https://github.com/open-rmf/rmf_site/pull/325/commits/ce33d1de56f8e64f6c24c5216b537fbe6ef737ae)

crates/rmf_site_editor/src/site/scenario.rs
xiyuoh: Updated in [1fb583d](https://github.com/open-rmf/rmf_site/pull/325/commits/1fb583dd98b1b5957bcf23eb0a88de5757678087).

[a091dad](https://github.com/open-rmf/rmf_site/pull/325/commits/a091dad247fb5c9461a644ff705aa6df7e4d34c7) gets rid of the check to exclude `.is_added()`, this might be what you were thinking of from the start, but I realized we can use this tracking system to update modifiers newly added in the current scenario directly instead of inserting them in `Property::insert()`

crates/rmf_site_editor/src/site/property_plugin.rs
```diff
@@ -188,15 +283,64 @@ fn on_add_property<T: Property, F: QueryFilter + 'static + Send + Sync>(
...
+
+    for (entity, new_value) in changed_values.iter() {
+        if new_value.is_changed() && changed_last_set_value.get(entity).is_err() {
+            commands.trigger(UpdateModifier::modify_without_trigger(
+                scenario,
+                entity,
+                new_value.clone(),
+            ));
+        }
+    }
```
mxgrey, line 345: While testing out this branch I noticed a tiny glitch when I would drag around a model and then switch to a child scenario where that model's pose is still linked to the parent scenario. Sometimes, but not always, the model would twitch when I switch to the child scenario and then it would remain in that exact same location when I returned to the parent scenario.

After putting in some debug printouts, I realized that the problem is this system sometimes accidentally ignores a change to the pose, so the pose doesn't get propagated to the current scenario modifier.

It seems the `changed_last_set_value` query is meant to filter out changes to the property that happen due during an `UpdateModifierEvent`, but a race condition can happen where the property values gets changed _again_ in between the `UpdateModifierEvent` and the time that this system gets run. When that happens, the new change won't get registered.

The fix that I came up with as a suggestion here makes the filter a bit more exact. We check if the `LastSetValue` has changed within the past cycle _and_ we check whether the current property value is equal to what's stored in the `LastSetValue`.

```suggestion
fn update_changed_property<T: Property, E: Element>(
    mut commands: Commands,
    mut change_current_scenario: EventReader<ChangeCurrentScenario>,
    changed_values: Query<(Entity, Ref<T>, Option<Ref<LastSetValue<T>>>), (With<E>, Without<Pending>)>,
    current_scenario: Res<CurrentScenario>,
) {
    // Do nothing if scenario has changed
    for ChangeCurrentScenario(_) in change_current_scenario.read() {
        return;
    }
    let Some(scenario) = current_scenario.0 else {
        return;
    };

    for (entity, new_value, last_set_value) in changed_values.iter() {
        if new_value.is_changed() {
            if let Some(last_set_value) = last_set_value {
                if last_set_value.is_changed() {
                    // The new value might have been set by UpdateModifierEvent.
                    // If the last_set_value was changed on this cycle and it
                    // matches new_value then we take this to be the case.
                    // TODO(@mxgrey): Think of a more robust way to track this
                    if **last_set_value == *new_value {
                        continue;
                    }
                }
            }

            // The user has set a new value for this property, so we should
            // update its modifier in the current scenario.
            commands.trigger(UpdateModifier::modify_without_trigger(
                scenario,
                entity,
                new_value.clone(),
            ));
        }
    }
```

This has eliminated the glitch that I was seeing, and I suspect it will work fine for the way all our systems are currently implemented, but it does seem a bit fragile to me, so it may be worth considering some other approaches we might take for this.

Note that in addition to the above change, we need to add ` + PartialEq` to the trait constraints for `Property` and `StandardProperty`.

crates/rmf_site_editor/src/site/property_plugin.rs
```diff
@@ -188,15 +283,64 @@ fn on_add_property<T: Property, F: QueryFilter + 'static + Send + Sync>(
...
+    state: &mut SystemState<Query<&Affiliation<Entity>>>,
+    events_state: &mut SystemState<EventWriter<ChangeCurrentScenario>>,
 ) {
     let scenarios = state.get_mut(world);
     if !scenarios.get(trigger.target()).is_ok_and(|p| p.0.is_none()) {
         return;
     }
 
-    T::insert_on_new_scenario(trigger.target(), world);
+    T::insert_on_new_scenario::<E>(trigger.target(), world);
```
mxgrey, line 300: What if we generalize this a little bit, e.g.

```rust
T::on_new_scenario(trigger.target(), affiliation, world);
```

So `Property::on_new_scenario` is called for **_every_** new scenario and receives the `Affiliation` value for the scenario. Then the `Property::on_new_scenario` hook can be used for any new scenario, not just root scenarios, and implementations of the method can just check the affiliation value to see whether it's a root scenario or not.

mxgrey (changes requested):
Thanks for all the iterating on this, these changes are fantastic. It's a rare treat to see a net -600 lines of code while also improving an implementation.

I left some comments for just a few more small changes that will finally have this PR ready to merge. Most of the suggestions are just a little bit of reorganization and tweaking semantics, but one of them fixes a subtle glitch that is being caused by a race condition.

crates/rmf_site_editor/src/site/property_plugin.rs
xiyuoh: Ah thanks for tracking down the cause and suggesting a fix! I'll think about this more and follow up again. In the meantime added your fix in 195f92a

mxgrey (approved):
Thanks for all the iterating!

I've done a final look through the code and played around with the demo, exercising every aspect of the scenario feature that I can think of.

This is a huge improvement, and I think it's ready for prime time!

PR #332, opened 2025-07-09
rmf_site ui -> rmf_site_egui split off

rydb:
### Why

rmf_site's ui has several generic widgets within it which aren't rmf_site_editor specific and would be better off separated into their own crate for re-usability(panel.rs, menu_bar.rs, etc..).

Also, if this project migrates away from egui, this split off will be a necessary step to peel of egui from the editor.

This pr will contribute to:
#292
#334

- "generic" widgets and their plugins have been moved to this crate.

- crate has a default bevy 3d scene to test widgets in.

### GenAI Use

- [ X] I did not use GenAI


mxgrey (approved):
Looks good!

PR #333, opened 2025-07-10
Scenarios misc follow up items

xiyuoh:
This PR targets the remaining items in https://github.com/open-rmf/rmf_site/issues/254 to
- Enable users to optionally select a scenario as the `DefaultScenario`. When the current scenario is removed, the site editor will automatically switch to the `DefaultScenario` if one exists.
- Add a diagnostic to detect if a model instance in a child scenario is very close (<= 0.1m) to the same model in its parent scenario, suggesting it might have been accidentally shifted. The value of 0.1 is selected after a little experimentation in the editor - it's not easy to shift things around with a stricter threshold.

Other miscellaneous improvements:
- `text_edit_singleline` has a fixed width, and our text boxes do not have a consistent look when there are widgets inserted on their left or right. While setting the text boxes to adjust to the available width in `view_scenarios`, I extended the same changes throughout text boxes in the right panel.
- There is currently a diagnostic issue for the Robot Property `PowerDissipation` because it does not have a Robot Property Kind. Updated the diagnostic since we want to allow all types of Robot Properties, with Kinds or not.


mxgrey (commented):
I'm surprised that the default scenario isn't something that gets saved to file so that we open the selected default scenario automatically when loading a file.

It might also be something we should use when exporting worlds and nav graphs: Allow the user to specify the scenario they want to export, or else use the saved default scenario if none is specified.

However the features in this PR are valuable as-is, and we have more crucial developments in flight, so we can save these ideas for later.

Once #360 and #326 are merged, I'll fix merge conflicts in this PR and approve it.

mxgrey (approved):
I think we can move forward with this and consider #254 closed.

Any remaining parts of the conversation there could be spun out into their own issue tickets. I've opened #365 to follow up on saving, loading, and utilizing the default scenario information.

PR #336, opened 2025-07-16
Revive mapf debugger into editor

Nielsencu:
Revive old draft PR
https://github.com/open-rmf/mapf/pull/25

Added multi-robot path planning capability to the site editor.

Steps:
1. Click on robot, inspect robot properties on right hand panel, and add collision model manually (circle collision with radius 0.2m for instance)

2. Add robot GoToPlace tasks 
image

3. Click on generate plan

image

- I did not use GenAI


crates/rmf_site_editor/src/mapf_rse/negotiation/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    robot_descriptions: Query<(&DifferentialDrive, &CircleCollision)>,
+    tasks: Query<(&RobotTask, &GoToPlace)>,
+) {
+    if negotiation_request.len() == 0 {
+        return;
+    }
+    if !compute_negotiation_task.is_empty() {
+        let (mut entity, mut negotiation_task) = compute_negotiation_task
+            .single_mut()
+            .expect("Error: Found multiple negotiation tasks!");
```
luca-della-vedova, line 243: I'd recommend going through all the `expect`s in this PR and making sure they are _really_ critical and required.
An `expect` is pretty much an `unwrap` which means the application will panic if, for example in this case, there is more than one negotiation task.
We should try to aggressively reduce possible panics unless they are really critical and instead have more graceful error handling, most likely reporting to the user through `warn!` or `error!` macros.

crates/rmf_site_editor/src/mapf_rse/negotiation/visual.rs
```diff
@@ -0,0 +1,129 @@
...
+                .get(&proposal.0)
+                .and_then(|e| {
+                    robots
+                        .get(*e)
+                        .ok()
+                        .and_then(|affiliation| affiliation.0)
+                        .and_then(|desc| robot_descriptions.get(desc).ok())
+                        .map(|cc| cc.radius)
+                })
+                .unwrap_or(DEFAULT_PATH_WIDTH / 2.0);
```
luca-della-vedova, line 106: I always get a bit wary of these kind of `and_then` chains.
By doing this continuous `and_then` chain we "collapse" multiple functions that could return `None` into a single `None` option that is then unwrapped or assigned a default value.
I feel whenever we have something that can return `None` we should be careful and make sure that is a desired output, and if returning `None` is an error we should report it as a warning to the user rather than silently assigning a default value and continuing as normal.
For this specific case I would argue that:

* A missing entry in the entity map is a real error (not sure about how the map is built).
* An entity in the above map that points to a non existing robot is a real error (that just shouldn't happen?)
* A robot without affiliation is a real error (since every robot should have an affiliation)
* A broken reference to a description is a real error (since references should always be valid)

The type of behavior I would recommend in these cases is a `let else` chain with proper error reporting, such as:

```
let Some(entity_id) = entity_id_map.get(&proposal.0) else {
    warn!("[...]");
    continue;
};
```

We can still assign a default value of `DEFAULT_PATH_WIDTH / 2.0` if needed but there should be a very clear place where the `Option` can be `None` and we can default to a value, not a series of 4 statements that could fail at any place.

luca-della-vedova (changes requested):
I had a high level look at the code and left some review here and there about cleanups, dependency handling and error handling

crates/rmf_site_editor/src/mapf_rse/misc.rs
```diff
@@ -0,0 +1,89 @@
...
+            .properties
+            .insert(Collision::label(), serialized_collision);
+    }
+
+    commands
+        .spawn(description)
+        .insert(Category::ModelDescription)
+        .insert(robot)
+        .insert(ChildOf(site_entity));
+}
```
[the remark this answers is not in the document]
Nielsencu: Deleted the file

crates/rmf_site_editor/src/mapf_rse/negotiation/mod.rs
```diff
@@ -0,0 +1,360 @@
...
+    // Only loop tasks that have specified a valid robot
+    for (task, go_to_place) in tasks.iter() {
+        // Identify robot
+        let robot_name = task.robot();
+        for (robot_entity, robot_site_name, robot_pose, robot_group) in robots.iter() {
+            if robot_name == robot_site_name.0 {
+                // Match location to entity
+                for (_, location_name, Point(anchor_entity)) in locations.iter() {
+                    if location_name.0 == go_to_place.location {
+                        let Ok((_, _, goal_transform)) = anchors.get(*anchor_entity) else {
```
luca-della-vedova, line 305: If you are not going to use a variable (i.e. here doing `_`) you should just remove it from the `Query` altogether

crates/rmf_site_editor/src/mapf_rse/negotiation/visual.rs
```diff
@@ -0,0 +1,187 @@
...
+    path_visuals: Query<Entity, With<PathVisualMarker>>,
+    mut meshes: ResMut<Assets<Mesh>>,
+    robots: Query<&Affiliation<Entity>, With<Robot>>,
+    robot_descriptions: Query<&CircleCollision, (With<ModelMarker>, With<Group>)>,
+    current_level: Res<CurrentLevel>,
+) {
+    // Return unless complete
+    let Ok(negotiation_task) = negotiation_task.single() else {
+        return;
+    };
```
luca-della-vedova, line 42: There is a slightly odd mix of resources (i.e. `ResMut<NegotiationDebugData>`) and queries that we only expect to match a single entity (`Query<&NegotiationTask>`).
Imho a query that matches a single entity is a bit of a footgun, if something is expected to be an app wide single instance it should be a resource. If we expect it to be at most one per site then it should be an entity that is a child of the site.

crates/rmf_site_editor/src/mapf_rse/negotiation/visual.rs
```diff
@@ -0,0 +1,187 @@
...
+                        line_transform(&start_pos, &end_pos),
+                        Visibility::default(),
+                    ))
+                    .insert(PathVisualMarker)
+                    .insert(ChildOf(level_entity));
+            };
+
+            for (i, _waypoint) in proposal.1.meta.trajectory.iter().enumerate().skip(1) {
+                let start_pos = proposal.1.meta.trajectory[i - 1].position.translation;
+                let end_pos = proposal.1.meta.trajectory[i].position.translation;
```
luca-della-vedova, line 169: Do we need to enumerate here? It seems you are just using the index to access the array that you are iterating anyway so maybe we can use [windows](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) instead?

crates/rmf_site_editor/src/occupancy.rs
```diff
@@ -143,6 +143,17 @@ pub struct CalculateGrid {
     pub ignore: HashSet<Entity>,
 }
 
+impl CalculateGrid {
+    pub fn from(cell_size_in: f32, entities_to_ignore: HashSet<Entity>) -> Self {
```
luca-della-vedova, line 147: `from` is a bit of an ambiguous function name in Rust, since it can send the message that it is a function from the `From<T>` trait. I would recommend renaming it to `new`

crates/rmf_site_editor/src/site/util.rs
```diff
@@ -21,6 +21,19 @@ use bevy::prelude::*;
...
+    let yaw = dp.y.atan2(dp.x);
+    let tilt = dp.z.atan2(dp.x.abs());
+    let center = (*p_start + *p_end) / 2.0;
+    Transform {
+        translation: Vec3::new(center.x, center.y, center.z),
+        rotation: Quat::from_euler(EulerRot::ZYX, yaw, -tilt, 0.),
+        ..default()
+    }
+}
+
```
luca-della-vedova, line 36: This is pretty much fully duplicated with the function below. I can't quite understand, why is it necessary to remove scaling from this function and add it to the mesh instead?
I would recommend reusing the `line_stroke_transform` instead. Another useful side effect is that if you spawn the same mesh multiple times Bevy will instance it and make rendering much more efficient as shown in [this example](https://bevy.org/examples/shaders/automatic-instancing/)

crates/rmf_site_editor/src/mapf_rse/negotiation/mod.rs
```diff
@@ -0,0 +1,360 @@
...
+                {
+                    Some(grid)
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        })
+        .next();
```
luca-della-vedova, line 277: `filter_map().next()` can just be written as `find_map()`

crates/rmf_site_editor/src/mapf_rse/negotiation/mod.rs
```diff
@@ -0,0 +1,360 @@
...
+                occupancy.entry(cell.y).or_default().push(cell.x);
+            }
+            for (_, column) in &mut occupancy {
+                column.sort_unstable();
+            }
+        }
+        None => {
+            occupancy.entry(0).or_default().push(0);
+            warn!("No occupancy grid found, defaulting to empty");
+        }
```
luca-della-vedova, line 291: I find this behavior a bit confusing.
I understand that if there is no occupancy grid it might not be very obvious how to generate one (you would need to send an event to another system and "temporarily pause" this one.
What I noticed however, is that if no occupancy grid is available and users ask to generate a plan, this warning will be printed, the plan will be generated according to an empty occupancy grid _and_ a new occupancy grid will be generated that makes the plan instantly invalid.
If using the newly generated occupancy grid for the plan is not possible I would recommend also not triggering the generation of a new one or the result can be quite confusing

luca-della-vedova (commented):
A few more scattered comments, some minor some a bit more substantial.
Can you look into the Z fighting for the occupation grid? There is some flickering happening because (I imagine) the grid's Z coordinates are too close to the floor so _sometimes_ it is rendered above while sometimes it is rendered below

image

crates/rmf_site_editor/src/occupancy.rs
luca-della-vedova: Or a possibly cleaner alternative could be to implement `Default` for the structure with sensible default values, then users can just instantiate it by doing:

```rust
let grid = CalculateGrid {
  cell_size,
  ignore,
  ..default()
};
```

crates/rmf_site_editor/src/mapf_rse/negotiation/mod.rs
Nielsencu: This is fixed by this PR https://github.com/open-rmf/rmf_site/pull/339, under point 2.

luca-della-vedova (approved):
Thanks for iterating on this! It's great to see this work finally ready :)

Issue #312, opened 2025-06-24
Improvements for task system UI


mxgrey:
A few follow-up ideas to consider for improving the task system implementation:
- [ ] The `Fleet:` value should be possible to infer by the choice of robot. At some point we should encode that kind of information on the robot itself
- [ ] `Fleet name:` is redundant with `Fleet:` in direct task. For dispatch task we might want to consider putting it under `More` and making it togglable.
- [ ] Items under `More` seem to flicker for a few frames each time they get toggled. I think this means we're not tracking their states quite right.
- [ ] The right-side panel is getting very cluttered. Maybe the tasks should go on the left side instead?
- [ ] I like the way the added tasks are in view-only panels, but it's a bit clunky that when you ask to edit them you have to do that in another place. I think it would be nice to be able to edit them in place.
- [ ] Maybe the form for filling in a new task should show up in the middle of the canvas to use the screen real estate a bit better.
- [ ] Make [`tasks:`](https://github.com/open-rmf/rmf_site/blob/cd224daa5c81411856aefc6444c6a74be9858d13/crates/rmf_site_format/src/scenario.rs#L73) optional for serialization and deserialization

PR #345, opened 2025-07-23
Improve tasks UI

xiyuoh:
This PR targets https://github.com/open-rmf/rmf_site/issues/312 to bring in the following improvements to the task UI:
- Add `fleet` param to `Robot`
- Move `Fleet` field into `More` for dispatch tasks, and remove redundant `Fleet name` for all task requests
- Moved tasks to toggle-able left panel, accessible through the top menu bar (new button next to Browse Fuel)
- Creating new tasks happen in a separate window that appears in the center of the screen
- In-place editing for existing tasks, with the exception of task kind description due to the need to access `world`
- Optional serialization for tasks and other modifier fields
- **[EDIT]**: Resolve  `More details` button flicker


xiyuoh:
After a little bit of poking around, it turns out that the UI flicker when clicking on the `More details` button was caused by having `egui::Grid` embedded inside a `CollapsingHeader`. Removing it fixes the issue. The grid was added to provide some alignment, but I think it's not worth having the entire UI glitching.

Another remaining bug is the the one-frame delay when toggling the `Start time` / `Request time` / `Priority` checkboxes. but I think they will be resolved with https://github.com/open-rmf/rmf_site/pull/325 where we handle `UpdateModifierEvent` with observers.

mxgrey:
It might be good for @Nielsencu to take a look at this and think about how to align these changes with some of the things we've discussed for the debugging tool. One similarity is that tasks have been moved to a toggleable panel on the left, similar to what we've discussed for the debugging tool.

Nielsencu:
I was thinking if it would make sense to combine both the tasks panel and debugger panel into one? The left panel would be divided vertically into two, where the tasks section would sit above the MAPF debugger section. I am not sure there are other functionalities for tasks at the moment, as far as I know they are only used for debugging MAPF. Given my understanding is correct, I think it makes sense to put them into one panel. We can even have a checkbox that toggles on and off the MAPF debugger section, if users only want to add tasks but do not want the debugger functionality (though I currently do not see the use case of only adding tasks without debugger).

assets/demo_maps/test.site.json
```diff
@@ -1013,6 +1013,7 @@
   },
   "robots": {
     "52": {
+      "fleet": "HospitalRobot",
```
luca-della-vedova, line 1016: Should we add tasks to this file so we make it part of our CI pipeline? It can be done in a followup if you prefer

crates/rmf_site_editor/src/widgets/tasks/mod.rs
```diff
@@ -337,7 +301,7 @@ fn show_task(
     } else {
         Color32::default()
     };
-    let mut edit = false;
+    let in_edit_mode = edit_task.0.is_some_and(|e| e == task_entity);
```
luca-della-vedova, line 304: It's not exactly here (a few lines below) but I'm a bit conflicted on using the task's `Entity` value for its display name, has there been a thought on having a `NameInSite` component for tasks? We could auto-generate it with a random ID the way we do in the RMF stack. Or if we want to go for IDs maybe just the `SiteID`? `Entity` values can be pretty random, I noticed that creating a new task in my test (screenshots in main review comment) created `Task 1078`, and creating a new one immediately after resulted in `Task 1081`, I would argue the very high + not even contiguous number is not too informative.

luca-della-vedova (commented):
I gave this a spin and found a few issues:

`Location` for `GoToPlace` and `Duration` for `WaitFor` tasks are in the wrong UI block when doing inplace edits, specifically they don't seem to be contained in the task itself but in the root left side panel

image

and

image

It seems that changing a task type can reorder it in the list of tasks, for example starting from the state above, if I change `Task 1081` from `WaitFor` to `GoToPlace` it gets automatically moved to the top of the list. For this I would suggest having a static sorting strategy, perhaps in alphabetical order of the task name?

image

I also noticed that `GoToPlace` tasks without a set location can still be added (both before and after this PR), is this intentional? Since a `GoToPlace` without a destination is invalid I feel the `Add task` button should be greyed out if it's not set.

image

xiyuoh:
@luca-della-vedova @Nielsencu Thanks both for the detailed review and suggestions! I believe I managed to address all of them:
- `TaskKind` widgets (`GoToPlace`, `WaitFor` UI) no longer below the task list, but within the same task widget where edits are being made ([f129a4a](https://github.com/open-rmf/rmf_site/pull/345/commits/f129a4a9f1e9c38eb4e8fb50ff0be7dd61fc253f))
- Order of tasks could be changed if edits were made - this was initially set as a future TODO to incorporate sorting based on `start_time` and `request_time`, but I just added in a simple implementation this time. ([a2390aa](https://github.com/open-rmf/rmf_site/pull/345/commits/a2390aab543220b8a2aa2fc61ba652184dc29c7b)) The tasks are now sorted as such:
  - If the task has a non-empty `start_time`, we'll use it for sorting.
  - If the task does not have a `start_time`, but has a non-empty `request_time`, we'll use the `request_time` for sorting.
  - If the task does not have `start_time` or `request_time`, we'll use its `created_time` for sorting. `created_time` is newly added and will just be the time at which the Task was first created (when `Create New Task` button is pressed).
  - Should not happen BUT if somehow a `created_time` was not successfully calculated and added to the Task upon creation, then the task will be pushed to the end of the queue and sorted by entity index. 
- `TaskKind` validity - `GoToPlace` now requires a valid location for the task to be added to the list. Along with this we can now define a `IsTaskValidFn` for each `TaskKind`. ([c23dffa](https://github.com/open-rmf/rmf_site/pull/345/commits/c23dffa010e8ee30afc766df7fcf4197febedf00))
- Use Entity to identify Robots and Locations in tasks instead of String names ([8c278e2](https://github.com/open-rmf/rmf_site/pull/345/commits/8c278e20f6c3db6544e7c098204a2bf0f38f6aca), [f67cc3b](https://github.com/open-rmf/rmf_site/pull/345/commits/f67cc3b127d8f98eda81251b6d6888a674b8665c))
- Highlight Robots and Locations when they are hovered over during task creation ([4946083](https://github.com/open-rmf/rmf_site/pull/345/commits/49460836f7f4c56550a765885a80091e97e394f8))
- Use timestamp as task identifier instead of Entity index ([ec9443d](https://github.com/open-rmf/rmf_site/pull/345/commits/ec9443d03438454922c75189f490b04d11bcc8eb))
- Added sample tasks to `test.site.json` for CI

Other add-ons:
- Applied alphabetical sorting for Locations when configuring `GoToPlace`, and Robots for Direct tasks.
- Fix task saving/loading - the previous implementation was able to save tasks but not load them. All the fields in `TaskParams` were also set to `skip_if_default`, but we should really save all of them to avoid confusion between empty modifiers and no modifiers. Done in [ec9443d](https://github.com/open-rmf/rmf_site/pull/345/commits/ec9443d03438454922c75189f490b04d11bcc8eb) and [3c43a73](https://github.com/open-rmf/rmf_site/pull/345/commits/3c43a73999078357ec7f4c4a8bee9967d898c02a)

crates/rmf_site_editor/src/widgets/tasks/mod.rs
```diff
@@ -337,481 +289,793 @@ fn show_task(
...
+                ui.horizontal(|ui| {
+                    ui.label("Fleet:").on_hover_text(
+                        "(Optional) The name of the fleet for this robot. \
+                            If specified, other fleets will not bid for this task.",
+                    );
+                    if !in_edit_mode {
+                        ui.label(task_request.fleet_name().unwrap_or("None".to_string()));
+                    } else {
+                        edit_fleet_widget(ui, &mut new_task);
+                    }
```
luca-della-vedova, line 593: I noticed that on my machine "some" tooltips don't work, specifically `Fleet` doesn't but `Requester` does, can you reproduce?
I also wonder since we are pre-populating the fleet, do we actually want the fleet parameter to be editable or should it just be fixed?
Making it editable creates a risk of disconnect, for example `TinyRobot1` is spawned in the world as part of `Fleet1`, users decide to assign a direct task to it and choose it from the dropdown, the fleet is automatically initialized to `Fleet1`, which is great.
However, the field is editable so users can change the fleet name to `Fleet2` and still create a valid task but it seems to me that is an invalid task that should not be allowed.
What do you think?

xiyuoh: Yeah I think you're right, it was quite an inconsistent design. Though now we do not have any other way to modify a robot's fleet, which was the intention of the text edit UI. I'll add one to the robot properties widget and make this non-editable.

mxgrey (commented):
Apologies for the absurd delay in doing a final review.

I think this UX is the best we can achieve with egui. The way the task fields adjust based on the type of task is particularly nice. But the overall experience highlights to me that egui doesn't scale well to increasing complexity of user interaction.

I just have a few incremental suggestions after mocking some use cases:
* What would you think of getting rid of the `fleet` field for `Robot` and just using the `NameInSite` that comes with `ModelDescription`? If I understand correctly there's a 1:1 relationship between each `ModelDescription` and `Robot`, so having a different `NameInSite` vs "fleet name" for the same model group seems redundant.
* In the `▸ Group` drop-down tab there's a `▸ Models Descriptions` drop-down that we can use to access fleets, but the fleets are drowned out by furniture assets. What if we add a `▸ Robots` drop-down that filters to only model descriptions that have a Robot component? I imagine that would make it easier for users to find robot descriptions to edit them.

xiyuoh:
Thanks for review and for helping to resolve some of the file conflicts previously!

> What would you think of getting rid of the fleet field for Robot and just using the NameInSite that comes with ModelDescription? If I understand correctly there's a 1:1 relationship between each ModelDescription and Robot, so having a different NameInSite vs "fleet name" for the same model group seems redundant.

That makes sense to me. Initially I was bundling ModelDescription to asset, and wanted to have a separate `fleet_name` field to identify between different fleets of the same ModelDescription, but we can always just import the same asset again under a different description for that kind of separation. Added in [bf509f4](https://github.com/open-rmf/rmf_site/pull/345/commits/bf509f4a918c408c374ef1aa30ebcb146ffd6c09).

> In the ▸ Group drop-down tab there's a ▸ Models Descriptions drop-down that we can use to access fleets, but the fleets are drowned out by furniture assets. What if we add a ▸ Robots drop-down that filters to only model descriptions that have a Robot component? I imagine that would make it easier for users to find robot descriptions to edit them.

Good idea, added in [500599f](https://github.com/open-rmf/rmf_site/pull/345/commits/500599f9bd29547b0a1dcbe79e4d4c01c9649481), but now Model Descriptions drop-down will not contain any robots - had to separate them due to querying rules.

--

An edge case I've found is if we
- Create a Direct task (e.g. GoToPlace for `TinyRobot/tinyRobot1`)
- With tinyRobot1 selected, toggle its `Robot` property off under Description Properties
- The task originally assigned to it would be be listed as assigned to `TinyRobot/<Unnamed>`
- If we toggle it back on, it'd be assigned back to `TinyRobot/tinyRobot1` again

I considered converting the Direct task into a Dispatch task if that happens, but we'd need some form of Recall implementation if the `Robot` property is toggled back and forth. Not a huge issue but something to think about in the future.

xiyuoh:
Ready for final review! CI was failing previously but re-running it did the trick

mxgrey (approved):
Thanks for all these improvements 🎉

Issue #340, opened 2025-07-21
Improve error handling for scenario saving


luca-della-vedova:
While working on #338 I noticed that saving scenarios makes aggressive use of `filter_map` when saving scenarios which has the effect of silencing errors and continuing when something unexpected is found, when I suspect the real behavior should be to explicitly return an error (or at least print a warn! if we want to still drop the value but continue) so the user is not surprised by missing elements.

For example:

* https://github.com/open-rmf/rmf_site/blob/69a65a415acaabfb5ee0e363fcc278b3e57ae7c9/crates/rmf_site_editor/src/site/save.rs#L1428-L1430 is a bit unclear to me, we iterate through the scenario's instance modifiers. If the `instance_modifier` query fails shouldn't we return an error for a broken reference to a modifier rather than silently continue and filter it out?
* https://github.com/open-rmf/rmf_site/blob/69a65a415acaabfb5ee0e363fcc278b3e57ae7c9/crates/rmf_site_editor/src/site/save.rs#L1431-L1441 where we silently drop the modifier if the affiliation doesn't point to a model instance, while I believe this should be a real error.

The same happens for tasks [here](https://github.com/open-rmf/rmf_site/blob/69a65a415acaabfb5ee0e363fcc278b3e57ae7c9/crates/rmf_site_editor/src/site/save.rs#L1447) and [here](https://github.com/open-rmf/rmf_site/blob/69a65a415acaabfb5ee0e363fcc278b3e57ae7c9/crates/rmf_site_editor/src/site/save.rs#L1451)

For the scenario parent https://github.com/open-rmf/rmf_site/blob/69a65a415acaabfb5ee0e363fcc278b3e57ae7c9/crates/rmf_site_editor/src/site/save.rs#L1461-L1469 we silently change the affiliation to `None` if it points to an invalid parent, which I believe should also be a proper error.

I haven't looked at other parts of the codebase but if we follow a similar pattern we should perhaps be more careful on how we handle errors to minimize unexpected behaviors

mxgrey:
Agreed, I commented on a specific case of it in [this review](https://github.com/open-rmf/rmf_site/pull/325#discussion_r2218198831). I think in most cases we can follow the recommendation there: Use `.ok_or(instance_id).for_site(site_id)?;` to convert into a site loading error and return early.

[...]

PR #347, opened 2025-07-28
Better error handling for saving scenarios

xiyuoh:
Targets https://github.com/open-rmf/rmf_site/issues/340 to add better error handling for scenario saving.

Note that `BrokenAffiliation` and the newly introduced `EmptyAffiliation` are different: `BrokenAffiliation` points to an existing entity wrapped in `Affiliation` that cannot be queried from its intended group, while `EmptyAffiliation` alerts users of an empty `Affiliation` when it should always be populated.

### DIFF
--- a/crates/rmf_site_editor/src/site/save.rs
+++ b/crates/rmf_site_editor/src/site/save.rs
@@ -46,6 +46,8 @@ pub enum SiteGenerationError {
     BrokenAnchorReference(Entity),
     #[error("an object has a reference to a group that does not exist")]
     BrokenAffiliation(Entity),
+    #[error("an object has a reference to an empty group")]
+    EmptyAffiliation(Entity),
     #[error("an object has a reference to a level that does not exist")]
     BrokenLevelReference(Entity),
     #[error("an object has a reference to a nav graph that does not exist")]
@@ -62,6 +64,8 @@ pub enum SiteGenerationError {
         "lift door {door:?} is referencing an anchor that does not belong to its lift {anchor:?}"
     )]
     InvalidLiftDoorReference { door: Entity, anchor: Entity },
+    #[error("an object has a reference to a modifier that does not exist")]
+    BrokenModifier(Entity),
 }
 
 /// This is used when a drawing is being edited to fix its parenting before we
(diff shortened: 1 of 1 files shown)


crates/rmf_site_editor/src/site/save.rs
```diff
@@ -1425,12 +1429,17 @@ fn generate_scenarios(
...
+                                        let (pose, visibility, affiliation) = instance_modifiers
+                                            .get(*e)
+                                            .map_err(|_| SiteGenerationError::BrokenModifier(*e))?;
+                                        let a = affiliation
+                                            .0
+                                            .ok_or(SiteGenerationError::EmptyAffiliation(*e))?;
+                                        Ok((
+                                            instances.get(a).map(|id| id.0).map_err(|_| {
+                                                SiteGenerationError::BrokenAffiliation(a)
+                                            })?,
```
luca-della-vedova, line 1442:
```suggestion
                                        let a = affiliation
                                            .0
                                            .ok_or(SiteGenerationError::EmptyAffiliation(*e))?;
                                        let id = instances.get(a).map(|id| id.0).map_err(|_| {
                                            SiteGenerationError::BrokenAffilation(a)
                                        })?;
                                        Ok((
                                            id,
```

Big `Ok` statements are a bit of a pain to read

crates/rmf_site_editor/src/site/save.rs
```diff
@@ -1440,31 +1449,39 @@ fn generate_scenarios(
...
+                                        let (inclusion, task_params, affiliation) = task_modifiers
+                                            .get(*e)
+                                            .map_err(|_| SiteGenerationError::BrokenModifier(*e))?;
+                                        let a = affiliation
+                                            .0
+                                            .ok_or(SiteGenerationError::EmptyAffiliation(*e))?;
+                                        Ok((
+                                            tasks.get(a).map(|id| id.0).map_err(|_| {
+                                                SiteGenerationError::BrokenAffiliation(a)
+                                            })?,
```
luca-della-vedova, line 1465: Same as above about bring the logic outside of the Ok statement to make it a bit easier to follow

luca-della-vedova (approved):

PR #348, opened 2025-07-28
Add grid and new arrow mesh

althea28:
This PR aims to:
- Add an infinite grid for better visualisation and guidelines in the world space
- Update the arrow mesh used for pivots

### DIFF
--- a/crates/rmf_site_mesh/src/primitives/arrow.rs
+++ b/crates/rmf_site_mesh/src/primitives/arrow.rs
@@ -123,14 +123,14 @@ pub fn make_triangular_arrow() -> MeshBuffer {
         [0.0, 0.3, 0.4],  // 4
     ];
 
-    let base_indices: Vec<u32> = vec![0, 1, 4, 0, 4, 2, 3, 4, 1, 3, 2, 4, 0, 3, 1, 0, 2, 3];
+    let base_indices: Vec<usize> = vec![0, 1, 4, 0, 4, 2, 3, 4, 1, 3, 2, 4, 0, 3, 1, 0, 2, 3];
 
     let indices: Vec<u32> = (0..18).collect();
 
     let positions: Vec<[f32; 3]> = base_indices
         .clone()
         .into_iter()
-        .map(|idx| base_positions[idx as usize])
+        .map(|idx| base_positions[idx])
         .collect();
 
     let normals: Vec<[f32; 3]> = (0..base_indices.len())
@@ -138,13 +138,13 @@ pub fn make_triangular_arrow() -> MeshBuffer {
         .into_iter()
         .map(|idx| {
             let chunk = [
-                base_indices[idx as usize],
-                base_indices[(idx + 1) as usize],
-                base_indices[(idx + 2) as usize],
+                base_indices[idx],
+                base_indices[idx + 1],
+                base_indices[idx + 2],
             ];
-            let p0: Vec3 = base_positions[chunk[0] as usize].into();
-            let p1: Vec3 = base_positions[chunk[1] as usize].into();
-            let p2: Vec3 = base_positions[chunk[2] as usize].into();
+            let p0: Vec3 = base_positions[chunk[0]].into();
+            let p1: Vec3 = base_positions[chunk[1]].into();
+            let p2: Vec3 = base_positions[chunk[2]].into();
             let n = (p1 - p0).cross(p2 - p0).normalize();
 
             n.into()


crates/rmf_site_editor/src/shapes.rs
luca-della-vedova, line 1: It seems this PR adds the `shapes.rs` file that was removed in https://github.com/open-rmf/rmf_site/commit/fa8f18f166745e5fdf7f6a9565076c30f79d5cb2, can you have a look and port your changes to the relevant files in `rmf_site_mesh` and remove the file from this PR?

crates/rmf_site_editor/src/interaction/assets.rs
```diff
@@ -234,7 +234,8 @@ impl FromWorld for InteractionAssets {
         let dagger_mesh = meshes.add(make_dagger_mesh());
         let halo_mesh = meshes.add(make_halo_mesh());
         let camera_control_mesh = meshes.add(Mesh::from(primitives::Sphere::new(0.02)));
-        let arrow_mesh = meshes.add(make_cylinder_arrow_mesh());
+        // let arrow_mesh = meshes.add(make_cylinder_arrow_mesh());
+        let arrow_mesh = meshes.add(make_triangular_arrow_mesh());
```
luca-della-vedova, line 238: Remove commented code

crates/rmf_site_editor/src/interaction/assets.rs
```diff
@@ -138,12 +138,12 @@ impl InteractionAssets {
             (
                 self.x_axis_materials.clone(),
                 Vec3::new(offset, 0., height),
-                Quat::from_rotation_y(90_f32.to_radians()),
+                Quat::from_rotation_y(90_f32.to_radians()) * Quat::from_rotation_z(90_f32.to_radians()),
```
luca-della-vedova, line 141: Why do these need to be rotated? If they need to can we just do it in a single constructor (i.e. building from [euler angles](https://docs.rs/bevy/latest/bevy/math/struct.Quat.html#method.from_euler)) rather than with a multiplication?

crates/rmf_site_editor/Cargo.toml
```diff
@@ -52,6 +52,7 @@ nalgebra = {workspace = true}
 anyhow = {workspace = true}
 uuid = { workspace = true, features = ["v4"] }
 bytemuck = {workspace = true}
+bevy_infinite_grid.workspace = true
```
luca-della-vedova, line 55: Nit, use the same format as the other dependencies above

Cargo.toml
```diff
@@ -115,6 +115,7 @@ bevy_ui = { git = "https://github.com/xiyuoh/bevy", branch = "fix-specialize-pan
 bevy_utils = { git = "https://github.com/xiyuoh/bevy", branch = "fix-specialize-panic" }
 bevy_window = { git = "https://github.com/xiyuoh/bevy", branch = "fix-specialize-panic" }
 bevy_winit = { git = "https://github.com/xiyuoh/bevy", branch = "fix-specialize-panic" }
+bevy_infinite_grid = { git = "https://github.com/althea28/bevy_infinite_grid", branch = "main" }
```
luca-della-vedova, line 118: From what I see your fork is up to date with the upstream main, why do we need to rely on this fork? It also seems that the crate is [released on crates.io](https://crates.io/crates/bevy_infinite_grid) and we should be able to just specify its version such as `bevy_infinite_grid = "0.15"`

luca-della-vedova (changes requested):

crates/rmf_site_editor/src/shapes.rs
althea28: Ah, I think I might have mixed up some things when I went back to some commits to fix the DCO check, sorry about this! I'll port over the relevant changes.

crates/rmf_site_editor/src/interaction/assets.rs
althea28: Noted! My mistake for missing out on this.

crates/rmf_site_editor/src/interaction/assets.rs
althea28: The updated mesh for the arrows are not symmetrical across all the axes hence 2 out of 4 of the arrows for each anchor need to be rotated once more. Noted on the suggestion, I'll use a single constructor instead!

crates/rmf_site_editor/Cargo.toml
althea28: Noted, sorry for missing this! I'll make the changes.

Cargo.toml
althea28: My bad on this, I was trying out some personal edits to this mod that I ultimately decided was not very useful. Sorry about that, I'll use the version from crates.io instead!

luca-della-vedova (approved):
The new arrows are very neat :sparkles:

althea28:
> Style is failing, can you run `cargo fmt` from the root of the repo?

Noted on this! I've run cargo fmt, which hopefully solves the style issue. Also, just an FYI that I've flattened the bottom of the arrow mesh for better visualisation following some discussion with Grey yesterday, which also required the arrows to be rotated more.

crates/rmf_site_mesh/src/primitives/arrow.rs
```diff
@@ -113,3 +113,48 @@ pub fn flat_arrow_mesh_between(
...
+pub fn make_triangular_arrow() -> MeshBuffer {
+    let base_positions = vec![
+        [0.0, 0.0, 1.0],  // 0
+        [0.5, 0.0, 0.0],  // 1
+        [-0.5, 0.0, 0.0], // 2
+        [0.0, 0.0, 0.2],  // 3
+        [0.0, 0.3, 0.4],  // 4
+    ];
+
+    let base_indices: Vec<u32> = vec![0, 1, 4, 0, 4, 2, 3, 4, 1, 3, 2, 4, 0, 3, 1, 0, 2, 3];
```
luca-della-vedova, line 126: I haven't tried it out but if you declared those as `Vec<usize>` instead of `Vec<u32>`, would that allow us to get rid of all `as usize` casts?

luca-della-vedova (approved):

PR #349, opened 2025-07-28
Add billboards with separate hover mesh

althea28:
This PR aims to add 2D billboards to locations.
- An external Bevy mod has been used to render the 2D billboards and corresponding texts
- An additional mesh with no material applied is spawned above the billboards for hovering interactivity, where the billboard text visibility is toggled when the user hovers over each billboard. 

An additional mesh is used for hovering interactivity as:
- The external billboard mod only shifts the vertices in the shader, and the actual billboard mesh does not move with the rendered texture. The billboard mesh hence currently cannot be used for hovering interactivity .
- There is a possible slight lag if we spawn our own billboard meshes and shift the location of the billboard according to the camera's view for every frame. Using the external mod, which only shifts the vertices in the shader, may be smoother. However, the extent of this possible lag is to be determined with a later PR which will attempt to shift the location of the billboard for every frame.


crates/rmf_site_editor/src/site/assets.rs
```diff
@@ -32,13 +32,29 @@ pub(crate) fn add_site_icons(app: &mut App) {
...
+        embedded_asset!(app, "src/", "icons/holdingBillboard.png");
+    }
+
+    #[cfg(all(target_family = "windows", not(target_env = "gnu")))]
+    {
+        embedded_asset!(app, "src\\", "icons\\baseBillboard.png");
+        embedded_asset!(app, "src\\", "icons\\chargingBillboard.png");
+        embedded_asset!(app, "src\\", "icons\\parkingBillboard.png");
+        embedded_asset!(app, "src\\", "icons\\holdingBillboard.png");
+    }
```
luca-della-vedova, line 57: Note this shouldn't be needed anymore (check https://github.com/open-rmf/rmf_site/pull/357)

althea28:
Closing in favor of another approach for adding billboards. I'll create a PR for that new approach once some bugs are fixed.

PR #350, opened 2025-07-29
Add support for mutex groups in lanes

luca-della-vedova:
### Implemented feature

A fairly rudimentary implementation for mutex groups that uses `Affiliation` for `Lane` and `Location`s to attach them to a mutex group.
This should bring us almost to feature parity with navigation properties with legacy traffic editor (I updated the TODO to denote that we are just missing the merge radius, imho we really shouldn't bother with `demo_mock_[...]` properties).

Most of the implementation is from existing bits and pieces of code. We should probably make `Affiliation` generic in followups and allow entities to have multiple types of affiliation, otherwise (for example) Lanes are affiliated to both nav graphs and mutexes through similar but just somewhat different approaches, and now that we added a mutex group affiliation lanes and locations won't be able to have any other sort of affiliation.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/crates/rmf_site_editor/src/widgets/workspace.rs
+++ b/crates/rmf_site_editor/src/widgets/workspace.rs
@@ -16,7 +16,7 @@
 */
 
 use crate::widgets::{FileMenu, MenuDisabled, MenuEvent, MenuItem, TextMenuItem};
-use crate::{AppState, CreateNewWorkspace, WorkspaceLoader, WorkspaceSaver};
+use crate::{AppState, WorkspaceSaver};
 use bevy::{ecs::hierarchy::ChildOf, prelude::*};
 
 #[derive(Default)]
@@ -33,19 +33,13 @@ impl Plugin for WorkspaceMenuPlugin {
 
 #[derive(Resource)]
 pub struct WorkspaceMenu {
-    new: Entity,
     save: Entity,
     save_as: Entity,
-    load: Entity,
 }
 
 impl FromWorld for WorkspaceMenu {
     fn from_world(world: &mut World) -> Self {
         let file_menu = world.resource::<FileMenu>().get();
-        let new = world
-            .spawn(MenuItem::Text(TextMenuItem::new("New").shortcut("Ctrl-N")))
-            .insert(ChildOf(file_menu))
-            .id();
         let save = world
             .spawn(MenuItem::Text(TextMenuItem::new("Save").shortcut("Ctrl-S")))
             .insert(ChildOf(file_menu))
@@ -56,22 +50,13 @@ impl FromWorld for WorkspaceMenu {
             ))
             .insert(ChildOf(file_menu))
             .id();
-        let load = world
-            .spawn(MenuItem::Text(TextMenuItem::new("Open").shortcut("Ctrl-O")))
-            .insert(ChildOf(file_menu))
-            .id();
 
         // Saving is not enabled in wasm
         if cfg!(target_arch = "wasm32") {
             world.entity_mut(save).insert(MenuDisabled);
             world.entity_mut(save_as).insert(MenuDisabled);
         }
-        Self {
-            new,
-            save,
-            save_as,
-            load,
-        }
+        Self { save, save_as }
     }
 }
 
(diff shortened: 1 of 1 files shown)


xiyuoh:
Thanks for bringing mutexes into the site editor! This is definitely an important step for legacy editor replacement.

I'm playing around with it and I notice (also from your PR description) that mutex groups can only be added to `Lane` and `Location`, but not `Anchor`. Given the nature of how we use mutex groups, I was expecting to be able to add them to any `Anchor`, or just those that are connected to a `Lane`, otherwise users would have to create several `Location`s for the sake of assigning mutexes to waypoints. From experience there were cases where we had to add mutex groups to many intermediate waypoints to help manage traffic in complicated spaces.

I'm wondering if it's possible to add mutexes as an `Affiliation<T>` to `AnchorBundle`? After a very brief scan I don't see any conflicts, and we can move towards using `Relationship` or other ways to generalize the concept of `Affiliation` in the future. Let me know what you think.

xiyuoh:
Just spitting out ideas: what if we only assign mutex groups to `Lane` when serializing? And then further specify which parts of the lane (start/end anchor and/or body) should the respective mutexes be added, making use of the referenced [anchors](https://github.com/open-rmf/rmf_site/blob/main/crates/rmf_site_format/src/lane.rs#L28). In the inspector we're already able to view and interact with the anchors associated with a lane so UI-wise we can probably work from there.

A very rough representation of what I'm suggesting:
```rust
pub struct LaneMutex<T: RefTrait> {
    pub start: Affiliation<T>,
    pub end: Affiliation<T>,
    pub body: Affiliation<T>,
}

pub struct Lane<T: RefTrait> {
    ...
    /// The mutexes that might be affiliated to this lane
    pub mutex: LaneMutex<T>,
    ...
}
```
This might require a couple extra steps to insert those mutex `Affiliation` components into `Anchor` entities during load, to ensure that each anchor is only affiliated with one mutex group.

In general I think this also makes it easier for users to visualize what they have assigned lane-by-lane instead of having to click around on each element. We wouldn't need mutex support for `Location` anymore too. But a drawback is having duplicate data serialized when a mutexed anchor is connected to multiple lanes.

If the cons aren't worth it we can probably just go ahead with the original implementation and address the `Location` inconvenience in a separate PR next time.

luca-della-vedova:
Ah that's also a very interesting idea! I think my only concern about it is that let's say we have an anchor that has 5 lanes that go into it with a mutex:

* At the site editor level it's just a mutex associated to a shared anchor, all good here
* When we serialize we will duplicate the data 5 times
* When we deserialize we will read the data also 5 times, with a potential error surface if one of those disagrees with the others.

I'm starting to also think about alternatives, in your experience is it common that lanes would have a certain mutex but their endpoints wouldn't?
Let's say there is a lane where its start has `mutex_a`, the lane `mutex_b` and its end `mutex_c`.
If a robot currently stays in `mutex_a` and wants to go to the end of the lane, would it need to claim both `mutex_b` and `mutex_c`? Or can it get started moving on the lane by only claiming `mutex_b` but will stop short of reaching its endpoint until it can claim `mutex_c` as well?
I am really no expert in the mutex design but I feel that if you _must_ always claim both the lane mutex and its endpoint mutexes there isn't really too much value in splitting them and you could automatically infer a mutex in a vertex if a lane connecting to it has one.
Now of course this ends up in the question of "what if multiple lanes with different mutexes go to the same endpoint"? And the answer is... I don't know... Another alternative could be that if a robot will always need to claim both the lane mutex and the endpoint mutex to navigate through a lane we could maybe get away with only supporting lane mutexes? But again you know more than I do in how mutexes are actually used in deployments

xiyuoh:
Yeah I agree, your concerns are very valid and those situations are far from ideal. You're also right in inferring that if a robot wants to travel across a single lane with 3 different mutexes on each anchor/body, it would have to obtain all 3 of them before it even begins moving.

I think these guidelines can be applied to most cases for robot movement across a single lane `A --[body B]-- C`:
- If a robot is moving away from endpoint `A` to `C` **into a mutex group** `mutex_1`, then `mutex_1` would be assigned to just `C`. We don't have to assign it to `B` as the robot shouldn't begin travelling across the lane unless it has already acquired all mutexes required.
- If a robot is moving from endpoint `A` to `C` **within the same group** `mutex_1`, then we would apply `mutex_1` to `A`, `B` and `C`. In fact, theoretically we wouldn't even need to assign anything to `B`.
- If a robot is moving away from endpoint `A` to `C` **out of a mutex group** `mutex_1`, then `mutex_1` would be assigned to `A` and `B`. Applying mutex to lane body `B` ensures that other robots don't start moving towards `A` before the previous robot has completely gotten out of the way.

More complex combinations such as having 3 distinct mutexes on a single lane are possible, e.g. criss-cross lanes (example [illustration](https://osrf.github.io/ros2multirobotbook/integration_nav-maps-strategies.html#example-maximum-occupancy-in-two-separate-rooms)), but it really depends on what the environment demands. So I guess what I'm saying is anything is possible and it's unlikely that we can get away with generalizing mutex assignment and infer them from just lane bodies 😢

An ugly alternative I can think of is to throw all member SiteIDs (`Anchor` or `Lane`) of each mutex group into `MutexGroup` and serialize them there instead, then we can have any entities added to a mutex group without duplicates. It would also make our files unnecessarily long.

Anyways, I'll start on reviewing this PR and brainstorm more in parallel.

luca-della-vedova:
Since how to do it for anchors is still a bit unclear I rescoped this PR to only add the support for lanes https://github.com/open-rmf/rmf_site/pull/350/commits/be2e9f187a0e1cec92b01c8d77a8497b9b650f08

PR #353, opened 2025-08-01
Support multiple paths in GZ_SIM_RESOURCE_PATH

xiyuoh:
The current asset loading searches for models within the `GZ_SIM_RESOURCE_PATH`, but `path.exists()` validates only a single filepath. This PR splits up `GZ_SIM_RESOURCE_PATH` in cases of multiple paths and enables the site editor to search for models in each of them.

### DIFF
--- a/crates/rmf_site_editor/src/site/inclusion.rs
+++ b/crates/rmf_site_editor/src/site/inclusion.rs
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2025 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


crates/rmf_site_editor/src/site_asset_io.rs
```diff
@@ -160,13 +161,21 @@ async fn fetch_asset<'a>(
     Ok(Box::new(VecReader::new(bytes)))
 }
 
-fn get_path_from_env() -> Result<PathBuf, env::VarError> {
+fn get_path_from_env() -> Result<Vec<PathBuf>, env::VarError> {
     let var = env::var(MODEL_ENVIRONMENT_VARIABLE)?;
-    let path = PathBuf::from(var);
+    let split: SmallVec<[&str; 8]> = var.split(':').collect();
```
mxgrey, line 166: We're going to run into some issues with this in Windows since path names tend to start with `C:\\...`.

What we should really use here is [`std::env::split_paths`](https://doc.rust-lang.org/std/env/fn.split_paths.html).

mxgrey (changes requested):
Great catch, but there's one tweak we for the implementation.

Yadunund (approved):
I can confirm this PR allows me to load assets located at different paths appended to `GZ_SIM_RESOURCE_PATH`.

mxgrey (approved):

PR #360, opened 2025-08-04
Fix site file saving

### LINKED ISSUE #352: Headless export error
Bumped into the following error when trying to run `--export-sdf` or `--export-nav`:

```
Encountered an error in system `librmf_site_editor::site::task::handle_task_edit`: Parameter `EventReader<EditModeEvent>::events` failed validation: Event not initialized
```

despite the event being added [here](https://github.com/open-rmf/rmf_site/blob/main/crates/rmf_site_editor/src/widgets/tasks/mod.rs#L66). Export works fine via the UI.

mxgrey:
#325 accidentally introduced a bug in the site file saving implementation that blocks virtually any site from being saved.

This PR fixes the bug and also introduces a unit test to do a round-trip load and save of a site containing a broad variety of elements. This should help catch any future bugs that might impede loading, saving, or headless exporting. In fact this also fixes #352.

Most of the lines of code changed in this PR are just models being added. The important changes are mostly in the `save.rs` module.


mxgrey:
There's been a failure happening in CI which isn't occurring on my local machine and which I don't have an explanation for. I'm attempting to debug but each debug cycle takes ~30 minutes :cry:

crates/rmf_site_editor/src/site/save.rs
```diff
@@ -1725,3 +1763,149 @@ pub fn save_site(world: &mut World) {
...
+        }
+
+        #[cfg(target_os = "windows")]
+        {
+            use std::os::windows::prelude::*;
+            // Windows uses different characters to represent newlines and path
+            // separators, so we cannot do a 1-to-1 comparison between the original
+            // and the generated file. We will simply check whether the new file
+            // is within 10% the size of the original file size since the differences
+            // incurred by these format changes should not be too significant.
```
luca-della-vedova, line 1854: I wonder, if we parsed both original and saved file into a string and:

* Did a search and replace that does "\r\n" -> "", followed by one that does "\n" -> "" for newlines and
* Did a search and replaces that does "\" -> "/" for path separators (a bit uglier)

Would that capture all cases and still allow us to have a more meaningful test? I'm not particularly worried about CI since we run both Windows and Linux and if something is wrong in the saving logic the Linux CI will probably catch it, it's more for local development for Windows users that they might not catch it.

PR #362, opened 2025-08-07
Remove dependency on custom bevy

luca-della-vedova:
It's been too long, the [upstream PR that we need](https://github.com/bevyengine/bevy/pull/19064) seems to be semi-abandoned, this workaround also forces downstream users to rely on the same custom bevy version.

Just add a workaround in the code instead. Specifically, the "guilty" bevy system is scheduled to run [after](https://github.com/bevyengine/bevy/pull/19064/files#diff-5db119571c62eac9218400a8191eee2417503cd13d3921cd711e5a7fecf8f936L291) `AssetEvents`, if we schedule all our systems to run _before_ `AssetEvents`, we should be able to get all our changes in. Furthermore, because of the automatically inserted command flush, all entities should be spawned before the system that used to panic is run.

### Test it!

I only tested it with lane spawning that seems to be fine, if you comment the only change in the code you should see that it crashes on lane spawning without it and it works fine with.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -517,8 +517,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bevy"
-version = "0.16.0"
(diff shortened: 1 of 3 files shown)


xiyuoh (approved):
LGTM, thanks!

PR #363, opened 2025-08-11
Support saving/loading downstream extension data, and loading assets serialized in-memory

xiyuoh:
This PR aims to support a feature that enables meshes to be received and loaded into the site editor via ROS 2 topics and service calls.


crates/rmf_site_format/src/asset_source.rs
```diff
@@ -33,6 +33,7 @@ pub enum AssetSource {
     Remote(String),
     Search(String),
     Package(String),
+    Ros(String),
```
mxgrey, line 36: I wonder if there may be a better name for this asset type than `Ros`. There's nothing in its implementation or how it's used that associates it with ROS. If anything it looks like it's for loading in-memory serialized models that need to be parsed.

Maybe we should call this `Memory`?

Issue #364, opened 2025-08-11
[Other issue]: Consider re-exporting bevy, bevy_impulse and clap as public modules


JohnTGZ:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### Description

Given that `rmf_site` could be used as a direct dependency in other projects, and that the main menu could be extended to accomodate that project's needs, should we also export `bevy`, `bevy_impulse` and `clap` as public modules? Given that we cannot mix and match different version of `bevy`, `bevy_impulse` and `rmf_site`, this might be useful for reducing incompatibility issues downstream. As for `clap`, I am not exactly sure what should be done. Right now, all these 3 modules need to be included as direct dependencies for other projects.

JohnTGZ:
Hi Grey, thanks for the comments, I will reply to your points below.

> Re-exporting `bevy` and `bevy_impulse` does make sense to me. Those dependencies are entangled with the public API, so I don't see any drawbacks.

Great, shall I go ahead with a PR then to re-export them as public?

> `clap` I'm less sure about. I expect downstream libraries to use clap on their own to define whatever custom arguments they need for their CLI tool. It's not obvious to me how or why `clap` would be part of the site editor's public API, but I could be overlooking something. If you can show an example of where `clap` would leak into the public API then I suppose we can consider it.

I agree with you on this. Now that I think about it, command line parsing is independent of how we would use the libraries from `rmf_site`.  Would be better for users to specify `clap` as a direct dependency.

mxgrey:
> Great, shall I go ahead with a PR then to re-export them as public?

Please go right ahead!

PR #366, opened 2025-08-11
Re-export bevy and rmf_site_animate modules as public

JohnTGZ:
This PR fixes issue #364 which re-exports `bevy`, `bevy_impulse` and `rmf_site_animate` modules as public ones so that downstream projects do not have to specify them separately as direct dependencies and can treat them as transitive ones.


mxgrey:
I don't see `bevy_impulse` included in the public exports.

JohnTGZ:
> I don't see `bevy_impulse` included in the public exports.

Thanks for the catch. I have made the changes and tested successfully.

luca-della-vedova (approved):

luca-della-vedova:
Merging this first, we following https://github.com/open-rmf/rmf_site/pull/366#discussion_r2269318586 we should look into re-exporting all the public dependencies as well

PR #368, opened 2025-08-13
Add camera panning movement to double-clicked object

althea28:
This PR aims to add camera panning to a double-clicked object. When a user double clicks an object in the 3D world, or double clicks the 'select' button for a model in the UI, the camera rotates to the selected model.

Design considerations:
- The global transform and transform translation value for some models return a value that is very close to the world centre (X and Y coordinates are both 0), though the model is not instantiated there. To get the actual centre of these models, the anchors of that model have to be found and the average has to calculated to find the target position for the camera to pan to. The current models I have identified for this are the lanes, walls, floors and measurements.
- The systems to pan the camera have been added within the rmf_site_editor crate. If there is a more appropriate place to add these systems, please let me know! I considered adding it under the rmf_site_camera crate, but it seems that I would need to add a dependency to rmf_site_picking. If this is preferable, I can move it here instead.

### DIFF
--- a/crates/rmf_site_editor/src/site/mod.rs
+++ b/crates/rmf_site_editor/src/site/mod.rs
@@ -447,8 +447,6 @@ impl Plugin for SitePlugin {
                 update_drawing_rank,
                 add_physical_camera_visuals,
                 check_selected_is_included,
-                check_double_click_event,
-                focus_camera_on_double_clicked_object,
                 check_for_missing_root_modifiers::<InstanceMarker>,
                 update_default_scenario,
             )
--- a/crates/rmf_site_editor/src/site/mod.rs
+++ b/crates/rmf_site_editor/src/site/mod.rs
@@ -447,6 +447,8 @@ impl Plugin for SitePlugin {
                 update_drawing_rank,
                 add_physical_camera_visuals,
                 check_selected_is_included,
+                check_double_click_event,
+                focus_camera_on_double_clicked_object,
                 check_for_missing_root_modifiers::<InstanceMarker>,
                 update_default_scenario,
             )
--- a/crates/rmf_site_editor/src/site/physical_camera.rs
+++ b/crates/rmf_site_editor/src/site/physical_camera.rs
@@ -17,8 +17,10 @@
 
 use crate::site::*;
 use bevy::{core_pipeline::tonemapping::Tonemapping, prelude::*};
+use rmf_site_camera::{active_camera_maybe, ActiveCameraQuery};
 use rmf_site_format::{PhysicalCameraProperties, Pose};
-use rmf_site_picking::Selectable;
+use rmf_site_picking::{DoubleClickSelection, Select, Selectable, Selection};
+use std::time::Instant;
 
 pub fn add_physical_camera_visuals(
     mut commands: Commands,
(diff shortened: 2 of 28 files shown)


crates/rmf_site_editor/src/site/mod.rs
```diff
@@ -447,6 +447,8 @@ impl Plugin for SitePlugin {
                 update_drawing_rank,
                 add_physical_camera_visuals,
                 check_selected_is_included,
+                check_double_click_event,
+                focus_camera_on_double_clicked_object,
```
luca-della-vedova, line 451: I suspect these should be in a mix of `rmf_site_camera` and `rmf_site_picking`, rather than in the site editor itself.

Specifically, `rmf_site_picking` takes care of selection so it would be the more relevant crate to put double clicking detection into, but it doesn't have knowledge of the camera itself.
On the other hand, `rmf_site_camera` has all the camera logic but no selection logic.

My suggestion (but I'm open to alternatives), would be to put:

* A system that detects double clicks on entities and sends an event when an `Entity` is double clicked (roughly similar in content to your `DoubleClickSelection` resource, except it would only need to contain the event that was double clicked). This system can reside in `rmf_site_picking` and would be mostly similar too `check_double_click` event, with the exception that instead of setting a resource it would send an event.
* A system that reads the event and does the camera animation. This would be similar to your second system, but would read an event instead and focus on the entity. This can reside in the `rmf_site_camera` crate.

This should also allow us to remove the use of global resources. If you have some data that only a single system needs you can put it in a `Local<Struct>` rather than a `Res<Struct>`, this would make the data persistent and local to the system that needs it.
The double clicking detection system could then keep only locally relevant information (the last clicked time, the last clicked entity) in its internal `Local` state.

crates/rmf_site_editor/src/site/physical_camera.rs
```diff
@@ -17,8 +17,10 @@
 
 use crate::site::*;
 use bevy::{core_pipeline::tonemapping::Tonemapping, prelude::*};
+use rmf_site_camera::{active_camera_maybe, ActiveCameraQuery};
 use rmf_site_format::{PhysicalCameraProperties, Pose};
-use rmf_site_picking::Selectable;
+use rmf_site_picking::{DoubleClickSelection, Select, Selectable, Selection};
+use std::time::Instant;
```
luca-della-vedova, line 23: Related to the above, `physical_camera.rs` has implementation of actual cameras in the site, not the "view camera", i.e. this:

image

crates/rmf_site_editor/src/site/physical_camera.rs
```diff
@@ -59,3 +61,125 @@ pub fn add_physical_camera_visuals(
...
+
+    for _ in select.read() {
+        if let (Some(last_entity), Some(last_time)) = (
+            double_clicked.last_selected_entity,
+            double_clicked.last_selected_time,
+        ) {
+            let elapsed_time = current_time.duration_since(last_time).as_millis();
+
+            if let Some(selected_entity) = selection.0 {
+                if last_entity == selected_entity && elapsed_time < 500 {
```
luca-della-vedova, line 80: Nit, should we parametrize this? It can be a global constant for now

crates/rmf_site_editor/src/site/physical_camera.rs
```diff
@@ -59,3 +61,125 @@ pub fn add_physical_camera_visuals(
...
+    floors: Query<&Path<Entity>, With<FloorMarker>>,
+    measurements: Query<&Edge<Entity>, With<MeasurementMarker>>,
+    mut double_clicking: ResMut<DoubleClickSelection>,
+) {
+    if double_clicking.entity.is_none() {
+        return;
+    }
+
+    let mut target_position;
+    if let Some(double_clicked) = double_clicking.entity {
```
luca-della-vedova, line 105: The returning if the entity is none, plus the if let below (with an else branch that returns) are effectively the same thing. I would suggest refactoring every

```
if let Some([...]) = [...] {
} else { 
   [...]
   return; 
}
```

Into a

```
let Some([...]) = [...] else {
   [...]
   return;
};
```
This will reduce indentation and make sure we don't forget an else statement

crates/rmf_site_editor/src/site/mod.rs
mxgrey: I think these are all good ideas and that this is the right direction, but I would just tweak it slightly:

* `DoubleClickSelection` should instead be an event called `DoubleClickSelect` in `rmf_site_picking` (just like you're suggesting)
* `PanToElement` should be a resource in `rmf_site_camera` that looks something like:

```rust
pub struct PanToElement {
    pub target: Option<Entity>,
    pub interruptible: bool,
    pub persistent: bool,
}
```

When `target` is set to Some, the animation pans the camera towards the target. When `interruptible` is set to true, any user input for controlling the camera should reset the `target` to `None`. When `persistent` is set to false, the `target` should be reset to `None` once the camera finishes panning to the target. A true value for `persistent` would allow this to support animated moving targets.

crates/rmf_site_picking/src/select/resources.rs
```diff
@@ -15,11 +15,19 @@ pub struct Selection(pub Option<Entity>);
 pub struct Hovering(pub Option<Entity>);
 
 /// Used as a resource to keep track of which entity is currently double clicked.
-#[derive(Default, Debug, Clone, Copy, Resource)]
+#[derive(Debug, Clone, Copy, Resource)]
 pub struct DoubleClickSelection {
```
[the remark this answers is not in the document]
althea28: Noted on this! I've removed the `init_resource` call and `Resource` derive. 

However, from my understanding, it seems that `DoubleClickSelection` has to be a pub struct, as the compiler needs to see the full signature of the `send_double_click_event` system that uses `DoubleClickSelection`, when the system is added in `plugin.rs`. Can I check if this is what you meant?

luca-della-vedova:
Thanks for iterating! LGTM

PR #369, opened 2025-08-18
added layers.rs to fix z-fighting issue

### LINKED ISSUE #343: z-fighting for planar renderings
Some items that are rendered on the ground plane have been experiencing z-fighting after migrating to newer versions of bevy. For example here the visual cue for the door is fighting with the floor:

https://github.com/user-attachments/assets/2e23cdcf-2309-4feb-992c-dbd432ee7222

In the past the z-fighting of ground items was minimized by automatically tweaking the height of objects using the [`RecencyRanking`](https://github.com/open-rmf/rmf_site/blob/main/crates/rmf_site_editor/src/recency.rs) components and having specific height ranges for certain kinds of visuals.

Some change to bevy's rendering pipeline has made this less effective, perhaps something related to numerical stability or floating point precision issues. We should consider other options for resolving the z-fighting, such as using [`depth_bias`](https://docs.rs/bevy/latest/bevy/pbr/struct.MaterialProperties.html#structfield.depth_bias).

Nielsencu:
Added layers.rs file to aggregate all z-values used for placing objects. To fix #343 

image

### DIFF
--- a/crates/rmf_site_editor/src/site/drawing.rs
+++ b/crates/rmf_site_editor/src/site/drawing.rs
@@ -19,8 +19,8 @@ use crate::{
     site::{
         get_current_workspace_path, Anchor, DefaultFile, FiducialMarker, GlobalDrawingVisibility,
         LayerVisibility, MeasurementMarker, MeasurementSegment, RecencyRank,
-        DEFAULT_MEASUREMENT_OFFSET, FLOOR_LAYER_START,
     },
+    layers,
     CurrentWorkspace,
 };
 use bevy::{asset::LoadState, ecs::hierarchy::ChildOf, math::Affine3A, prelude::*};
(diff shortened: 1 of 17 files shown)


crates/rmf_site_editor/src/site/drawing.rs
```diff
@@ -249,7 +249,7 @@ pub fn update_drawing_rank(
                 if let Ok(segment) = measurements.get(*child) {
                     transforms
                         .get_mut(**segment)
-                        .map(|mut tf| tf.translation.z = z + DEFAULT_MEASUREMENT_OFFSET)
+                        .map(|mut tf| tf.translation.z = z + layers::ZLayer::get_offset())
```
luca-della-vedova, line 252: This looks a bit suspicious.
If I understand the intention of the original code correctly, it tries to offset a measurement to be above the drawing by a fixed value.
Can we just do that by adding `ZLayer::Measurement.to_z() - ZLayer::Drawing.to_z()`? Note that in your enum definition the distance between drawings and measurement is "2 offsets" since they are two enum variants away, this would make it only "one offset" that would actually cause Z fighting with floors.

I also lowkey wonder if this call was the only reason `get_offset()` is public and whether we should just make it private instead.

luca-della-vedova (changes requested):
Had a look at the code, mostly minor feedback but one (`unsafe` usage) is more substantial

PR #370, opened 2025-08-19
Add lane arrow shader for overlapping arrows

althea28:
This PR aims to add a lane arrow shader for better visualisation of lane directions. The direction and speed of the arrows should reflect the speed limit and Reverse Motion options that the user selects in the Inspector.

The shader shows overlapping arrows in different directions as per below. This PR is to be compared with another [PR #371](https://github.com/open-rmf/rmf_site/pull/371) which show the arrows in adjacent lanes instead. The implementation for the lane shader in both PRs are the same, the only difference is the shader file code, and the color of the arrows moving in the reverse direction.

https://github.com/user-attachments/assets/b2bd4c0d-9b48-488d-ad7e-fb2b99da6dd4

Some points noted:
- There weren't any shaders in the site editor that I could refer to, please let me know if the way I imported it is alright!
- The bidirectional boolean value has to be passed into the shader as a u32 as WGSL cannot take in bool values.
- When instantiating new lane visuals under add_lane_visuals, the choice has been made to directly instantiate a new ExtendedMaterial, instead of cloning from the Assets. 

```
        let mid = commands
            .spawn((
                Mesh3d(assets.lane_mid_mesh.clone()),
                MeshMaterial3d(extended_materials.add(ExtendedMaterial {
                    base: StandardMaterial {
                        depth_bias: 3.0,
                        ..default()
                    },
                    extension: assets::LaneArrowMaterial {
                        single_arrow_color: LANE_SINGLE_ARROW_COLOR.into(),
                        double_arrow_color: LANE_DOUBLE_ARROW_COLOR.into(),
                        background_color: LANE_BASE_COLOR.into(),
                        number_of_arrows: (start_anchor - end_anchor).length() / LANE_WIDTH,
                        forward_speed: forward_speed_limit,
                        backward_speed: backward_speed_limit,
                        bidirectional: is_bidirectional as u32,
                    },
                })),
                ...
```

As each lane material has its own unique values (e.g. forward_speed, bidirectional), doing the alternative and cloning the ExtendedMaterial from Assets, would result in the code looking like this:

```
        let lane_arrow_handle = assets.lane_arrow_material.clone();
        let Some(material) = extended_materials.get_mut(&lane_arrow_handle) else {
            return;
        };
        let mut material_instance = material.clone();

        material_instance.extension.number_of_arrows = (start_anchor - end_anchor).length() / LANE_WIDTH;
        material_instance.extension.forward_speed = forward_speed_limit;
        material_instance.extension.backward_speed = backward_speed_limit;
        material_instance.extension.bidirectional = is_bidirectional as u32;

        let mid = commands
            .spawn((
                Mesh3d(assets.lane_mid_mesh.clone()),
                MeshMaterial3d(extended_materials.add(material_instance)),
                ...
```

This involves 2 cloning operations, then adding the new material instance to extended_materials again. From my understanding, this seems to be more inefficient than directly instantiating a new ExtendedMaterial as per above. However, if this method is preferred, I can change it to this instead.

### DIFF
--- a/crates/rmf_site_editor/src/site/assets.rs
+++ b/crates/rmf_site_editor/src/site/assets.rs
@@ -19,14 +19,48 @@ use crate::site::*;
 use bevy::{
     asset::embedded_asset,
     math::{primitives, Affine3A},
+    pbr::MaterialExtension,
     prelude::*,
+    render::render_resource::{AsBindGroup, ShaderRef},
 };
 use rmf_site_mesh::*;
 
-pub(crate) fn add_site_icons(app: &mut App) {
+const LANE_SHADER_PATH: &str = "embedded://librmf_site_editor/site/shaders/lane_arrow_shader.wgsl";
+
+pub(crate) fn add_site_assets(app: &mut App) {
     embedded_asset!(app, "src/", "icons/battery.png");
     embedded_asset!(app, "src/", "icons/parking.png");
     embedded_asset!(app, "src/", "icons/stopwatch.png");
+
+    embedded_asset!(app, "src/", "shaders/lane_arrow_shader.wgsl");
+}
+
+#[derive(Asset, AsBindGroup, Reflect, Debug, Clone, Component)]
+pub struct LaneArrowMaterial {
+    #[uniform(100)]
+    pub single_arrow_color: LinearRgba,
+    #[uniform(101)]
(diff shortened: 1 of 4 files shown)


mxgrey:
Closing this PR in favor of https://github.com/open-rmf/rmf_site/pull/378

PR #371, opened 2025-08-19
Add lane arrow shader for adjacent arrows

althea28:
The aim of this PR is to compare the lane arrows visualisation with [PR #370](https://github.com/open-rmf/rmf_site/pull/370). The implementation of this lane arrow shader is the same as PR #370. The only difference between both PRs is the shader file code, and the color of the arrows moving in the reverse direction.

This shader shows arrows moving in different directions as adjacent lines as per below. This PR is to be compared with PR #370 that show overlapping arrows instead.

https://github.com/user-attachments/assets/75e82996-0ebb-428b-93ea-717e5c69408d

### DIFF
--- a/crates/rmf_site_editor/src/site/assets.rs
+++ b/crates/rmf_site_editor/src/site/assets.rs
@@ -19,14 +19,48 @@ use crate::site::*;
 use bevy::{
     asset::embedded_asset,
     math::{primitives, Affine3A},
+    pbr::MaterialExtension,
     prelude::*,
+    render::render_resource::{AsBindGroup, ShaderRef},
 };
 use rmf_site_mesh::*;
 
-pub(crate) fn add_site_icons(app: &mut App) {
+const LANE_SHADER_PATH: &str = "embedded://librmf_site_editor/site/shaders/lane_arrow_shader.wgsl";
+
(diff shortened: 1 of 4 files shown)


luca-della-vedova:
This has been stuck for some time because of the two different approaches. I propose, to solve it and move forward, can we just make the behavior a parameter (a constant defined in the shader is OK) so that we can have an approach first (perhaps this one?) but if we change our mind can easily change the appearance later?

My proposed parameters would be something like:

```
float LEFT_ARROW_WIDTH_RATIO = 0.5;
float RIGHT_ARROW_WIDTH_RATIO = 0.5;
```

Where the value is the percentage of width that the arrow takes, two values of 0.5 will show this behavior, two values of 1.0 will show the overlapping behavior, intermediate values (i.e. > 0.5 but < 1.0) will have partially overlapping behavior.
Then we can easily tune in the future if this show to be misleading

althea28:
Noted on the points above! @mxgrey created a branch earlier that edits a system to return the Nav Graph color. I've built on those changes and opened a new PR #378 based on the above discussions.

mxgrey:
Closing this PR in favor of #378

Issue #351, opened 2025-07-30
Misaligned outlines in drawing editor


xiyuoh:
When using the editor's 2D mode, the outlines generated for drawings are not aligned with the image's borders. It seems like the line transforms are taken from the last 3D view of the floorplan right before entering the `Edit drawing` mode.

<img width="1594" height="888" alt="Image" src="https://github.com/user-attachments/assets/fb2be8fb-556f-4044-ba2d-cdd9c603f220" />

The outlines are expected to fit nicely with the drawing border.

PR #374, opened 2025-08-22
Propagate is_active to camera children

xiyuoh:
This PR closes https://github.com/open-rmf/rmf_site/issues/351 to fix misaligned outlines when in 2D drawing mode. It turns out that the active camera being used to render outlines was wrong - `is_active` for children cameras for both Perspective and Orthographic modes weren't being updated.

### DIFF
--- a/crates/rmf_site_camera/src/systems.rs
+++ b/crates/rmf_site_camera/src/systems.rs
@@ -156,42 +156,66 @@ pub fn init_cameras(
 pub fn change_projection_mode(
     projection_mode: Res<ProjectionMode>,
     mut cameras: Query<(&mut Camera, &mut Visibility)>,
-    ortho_cams: Query<Entity, With<OrthographicCameraRoot>>,
-    persp_cams: Query<Entity, With<PerspectiveCameraRoot>>,
+    ortho_cams: Query<(Entity, &Children), With<OrthographicCameraRoot>>,
+    persp_cams: Query<(Entity, &Children), With<PerspectiveCameraRoot>>,
 ) {
     let projection_mode = *projection_mode;
 
     match projection_mode {
         ProjectionMode::Perspective => {
-            for ortho_cam in ortho_cams {
+            for (ortho_cam, children) in ortho_cams {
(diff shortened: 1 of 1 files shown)


crates/rmf_site_camera/src/systems.rs
```diff
@@ -156,42 +156,66 @@ pub fn init_cameras(
 pub fn change_projection_mode(
     projection_mode: Res<ProjectionMode>,
     mut cameras: Query<(&mut Camera, &mut Visibility)>,
-    ortho_cams: Query<Entity, With<OrthographicCameraRoot>>,
-    persp_cams: Query<Entity, With<PerspectiveCameraRoot>>,
+    ortho_cams: Query<(Entity, &Children), With<OrthographicCameraRoot>>,
+    persp_cams: Query<(Entity, &Children), With<PerspectiveCameraRoot>>,
```
luca-della-vedova, line 160: Why not just combine the `cameras` query component into the `ortho_cams` or `persp_cams`? This would also get rid of some of the `let else` statements below.
I suppose doing the naive way would panic because of conflicting access for mutable variables but you could use a disjoint query such as:

`ortho_cams: Query<(Entity, &Children, &mut Camera, &mut Visibility), (With<OrthographicCameraRoot>, Without<PerspectiveCameraRoot>)>`

And viceversa for the perspective cameras

xiyuoh: I might be understanding you wrongly, but `OrthographicCameraRoot` and `PerspectiveCameraRoot` are only inserted in the base cameras, so we wouldn't be able to query for the `Camera` of their children.

mxgrey: There are several ways we could consider tweaking this implementation, like we could introduce `OrthographicUserCamera` and `PerspectiveUserCamera` components which we apply to to their respective camera entities, and then we can have a query that filters on those components instead of iterating through children.

But I think this fix is important enough that I don't want to spend longer than necessary deliberating. We can always follow up with a more elegant revision later if we decide to do a cleanup of the codebase.

mxgrey (approved):
Thanks for tracking down this tricky bug @xiyuoh

PR #378, opened 2025-08-31
Animated arrows to show robot motion along lanes

althea28:
This PR aims to add lane hovering behaviour, as well as make the lane arrow widths adjustable.
image

Current behaviour:
- Lane arrows of a lane only move when the user hovers over or selects that lane.
- If the lane is bidirectional and the user is not hovering over or has not selected that lane, no arrows are rendered. If the lane is unidirectional and the user is not hovering over or has not selected that lane, the arrows will be rendered on the lane, but will not be moving.
- The color of the forward and backward arrows are calculated from the color of the Nav Graph of that lane, and will change accordingly with the current lane color.

Lane arrows can be adjustable within the shader `lane_arrow_shader.wgsl` itself
```
        let forward_arrow_width_ratio = 0.5;
        let backward_arrow_width_ratio = 0.5;
```
Adjusting these ratio values larger than 0.5 will show overlapping arrows.


mxgrey:
To simplify things, I've retargeted this PR at `main`. I'll close the other related PRs to consolidate everything.

mxgrey (approved):
This is a huge aesthetic and UX upgrade that users have wanted for a long time. I'm very excited that the site editor will finally have this feature!

And the hover+select behaviors introduced by this PR make it even better than expected!

We might still need to deliberate on how to express that a lane doesn't have a speed limit, but I think that can be done by the project committee in a follow-up effort.

Thanks for all the meticulous work on this, @althea28 !

PR #379, opened 2025-09-01
Add door positions, name and cues

althea28:
This PR aims to add positions, names and cues for doors, as well as add names for lifts.
image

Features added:
- Door cues have been updated to have a striped texture instead. These striped texture remain proportional and scale with the door length.
- Door positions have been added in the Door Inspector and can be adjusted using a slider between the value 0 (closed) and 1 (open). By default, all doors are displayed as open.
- Door names have been added to be shown on the door and floor. Floor names are included for user reference, as when sliding doors are fully open, the name on the door is not visible. Scaling logic has also been added such that the name scales with the door lengths.
- Lift names have also been added similarly, with scaling logic applied as well.


mxgrey (approved):
This is an excellent aesthetic and informational improvement! I made the following tweaks to clean up the UX:
* Lift and door labels that are on the floor will always appear above lanes, including when lanes are hovered
* Lift doors stopped working correctly because of an invalid query. This is fixed now.
* I've made lift text larger so it's easier to see lift names.
* Door text labels can now be selected
* Visual cues for the door motions no longer have an outline, instead their material is tweaked during interaction

PR #383, opened 2025-09-04
Fix billboard texture asset path

althea28:
This PR is just to resolve a very minor billboard texture path bug, which made the billboards not visible in the world! Probably happened when reorganising the textures into a new folder.


mxgrey (approved):
Whoops, it looks like I updated the [`embedded_asset`](https://github.com/open-rmf/rmf_site/pull/379/files#diff-e3f99c8709b52986b0c71d709c3cb4b3fb1877ce00c388da543b626d0df14ae3R37-R41) calls, but forgot that the paths are referred to twice. Thanks for catching this!

PR #384, opened 2025-09-05
Add site map server

luca-della-vedova:
### Implemented feature

Implement a map server based on `rmf_site_format`.

The server is a replacement for traffic editor's [building_map_server](https://github.com/open-rmf/rmf_traffic_editor/tree/main/rmf_building_map_tools/building_map_server). It still allows publishing legacy maps but allows publishing site files as well.
Note that there is a major caveat with adding this crate to this repo that we should figure out before merging it, specifically, it depends on `rclrs` (OK) and message packages (not OK).
Depending on message packages _right now_ has a few issues.

#### Buildfarm distribution

The messages distributed in the build farm don't (yet) include rust bindings. This might change in the future but in the current stage it will make the package fail to build unless it is in a colcon workspace that also contains the [rust generators](https://github.com/ros2-rust/rosidl_rust) as well as all the message packages that this server depends on (`rmf_building_map_msgs` and `service_msgs`).

#### Colcon building

Even if the above was to be solved, the way Rust packages are able to find ROS dependencies (in this case messages) is through a combination of [cargo-ament-build](https://github.com/ros2-rust/cargo-ament-build/) and [colcon-ros-cargo](https://github.com/colcon/colcon-ros-cargo). Specifically, the latter will generate a `.cargo/config.toml` file that points subsequent cargo builds to the messages found in the ament prefix.
This means that, unless we reimplement some of this logic manually (i.e. in a `build.rs` file), the crate will only be buildable through `colcon`, and not through `cargo` anymore. Or rather to be precise, the _first_ build will have to be through `colcon`, which will generate the `config.toml` file, then subsequent builds can be through `cargo build`.

#### Unable to skip building workspace members

I would be OK with accepting the issues above if there was a way to skip building a workspace member based on a feature flag or something along those lines. For example, we could have a `build_map_server` feature, disable it by default, and make sure people can still build the site editor normally without having to go through the message generation + colcon trouble mentioned above.
However, I couldn't find a way to fully skip compilation of workspace members based on feature flags, happy to hear if there is any feedback.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "crates/*"
 ]
+default-members = ["crates/rmf_site_editor"]
 
 # This is needed for packages that are part of a workspace to use the 2nd version
 # of the cargo dependency resolving algorithm. In ordinary crates this would be
--- a/crates/rmf_site_map_server/Cargo.toml
+++ b/crates/rmf_site_map_server/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "rmf_site_map_server"
+version = "0.0.1"
+authors = ["Luca Della Vedova <lucadv@intrinsic.ai>"]
+edition = "2021"
+
+[dependencies]
+anyhow = {version = "1", features = ["backtrace"]}
+rmf_site_format = { workspace = true }
+
+[[bin]]
+name = "site_map_server"
+path = "src/main.rs"
+
+[dependencies.rclrs]
+version = "0.4"
+
+[dependencies.rosidl_runtime_rs]
+version = "0.4"
+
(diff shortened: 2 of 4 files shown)


mxgrey:
Thanks for getting this PR up. Considering the difficulties that currently exist around including `rclrs` and its message bindings into a colcon workspace, I think we should just introduce a new repo for this crate. Maybe if we reach a point where rosdep can sort out these dependencies, we could consider including it in the `rmf_site` repo.

luca-della-vedova:
Agreed, closing in favor of the [rmf_site_map_server](https://github.com/open-rmf/rmf_site_map_server) repo

Issue #375, opened 2025-08-26
[Bug]: Fail to load models in a large world


JohnTGZ:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

N/A

### ROS distribution

Jazzy

### ROS installation type

Binaries

### Description of the bug

When loading a `.site.ron` file with a large amount of models. Some of the models fail to load, presenting the following error for a few models:
```bash
2025-08-26T06:11:18.062241Z ERROR librmf_site_editor::site::model: Failed to execute model loading request for entity 11320v1#4294978616 and source Search("WorkcellEnclosure/enclosure_visual.glb") Reason: Asset did not contain a model
``` 

Loading a smaller world with these same files does not present the same problem.

### Steps to reproduce the bug

N/A

JohnTGZ:
From further discussions and some debugging, it seems like  [this line](https://github.com/open-rmf/rmf_site/blob/fff230d9009726897a1c7cb8391f8a33a16edec5/crates/rmf_site_editor/src/site/model.rs#L197) is assigning a `None` type to the GLTF variable, so it could very be possible that the GLTF has not arrived inside the Assets<Gltf> database by the time we try to get the handle of the specific GLTF id.

One possible fix is to have a system that periodically retries upon receiving a `None` value when fetching `Gltf` meshes  from Assets database.

PR #386, opened 2025-09-10
Retry spawning scenes if assets are not ready

xiyuoh:
This PR closes https://github.com/open-rmf/rmf_site/issues/375 where some models may not load successfully in large worlds. It turns out some Gltf assets have not arrived in the `Assets<Gltf>` database, so the current approach of failing the entire model loading workflow after one try would result in some models not being spawned, even if they are valid. Here we introduced a timeout loop and implemented a retry mechanism for these late-arriving assets.

Additionally I added some `skip_serializing_if = "is_default"` for our new door features, so that older site files can still be imported fine.

### DIFF
--- a/crates/rmf_site_editor/src/site/model.rs
+++ b/crates/rmf_site_editor/src/site/model.rs
@@ -321,20 +321,38 @@ fn handle_model_loading(
             ));
         };
         // Now we have a handle and a parent entity, call the spawn scene service
-        let res = channel
-            .query(
-                (request.parent, handle, request.source.clone()),
-                spawn_scene_for_loaded_model.into_blocking_callback(),
-            )
-            .await
-            .available()
-            .ok_or_else(|| {
-                ModelLoadingError::new(
-                    request.clone(),
-                    ModelLoadingErrorKind::WorkflowExecutionError,
+        // Keep it in a loop to ensure that we retrieve the assets even if they
+        // arrive late
+        let t_initial = std::time::Instant::now();
+        // TODO(@xiyuoh) make this configurable
+        let timeout = std::time::Duration::from_secs(5);
+        let spawn_scene = loop {
+            let res = channel
+                .query(
+                    (request.parent, handle.clone(), request.source.clone()),
+                    spawn_scene_for_loaded_model.into_blocking_callback(),
                 )
-            })?;
-        let Some((scene_entity, is_scene)) = res else {
+                .await
+                .available()
+                .ok_or_else(|| {
+                    ModelLoadingError::new(
+                        request.clone(),
+                        ModelLoadingErrorKind::WorkflowExecutionError,
+                    )
+                })?;
+            if let Some((scene_entity, is_scene)) = res {
+                break Some((scene_entity, is_scene));
+            }
+            let elapsed = std::time::Instant::now() - t_initial;
+            if timeout < elapsed {
+                info!(
+                    "Attempt to spawn scene for loaded model exceeded timeout '
+                    'of {timeout:?}: {elapsed:?}"
+                );
+                break None;
+            }
+        };
+        let Some((scene_entity, is_scene)) = spawn_scene else {
             return Err(ModelLoadingError::new(
                 request.clone(),
                 ModelLoadingErrorKind::NonModelAsset,
--- a/crates/rmf_site_format/src/door.rs
+++ b/crates/rmf_site_format/src/door.rs
@@ -109,6 +109,7 @@ impl Default for DoorType {
 pub struct SingleSlidingDoor {
     /// Which side the door slides towards
     pub towards: Side,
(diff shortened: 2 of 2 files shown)


xiyuoh:
@JohnTGZ Can't seem to add you as a reviewer but feel free to take a look!

xiyuoh:
Updated the PR to use workflows cycle to trigger retries instead of keeping the asset loading in a loop.

luca-della-vedova (dismissed):
I can't say I'm 100% convinced by this change, I also am not very excited about the hard coding of the `AssetSource::Memory` variant.

If I was to inspect this variant I would expect that it involves any asset that is currently loaded in memory. However, it seems to be [hardcoded](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L167) to only be a scene contained in a GLTF.
This exact `if let` is where I believe the root cause of this bug actually lies.

If you look at the [load_asset_source](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L155-L193) function, it tries to parse an asset path then it queries the asset server to [load the asset _asynchronously_ then awaits its result](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L177-L178).
Awaiting for the async load means that we are guaranteed that by the end of that function the asset will be loaded. Now the issue is that in the new branch for `Memory` assets we are using a [normal `load`](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L167) which will return immediately, without actually waiting for the asset to be loaded.

The `NonModelAsset` error is named in such a way because it happens if we try to load a model from a handle and fail to fetch a model. The assumption is that because the previous step in the workflow awaits for the load to be completed, an error at that step usually means that for some reason the loaded asset does not actually contain a model.
This assumption however fails if we call this function without making sure to await for model loading, in which case we will fail because the asset is just not loaded yet (I'm guessing in your case the early return happens [here](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L213)?).

So in a way the immediate fix is to change your `load` into `load_untyped_async` then awaiting like the other branch. However, the change I would really like to see is removing the special casing for `Memory` assets.

I'm not a big fan that `Memory` assets are hardcoded to be gltfs, are they always expected to be such? Could we just change the upstream code to append a `.glb` extension to them so the asset loader knows what to do automatically without having to hardcode a GLTF format [here](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L167)?
Ideally we should just be able to reuse the [conversion to string](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_editor/src/site/model.rs#L170) to convert the `AssetSource::Memory` variant into something that can be parsed by the `AssetServer`. The call is made by:

* [AssetSource::with_base_path](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_format/src/asset_source.rs#L105-L126) which already, correctly, does nothing unless the `AssetSource` is of the `Local` variant.
* [String::try_from<&AssetSource>](https://github.com/open-rmf/rmf_site/blob/0ec90fdd3b4344b7c27bb10b3514952fc74d2b1d/crates/rmf_site_format/src/asset_source.rs#L136-L149) that actually was introduced to remove a nasty source of crashes at the bevy level, and should still be applicable to `Memory` variants.

So again I'm not 100% sure how the `Memory` variant is meant to be used but if we removed the special casing I believe we would get rid of this bug altogether

xiyuoh:
@luca-della-vedova Sorry it took a while for me to get to this, the API is a great find! I tried applying your suggestion, it works great! Thanks for coming up with a better and more elegant approach. Made the changes in [cd763eb](https://github.com/open-rmf/rmf_site/pull/386/commits/cd763ebd73b958ac482363ce5e7b92072115f5e1).

mxgrey (approved):
This is a very nice solution. Thanks for all the iterating @xiyuoh, and thanks @luca-della-vedova for finding the right API to use here.

I'll merge this by EOD tomorrow if you don't have any other concerns @luca-della-vedova

luca-della-vedova (approved):

PR #387, opened 2025-09-16
Use web-time for wasm compatibility

mxgrey:
Some recent PRs added the use of `std::time::Instant::now` throughout the repo. Unfortunately that function does not have an implementation for `wasm32-unknown-unknown` targets and leads to a crash when trying to run the site editor in the web browser.

This PR replaces all uses of `Instant::now` with an implementation from the `web-time` crate which does provide an implementation for the `wasm32-unknown-unknown` target. This will allow `main` to be deployed on the web.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5720,6 +5720,7 @@ dependencies = [
  "urdf-rs",
  "utm",
  "uuid",
+ "web-time",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


mxgrey:
This PR has expanded to fixing a variety of issues related to wasm compatibility, including fixing the custom shader for lane arrow animations.

mxgrey:
Since there are no new features being introduced -- just fixes, mostly specific to wasm -- and we're short on reviewers at the moment due to scheduling conflicts, I'm going to move this PR forward without the usual review requirement so it doesn't get blocked up.

PR #389, opened 2025-09-22
Re-Introduces Occupancy Grid Export

arjo129:
### Implemented feature

Adds support to export an occupancy grid to a PNG so other external tools may consume the exported occupancy grid.

### GenAI Use

- [x] I did not use GenAI

Generated-by:


mxgrey:
The other PR has been merged. The conflicts for this PR are now visible.

arjo129:
Works on desktops. You can simply click `File`>`Export Occupancy` as shown in the video below. It is currently disabled on the web as we don't support IO there.

https://github.com/user-attachments/assets/7e3340cf-4b6b-4baa-ad96-3738d3ed6213

crates/rmf_site_editor/src/occupancy.rs
```diff
@@ -185,7 +198,50 @@ enum Group {
     None,
 }
 
-fn handle_calculate_grid_request(
+fn handle_export_request(
+    mut request: EventReader<ExportGridRequest>,
+    //occupancy_info: Res<OccupancyInfo>,
```
luca-della-vedova: Nit commented code

crates/rmf_site_editor/src/site/save.rs
```diff
@@ -1759,6 +1761,39 @@ pub fn generate_site(
     });
 }
 
+pub fn export_grid(world: &mut World, path: &PathBuf) {
+    let mut system_state: SystemState<(Query<(&Grid, &ChildOf)>, Query<&NameInSite>)> =
+        SystemState::new(world);
+    let (grids, names) = system_state.get(&world);
+    let mut i = 0;
+    for (grid, parent) in grids {
```
luca-della-vedova: Nit, can we `enumerate()` rather than have a separate `let mut i` variable?

crates/rmf_site_editor/src/site/save.rs
```diff
@@ -1759,6 +1761,39 @@ pub fn generate_site(
...
+            img.put_pixel(
+                (cell.x - grid.range.min_cell().x) as u32,
+                (cell.y - grid.range.min_cell().y) as u32,
+                image::Rgb([0, 0, 0]),
+            );
+        }
+
+        let mut path = path.clone();
+        path.push(format!("occupancy.{}.png", name.0));
+        img.save(path).unwrap();
```
luca-della-vedova: I would avoid unwrapping here, I suppose the main failure scenario here would be an invalid `PathBuf` passed to the function which should not really happen but again catching the error and logging it + continuing it would be a safer behavior

crates/rmf_site_editor/src/occupancy.rs
```diff
@@ -185,7 +198,50 @@ enum Group {
     None,
 }
 
-fn handle_calculate_grid_request(
+fn handle_export_request(
```
luca-della-vedova: I'm generally not too excited about this function since it's pretty much the exact same as the one below, with one line of difference in "what we do after `calculate_grid` is over".

My preference would be to refactor them into a single function, on top of my head I can think of a few ways:

* Refactor the two input events `ExportGridRequest` and `CalculateGridRequest` into a single event that has an additional field that says "what do we do with the output", we could match it and either send the saving event (your function) or send the replan event (the original function).
* Make the function generic, perhaps as a generic parameter a closure that takes `World` and decides what to do after the grid is exported?
* Bite the bullet and make this a workflow, which is quite a good fit for "sequence of operations that we need to run one after the other", so we could have the first node of the workflow that handles the `calculate_grid` request in a common service, then two workflows, one to calculate the occupancy grid and replan and the other one to calculate the occupancy grid and send a file dialog request.

First is quick and easy, second is possibly not a great idea, third would probably be the best but a bit more work to implement.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ opt-level = 3
 [workspace.dependencies]
 serde = { version = "1.0" }
 strum = { version = "0.26", features = ["derive"] }
-
+image = "0.24"
```
luca-della-vedova: Let's depend on `0.25` since that's the version used by bevy and add a note to keep it in sync when updating bevy.
Currently, this adds a duplicated version for `image` to the dependency tree as shown in the `Cargo.lock` changelog which is not really necessary.

crates/rmf_site_editor/src/occupancy.rs
arjo129: [d08abd1](https://github.com/open-rmf/rmf_site/pull/389/commits/d08abd14a9b5527c61951183389e8697101254cd) Introduces crossflow just within the occupancy grid.

[9fc5526](https://github.com/open-rmf/rmf_site/pull/389/commits/9fc5526c17c597fd964791b03efec1bb5c880f4d) adds crossflow to mapf_rse so that this whole thing is crossflow-ified.

crates/rmf_site_editor/src/occupancy.rs
```diff
@@ -185,8 +221,8 @@ enum Group {
     None,
 }
 
-fn handle_calculate_grid_request(
-    mut request: EventReader<CalculateGridRequest>,
+fn calculate_occupancy_grid(
+    In(BlockingService { .. }): BlockingServiceInput<()>,
```
mxgrey, line 225: This doesn't need to change but I'll just point out for future reference that this line can be simplified if the input is completely unnecessary:

```suggestion
    _: BlockingServiceInput<()>,
```

crates/rmf_site_editor/src/occupancy.rs
mxgrey, line 317: We don't particularly need to change this, but I'll point out that this is an awful lot of arguments for a function that seems to only ever be called in one location, inside a function that has almost nothing else going on.

If I were writing this, I might do something like:
* make `fn calculate_grid` into a blocking service with `CalculateGrid` as its request input
* delete `fn calculate_occupancy_grid`
* introduce a blocking service `fn get_occupancy_settings(_) -> CalculateGrid` that just fetches the relevant `OccupancyInfo` and robot entities to populate a `CalculateGrid` request
* have the occupancy grid workflows chain `get_occupancy_settings` -> `calculate_grid` instead of using `grid_svc`

mxgrey (approved):
I really like the use of crossflow here. I'd be interested in hearing what pain points you experienced while implementing this.

I did leave a couple minor comments on tweaks that I'd recommend, but they're not important enough to hold up this PR any longer, so I'm happy to approve.

PR #396, opened 2025-09-30
Fix panic when deleting new edges

### LINKED ISSUE #390: [Bug]: Editor crashes when deleting lanes

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Other Open-RMF installation methods

_No response_

### Open-RMF version or commit hash

jazzy

### ROS distribution

Jazzy

### ROS installation type

Binaries

### Other ROS installation methods

_No response_

### Package or library, if applicable

_No response_

### Description of the bug

Getting this error when I created a lane and trying to delete it:

```
Encountered an error in command `<bevy_ecs::system::commands::entity_command::insert<rmf_site_camera::CameraTarget>::{{closure}} as bevy_ecs::error::command_handling::CommandWithEntity<core::result::Result<(), bevy_ecs::world::error::EntityMutableFetchError>>>::with_entity::{{closure}}`: The entity with ID 755v3 does not exist (enable `track_location` feature for more details)

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic when applying buffers for system `librmf_site_editor::interaction::user_camera::update_camera_targets`!
Encountered a panic in system `bevy_ecs::apply_deferred`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!

```

1. Add lane
2. Delete lane

### Expected behavior

_No response_

### Actual behavior

_No response_

_No response_

luca-della-vedova:
### Fixed bug

Fixes #390 

### Fix applied

Change detection for `GlobalTransform` triggers every time the mouse is moved and it causes the component to be inserted on a deleted entity. By contrast, the one for `Transform` only triggers, correctly, when the actual anchor transform is changed.
We should investigate why `GlobalTransform` change detection seems to trigger so often or, if no real cause is found, investigate all our queries for `Changed<GlobalTransform>` and see how often they get triggered and if we can get away with a simple `Changed<Transform>` instead.
For now this PR fixes panics that occurred when deleting any edge after it was created (the original issue doesn't mention it but walls panic as well)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ testdir = "*"
 file_diff = "*"
 web-time = "1"
 
-bevy = "0.16"
+bevy = {version = "0.16", features = ["configurable_error_handler"]}
 bevy_a11y = "0.16"
 bevy_animation = "0.16"
 bevy_app = "0.16"
@@ -89,7 +89,7 @@ bevy_derive = "0.16"
 bevy_dev_tools = "0.16"
 bevy_diagnostic = "0.16"
 bevy_dylib = "0.16"
-bevy_ecs = "0.16"
+bevy_ecs = {version = "0.16", features = ["configurable_error_handler"]}
 bevy_encase_derive = "0.16"
 bevy_gilrs = "0.16"
 bevy_gizmos = "0.16"
--- a/crates/rmf_site_editor/src/interaction/user_camera.rs
+++ b/crates/rmf_site_editor/src/interaction/user_camera.rs
@@ -76,7 +76,13 @@ fn calculate_new_target(
 
 pub fn update_camera_targets(
     mut commands: Commands,
-    changed_anchors: Query<&Dependents, (With<Anchor>, Or<(Changed<Anchor>, Changed<Transform>)>)>,
+    changed_anchors: Query<
+        &Dependents,
+        (
+            With<Anchor>,
+            Or<(Changed<Anchor>, Changed<GlobalTransform>)>,
+        ),
+    >,
     changed_elements: Query<
         Entity,
         Or<(
--- a/crates/rmf_site_editor/src/interaction/user_camera.rs
+++ b/crates/rmf_site_editor/src/interaction/user_camera.rs
@@ -76,13 +76,7 @@ fn calculate_new_target(
 
 pub fn update_camera_targets(
     mut commands: Commands,
-    changed_anchors: Query<
-        &Dependents,
-        (
-            With<Anchor>,
-            Or<(Changed<Anchor>, Changed<GlobalTransform>)>,
-        ),
-    >,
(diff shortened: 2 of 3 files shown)


luca-della-vedova:
> It's quite alarming that `GlobalTransform` is being dirtied any time the mouse moves. That's worth opening an issue ticket for. I think it's important for us to safely track changes in `GlobalTransform` because I expect we'll want to support features that involve user-defined reference frames and constraints between objects, which will require us to know when an object's absolute position has changed, not just its position relative to its parent.
> 

Fully agree, this is meant more as a hotfix for a pretty breaking issue, since I expect the full investigation to be quite a bit more involved.

> Putting that aside, I'm concerned that the approach in this PR isn't as sound as it could be. It seems the fundamental issue is that we're seeing the `Changed` signal for an entity that is already deleted and then manipulating that entity anyway. In theory this could show up for more than just the `Changed<GlobalTransform>` tracking.
> 

To be honest this kind of issues pops up all the time and we have to come up with all sorts of workarounds for it (i.e. the whole `ModelTrashCan` approach in the past).
I wonder if we should just consider overriding the command error handler for invalid commands from a panic to a warning and at least make this sort of issues not crash the whole app. From my memory in the vast majority (or even all?) the cases where we got a panic because of an operation on a non existing entity it wasn't a major logic flaw where we were trying to operate on a non valid entity, but rather a system ordering issue where a system was trying to add a component to an entity that was being deleted (like this case), for which an "ignore, print a warning and continue" is a safe default.
https://github.com/open-rmf/rmf_site/pull/396/commits/90a075f46a615a4d779b2379b5d150649ba0e863 is what this would look like, now a warning will be printed whenever we have an invalid command. What was previously a crash now becomes:

```
2025-09-30T14:03:33.094499Z  WARN librmf_site_editor: Encountered an error in command `<bevy_ecs::system::commands::entity_command::insert<rmf_site_camera::CameraTarget>::{{closure}} as bevy_ecs::error::command_handling::CommandWithEntity<core::result::Result<(), bevy_ecs::world::error::EntityMutableFetchError>>>::with_entity::{{closure}}`: The entity with ID 1076v3 does not exist (enable `track_location` feature for more details)
```

What do you think of this method instead to "patch this" while we try to nail down where the spurious change detection is being triggered?
Note that in this specific case the error gets printed continuously, making me think that there is a missing cleanup in anchor dependents.

xiyuoh:
Thanks for the hotfix @luca-della-vedova , tried it out and it doesn't crash anymore! I'd personally prefer the error handler approach too to cover other cases of panic.

I do think it's valuable to have this merged in first and follow up with a proper/more reliable solution, since it is quite a disruptive bug and could risk users spending a good chunk of time designing their site only to have all the changes lost before saving (speaking from experience, though it was only 10 minutes for me 🙈 ).

Issue #337, opened 2025-07-16
Quality of life improvements to the editor


xiyuoh:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

This ticket proposes some improvements that can be made to the site editor that will hopefully be beneficial to the user experience. After spending some time interacting with the editor as a user, here's a list of suggestions I have:

- [ ] Ability to select multiple elements
  - For models, I find myself wanting to apply changes to multiple models at one go. Example of a relevant use case: rotate 10 model instances by 90 degrees. It can get tiresome to repeat this change many times, especially when working with a considerably large site.
  - For anchors, being able to select multiple of them allows for other features such as lane alignment (elaborated below)
- [ ] Lane alignment
  - We used to have this capability in the traffic editor, where adjacent lanes can be automatically aligned to be in a straight line using `Ctrl + /` (I think) if we were to select their anchors accordingly.
  - Another feature missed from the traffic editor is having lanes aligned to the X- or Y- axis if the `Shift` button is pressed during creation.
- [ ] Anchor/floor alignment
  - When creating a site from scratch (floors -> doors/walls), users may not be able to align the door/wall anchors with the edges of the floor polygon well. This is more of a nitpick, as visually I would be able to see a gap between the edges formed by different elements of the site. It would be cool if we are able to snap anchors to existing edges.
- [ ] Continuous model spawning
  - The current workflow to spawn a new instance of an existing description is to select the `+` button for the description under `Groups > Model Descriptions`, and my model would be created. After this, if I would like to spawn the same model, I'd have to press on `+` again.
  - In cases where a user wishes to spawn multiple model instances of the same description, we can have less of a back-and-forth with the `+` button, and have a new model preview ready once the previous instance has been spawned without the need for another button click.
  - If the user wishes to stop spawning this same model, they can click on `Esc` to exit the model preview mode, which is what we currently have.
- [ ] Model alignment
  - This is just a nice-to-have: identify the bounding box of models, and enable users to optionally enforce collision between models. This could be useful if I were to load many desks that should be adjacent to one another, such that their sides wouldn't overlap.
  - A bonus effort could be to have bounding box vertices snap to one another.
- [ ] Lane enhancements
  - Visual cues: this has been discussed offline and I know there has been some work planned to tackle this item, i.e. add cues to indicate lane direction for those with orientation constraints
[...]

PR #402, opened 2025-10-17
Support aligning lanes to x or y axis during creation

xiyuoh:
Targets one of the QOL items in https://github.com/open-rmf/rmf_site/issues/337.

This PR adds support for lane alignment - when creating lanes, pressing `ShiftLeft` would "lock" the end anchor transform such that the created lane would be aligned to the X- or Y- axis, depending on the cursor displacement from the start anchor. Releasing the button "unlocks" the alignment.
Hovering over an existing anchor also "unlocks" the lane alignment, so users can still select existing anchors even if the `ShiftLeft` button is pressed.

Changes also include updating the `create_edges` workflow to pass around a new `ButtonInputType` struct from `KeyboardServices`. This allows us to differentiate between `pressed`, `just_pressed` and `just_released` button interactions from within the workflow.

### DIFF
--- a/crates/rmf_site_editor/src/interaction/mod.rs
+++ b/crates/rmf_site_editor/src/interaction/mod.rs
@@ -230,7 +230,6 @@ impl Plugin for InteractionPlugin {
                     update_billboard_location,
                     update_billboard_text_hover_visualisation,
                     update_billboard_hover_visualization,
-                    apply_creation_settings,
                 )
                     .run_if(in_state(InteractionState::Enable)),
             )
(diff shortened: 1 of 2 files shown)


luca-della-vedova:
I tried this on a "low FPS" machine and noticed that there is a one frame delay between the lane update and the snapping feature, which makes the lane a bit wobbly and not clearly snapping to one axis when dragging, as shown below:

https://github.com/user-attachments/assets/3938bb03-db83-45f4-8ed6-9e990651801a

On a more personal opinion that can be argued against I wonder if it would make sense to change the transform of the `cursor` itself instead of only the anchor? The main reason is that there is now a mismatch between the position of the cursor indicator (the rotating circle) and the anchor that is being placed.
Edit _perhaps_ this could also address the frame delay issue since the position of the anchor is based on the cursor, but at the same time could have unintended consequences if we have two different systems updating the cursor transform so not too sure still

PR #403, opened 2025-10-18
Add MultiSelection Resource to enable selecting multiple entities. Up…

JohnTGZ:
### Implemented feature

This PR implements the ability to select multiple models in the site which is briefly described in issue #337 . 

User workflow:
1. Hold down the Left SHIFT key.
2. Select multiple models

One of the primary changes involves introducing the `MultiSelection` resource which is a `HashSet`  to hold multiple entities. 
While multiple entities are selected by holding down the SHIFT key:
- The `Inspect` widget is disabled.
- The `MultiSelection` resource will be updated to hold the selected entities
-  Multiple entities can be deleted at one go.

### GenAI declaration

- [X] I did not use GenAI

### DIFF
--- a/crates/rmf_site_editor/src/widgets/selector_widget.rs
+++ b/crates/rmf_site_editor/src/widgets/selector_widget.rs
@@ -22,7 +22,7 @@ use crate::{
 use bevy::{ecs::system::SystemParam, prelude::*};
 use bevy_egui::egui::{Button, Ui};
 use rmf_site_egui::{ShareableWidget, WidgetSystem};
-use rmf_site_picking::{Hover, Select, Selection, InspectionSettings};
+use rmf_site_picking::{Hover, InspectionSettings, Select, Selection};
 
 /// A widget that can be used to select entities.
 #[derive(SystemParam)]
@@ -61,10 +61,8 @@ impl<'w, 's> SelectorWidget<'w, 's> {
         let response = ui.add(Button::image_and_text(icon, text));
 
         if response.clicked() {
-            self.select.write(
-                Select::new(Some(entity))
-                .multi_select(self.inspection.multi_select)
-            );
+            self.select
+                .write(Select::new(Some(entity)).multi_select(self.inspection.multi_select));
         } else if response.hovered() {
             self.hover.write(Hover(Some(entity)));
         }


JohnTGZ:
@xiyuoh Hi, please take a look and provide any thoughts/suggestions on the implementation, thank you!

crates/rmf_site_picking/src/select/resources.rs
```diff
@@ -10,6 +11,10 @@ use web_time::Instant;
 #[derive(Default, Debug, Clone, Copy, Deref, DerefMut, Resource)]
 pub struct Selection(pub Option<Entity>);
 
+/// Used as a resource to keep track of selecting multiple entities
+#[derive(Default, Debug, Clone, Deref, DerefMut, Resource)]
+pub struct MultiSelection(pub HashSet<Entity>);
```
mxgrey: I don't think we should fragment the selection pipeline like this. Let's just update the existing `Selection` resource to contain a `HashSet<Entity>` instead of an `Option<Entity>`, and we can leave it to downstream systems to decide what to do with the hash set.

In the interest of generalizing `Selection` I think we should change it into a struct, e.g.:

```rust
pub struct Selection {
    pub selected: HashSet<Entity>,
}
```

We should get rid of the `Deref` and `DerefMut` and instead provide helper functions:

```rust
impl Selection {
    pub fn get_single(&self) -> Option<Entity> {
        if self.selected.len() > 1 {
            return None;
        }

        self.selected.iter().next().copied()
    }
}
```

JohnTGZ: Agreed. Having `get_single` as a method is a good idea and tidies up the use of `Selection` in other areas

crates/rmf_site_picking/src/select/systems.rs
```diff
@@ -40,20 +40,35 @@ pub fn selection_update(
     }): BlockingServiceInput<Select>,
     mut selected: Query<&mut Selected>,
     mut selection: ResMut<Selection>,
+    mut multi_selection: ResMut<MultiSelection>,
+    keyboard_input: Res<ButtonInput<KeyCode>>,
```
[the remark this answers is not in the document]
JohnTGZ: I think this approach is better. I was wondering what would be the best way to restrict multi-selection to say `model_instances` and not `lanes`, `anchors` etc.

To summarize:

1. We update the `Select` event to have a `multi_select` field which will determine if we are able to select multiple entities:
```rust
pub struct Select {
    #[deref]
    pub selection_candidate: Option<SelectionCandidate>,
    pub multi_select: bool,
}
```

2. Add a system that monitors `Res<ButtonInput<KeyCode>>`, where upon change in the SHIFT keyboard input, it will update the `InspectionSettings` resource.

3. The individual inspector functions will read the `InspectionSettings` resource to update the `multi_select` field of `Select` when triggering a `Select` event.

4. The `selection_update` service receives the `Select` event and reads it to determine whether to enable/disable multi-selection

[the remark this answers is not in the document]
JohnTGZ: In the midst of implementing this and there are some things Im unclear about and your advice would help:
1. Where is a good place to put the `InspectionSettings` resource? I am inclined to put it in `crates/rmf_site_picking/src/select/resources.rs` 
2. For checking if the SHIFT key goes up or down, I was thinking to extend [deselect_on_esc](https://github.com/open-rmf/rmf_site/blob/f7c20fe0add6bfc51d59b1df092e5cbe289ef6a6/crates/rmf_site_picking/src/select/systems.rs#L334) to check for `.pressed(…)` or `.released(…)` on SHIFT and then, updating the `InspectionSettings` resource accordingly.

JohnTGZ:
Alright, I have made significant changes since the last time I requested a review, so I will summarize the entire PR here:
- Multi-selection of entities in the world scene:
  - Holding down SHIFT key should enable `multi_select`:
    - Unselected entities can be added to the existing selection by clicking on them 
    - Deselect an entity by clicking on an already selected entity
  - Releasing the SHIFT key should disable `multi_select`:
    - Only a single entity can be selected. Clicking on an already selected entity just re-selects it.
    - If multiple entities are already selected, clicking on an entity will clear the current selections and result in just that one entity being selected. 

multi_select

- Added the `ViewMultiSelection` collapsible widget into the Standard Properties Panel.
  - This widget is only visible when at least 2 entities are selected. 
  - Displays all selected entities and has buttons to deselect and delete individual entities
  - Has a `MultiEditPoseWidget` which allows us to modify the XYZ Transform and yaw of all selected entities at once.

image

crates/rmf_site_picking/src/select/events.rs
```diff
@@ -23,15 +23,25 @@ pub struct Hover(pub Option<Entity>);
...
+pub struct Select {
+    #[deref]
+    pub selection_candidate: Option<SelectionCandidate>,
+    pub multi_select: bool,
+}
 
 impl Select {
-    pub fn new(candidate: Option<Entity>) -> Select {
-        Select(candidate.map(|c| SelectionCandidate::new(c)))
+    pub fn new(candidate: Option<Entity>, multi_select: bool) -> Select {
```
mxgrey: I think we can cut down on a lot of unnecessary changes in this PR by tweaking this API:

```rust
pub fn new(candidate: Option<Entity>) -> Self {
    Self {
        candidate,
        multi_select: false,
    }
}

pub fn multi_select(mut self, on: bool) -> Self {
    self.multi_select = on;
    self
}
```

This will allow us to revert the changes that were made to many of the existing uses of `Select::new`.

crates/rmf_site_picking/src/select/events.rs
```diff
@@ -23,15 +23,25 @@ pub struct Hover(pub Option<Entity>);
...
+        Select {
+            selection_candidate: candidate.map(|c| SelectionCandidate::new(c)),
+            multi_select,
+        }
     }
 
-    pub fn provisional(candidate: Entity) -> Select {
-        Select(Some(SelectionCandidate::provisional(candidate)))
+    pub fn provisional(candidate: Entity, multi_select: bool) -> Select {
```
mxgrey: Same as before, let's take the `multi_select` argument out of here and just use the `.multi_select(_)` builder pattern.

crates/rmf_site_editor/src/widgets/inspector/mod.rs
```diff
@@ -350,53 +358,81 @@ impl<'w, 's> WidgetSystem<Tile> for Inspector<'w, 's> {
                     return;
                 };
 
-                let Some(mut selection) = selection.0 else {
-                    ui.label("Nothing selected");
+                if selection.selected.is_empty() {
+                    ui.label("Nothing selected or multiple models selected!");
```
mxgrey: I'm confused by the addition of "or multiple models selected". Is this a leftover from an earlier implementation?

crates/rmf_site_editor/src/widgets/view_multi_selection.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl<'w, 's> WidgetSystem<Tile> for ViewMultiSelection<'w, 's> {
+    fn show(_: Tile, ui: &mut Ui, state: &mut SystemState<Self>, world: &mut World) -> () {
+        let mut params = state.get_mut(world);
+        if params.selection.selected.len() < 2 {
+            return;
+        }
+
+        CollapsingHeader::new("ViewMultiSelection")
```
mxgrey: Currently this is being conditionally popped in and out as its own major widget in the properties panel, separated from the existing inspector widget with several other major widgets in between. I don't think this is makes sense from a UX standpoint.

Instead of being separate from the Inspector widget, I think this should be a different mode for the inspector widget that activates when there is a multi-selection happening. I would recommend the following steps:

1. Rename this to `inspect_multi_selection`
2. Move it into the `inspector` directory
3. Remove this collapsing header
4. Change the implementation of `WidgetSystem` for `Inspector` so that when `selection.selected.len() > 1` it simply calls shows this widget instead, and doesn't show the usual inspectors at all.

crates/rmf_site_editor/src/widgets/inspector/mod.rs
```diff
@@ -350,53 +358,81 @@ impl<'w, 's> WidgetSystem<Tile> for Inspector<'w, 's> {
...
+                }
+
+                for InspectionProperties {
+                    selection,
+                    label,
+                    site_id,
+                    children,
+                } in selection_properties
+                {
+                    //todo(@johntgz) Find a better way to deal with egui id conflicts
```
mxgrey: Once we stop displaying all the inspection data during multi-selection, we won't have to worry about this.

crates/rmf_site_editor/src/widgets/multi_edit_pose.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn show_widget(&mut self, instances: Vec<Entity>, ui: &mut Ui) {
+        ui.label("Modify poses");
+
+        // Calculate centroid of selected instances
+        let mut centroid = Pose::default();
+        for instance in &instances {
+            if let Ok(pose) = self.poses.get(*instance) {
+                centroid.trans[0] += pose.trans[0];
+                centroid.trans[1] += pose.trans[1];
+                centroid.trans[2] += pose.trans[2];
```
mxgrey: Please resist the urge to write out vector arithmetic like this. It creates a high risk of small but crucial math errors that are easy to overlook. E.g. what if you accidentally wrote `-=` instead of `+=` and overlook it because it's just a few pixels of difference.

Our `Pose` data structure is for formatting data into a file, it's not a type for matrix arithmetic. Since `centroid` is meant for doing math, you should use a proper math type for it, e.g.

```rust
use bevy::math::Vec3A;
let mut centroid = Vec3A::ZERO;
let mut orientation = Angle::Deg(0.0);
...
centroid += Vec3A::from(pose.trans);
...
centroid /= instances.len() as f32;
...
orientation = pose.rot.yaw();
...
let frame = Pose {
    trans: centroid.into(),
    rot: Rotation::Yaw(orientation),
};
if let Some(new_frame) = InspectPoseComponent::new(&frame).show(ui) {
    let trans_offset = Vec3A::from(new_frame.trans) - centroid.translation;
    let rot_offset = new_frame.rot.yaw() - orientation;
    ...
        new_pose.trans = (Vec3A::from(new_pose.trans) + trans_offset).into();
}
```

crates/rmf_site_editor/src/widgets/multi_edit_pose.rs
```diff
@@ -0,0 +1,73 @@
...
+            for instance in instances {
+                if let Ok(pose) = self.poses.get(instance.clone()) {
+                    let mut new_pose = *pose;
+                    new_pose.trans[0] += trans_offset[0];
+                    new_pose.trans[1] += trans_offset[1];
+                    new_pose.trans[2] += trans_offset[2];
+
+                    let yaw_offset = match rot_offset + new_pose.rot.yaw().degrees() {
+                        deg if deg < -360.0 => Angle::Deg(rot_offset + 360.0),
+                        deg if deg > 360.0 => Angle::Deg(rot_offset - 360.0),
```
mxgrey: It's a good idea to ensure that the yaw is being wrapped to pi or 180-degrees, but this is not the right place to do it. Even if you clamp the yaw offset here, it might still cause the final yaw value to wrap past pi after you've applied it.

We should remove this code here and instead introduce the following:

```rust
pub fn wrap_to_extent(&mut value: f32, extent: f32) {
    // TODO: Subtract by extent as long as value is higher than +extent
    // TODO: Add by extent as long as value is lower than -extent
}

impl Angle {
    pub fn wrap_to_pi(&mut self) {
        match self {
            Self::Deg(value) => wrap_to_extent(value, 180.0),
            Self::Rad(value) => wrap_to_extent(value, std::f32::consts::PI),
        }
    }
}

impl Rotation {
    pub fn wrap_to_pi(&mut self) {
        match self {
            Self::Yaw(yaw) => /* TODO: apply wrap_to_pi for yaw */
            Self::EulerExtrinsicXYZ(angles) => /* TODO: apply wrap_to_pi for each value in angles */
            Self::Quat(_) => { /* do nothing */ }
        }
    }
}
```

After introducing the above, we can call `self.wrap_to_pi()` at the end of `Rotation::apply_yaw`.

crates/rmf_site_editor/src/widgets/multi_edit_pose.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        centroid.trans[0] /= instances.len() as f32;
+        centroid.trans[1] /= instances.len() as f32;
+        centroid.trans[2] /= instances.len() as f32;
+        // Use the orientation of the first instance for the centroid pose
+        if let Ok(pose) = self.poses.get(instances[0]) {
+            centroid.rot = pose.rot;
+        }
+
+        if let Some(new_centroid) = InspectPoseComponent::new(&centroid).show(ui) {
```
mxgrey: We're somewhat abusing the `InspectPoseComponent` widget here because

* In the current implementation we will only be respecting changes in yaw and discarding all other changes the user inputs.
* In the implementation that I've recommended in an earlier comment, we'll be locking the user into only having access to yaw with degrees representation. If they try to change to radians, Euler angles, or quaternions, then those UI interactions will be ignored.

I think this is okay for a first draft, but we should leave a TODO here to consider the following:
1. Should items actually be rotated about the centroid instead of being applied to each object individually? If so, should users be able to adjust the location of the centroid?
2. Should users be able to change the rotation representation to something other than degrees-yaw?
3. Should we even allow rotation to be modified? Maybe we should only allow translation.

crates/rmf_site_picking/src/select/resources.rs
```diff
@@ -2,13 +2,25 @@ use bevy_derive::{Deref, DerefMut};
...
-#[derive(Default, Debug, Clone, Copy, Deref, DerefMut, Resource)]
-pub struct Selection(pub Option<Entity>);
+#[derive(Default, Debug, Clone, Resource)]
+pub struct Selection {
+    pub selected: HashSet<Entity>,
+}
+
+impl Selection {
+    pub fn get_single(&self) -> Option<Entity> {
+        if self.selected.is_empty() || self.selected.len() > 1 {
```
mxgrey: Small nitpick, you don't need to check for `self.selected.is_empty()` here because it will naturally come out as `None` for `self.selected.iter().next().copied()`.

crates/rmf_site_picking/src/select/systems.rs
```diff
@@ -39,26 +39,53 @@ pub fn process_new_selector(
...
-            if let Ok(mut selected) = selected.get_mut(previous_selection) {
+    if let Some(selection_candidate) = selection_candidate.map(|s| s.candidate) {
+        if multi_select {
+            let Ok(mut selected) = query_selected.get_mut(selection_candidate) else {
+                return;
+            };
+
+            // If selection candidate is not in current selections, add to current selections.
+            // Else, deselect selection candidate by removing it from current selections.
+            if !selection.selected.contains(&selection_candidate) {
```
mxgrey: Very minor nitpick, but you can reduce the number of hashmap lookups with the following:

```rust
if selection.selected.remove(&selection_candidate) {
    selected.is_selected = false;
} else {
    selected.is_selected = true;
    selection.selected.insert(selection_candidate);
}
```

It's fine to leave it as-is if you feel that the current approach is more clear semantically, but if you do keep it then I would at least recommend changing it to

```rust
if selection.selected.contains(&selection_candidate) {
    selected.is_selected = false;
    selection.selected.remove(&selection_candidate);
} else {
    selected.is_selected = true;
    selection.selected.insert(selection_candidate);
}
```

because I find that `!` in if-statements is generally harder to see and interpret correctly.

crates/rmf_site_picking/src/select/systems.rs
```diff
@@ -267,19 +294,19 @@ pub fn clear_hover_select(
...
 ) {
     if let Some(previous_hovering) = hovering.0.take() {
         if let Ok(mut hovered) = hovered.get_mut(previous_hovering) {
             hovered.is_hovered = false;
         }
     }
 
-    if let Some(previous_selection) = selection.0.take() {
-        if let Ok(mut selected) = selected.get_mut(previous_selection) {
+    selection.selected.iter().for_each(|e| {
```
mxgrey: Use a regular for-loop instead of `.for_each`.

crates/rmf_site_editor/src/widgets/view_multi_selection.rs
```diff
@@ -0,0 +1,119 @@
...
+                            .on_hover_text("Deselect instance from current selections")
+                            .clicked()
+                        {
+                            self.select.write(Select::new(Some(instance_entity), true));
+                        }
+                        // Delete instance from this site (all scenarios)
+                        if ui
+                            .add(ImageButton::new(self.icons.trash.egui()))
+                            .on_hover_text("Remove instance from all scenarios")
+                            .clicked()
```
mxgrey: Notice how the `selector_widget` [also checks for `.hovered()`](https://github.com/open-rmf/rmf_site/blob/b53924b19fbefc6d43ec7d50d66e896f1b35ceea/crates/rmf_site_editor/src/widgets/selector_widget.rs#L64-L66) and then emits the `Hover(_)` event. We should do likewise to help users know which item is being considered for removal before they click.

mxgrey (changes requested):
Thanks @JohnTGZ , there are a lot of great concepts being introduced in this PR.

I've left some recommendations for how I think we can tighten up the implementation.

crates/rmf_site_picking/src/select/events.rs
JohnTGZ: Great idea!

crates/rmf_site_picking/src/select/systems.rs
```diff
@@ -452,3 +479,10 @@ pub fn send_double_click_event(
         double_clicked.last_selected_time = current_time;
     }
 }
+
+pub fn multi_select_on_shift(
+    keyboard_input: Res<ButtonInput<KeyCode>>,
+    mut inspection_settings: ResMut<InspectionSettings>,
+) {
+    inspection_settings.multi_select = keyboard_input.pressed(KeyCode::ShiftLeft);
```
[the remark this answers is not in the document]
JohnTGZ: Ah yes, great point on only updating the `InspectionSettings` resource only if there are actual changes. I'm in favour of approach (1) too since we just need to know if the key is held down, `just_pressed` and `just_released` seems to detect changes in input for that one frame which is not necessary for us.

crates/rmf_site_picking/src/select/systems.rs
JohnTGZ: I definitely agree with you, let's do away with the `!` and make it more clear. I will change to the first code segment you suggested to make it more optimized!

crates/rmf_site_editor/src/widgets/multi_edit_pose.rs
JohnTGZ: I think this is better in long-term. I will add implement this function for `Angle` in `crates/rmf_site_format/src/misc.rs` so that it can also be used by other widgets.

Issue #398, opened 2025-10-02
Harden relationships between Points/Edges/Paths and Anchors


mxgrey:
Following up on #396 there are some known issues where changes to anchor-dependent types (Point, Edge, Path) can result in inconsistent state where Anchors mistakenly think it has a dependent that it actually doesn't. There could also be cases where Anchors aren't aware of a dependent that it does have, and we just haven't caught that yet.

This problem is strongly tied into the concept of [relationships](https://taintedcoders.com/bevy/relationships) in bevy, but unfortunately bevy relationships currently only support one-to-many relationships, but anchors need many-to-many.

My instinct would be to use observers or component hooks, but I'm not aware of a way to trigger either for a modification (rather than an insert). Instead we might need to make systems that track `Changed<Point>`, `Changed<Edge>`, and `Changed<Path>`. We'll need to introduce a `LastKnown<T>` component to store the last known value of `Point`, `Edge`, or `Path` and do a comparison between the last known value and the current value every time a change is detected. Then we can subtract dependents from anchors that are no longer used and add dependents to newly used anchors.

PR #405, opened 2025-10-20
Always clear dependents from preview anchor

mxgrey:
#396 side-stepped a panic that would occur after deleting a newly created edge, but instead of panicking, the site editor would spam warning messages for the remainder of its run.

The ideal solution to the problem is described in #398, but that may take a considerable amount of thought and effort. In the meantime this PR takes a brute force approach to resolving the underlying problem by manually purging the preview anchor of any dependents during the cleanup phase of the edge creation mode.


xiyuoh (approved):
Thanks for putting in this PR, it's definitely a step towards cleaning up dependent relationships across anchors of various states.

Issue #354, opened 2025-08-01
Fix support for creating or opening new files while a site is already loaded


mxgrey:
The `File` menu allows users to select `New` or `Open` while a site is already open, but the current implementation of the site editor does not give a clean separation between the two sites, nor does it allow the user to return to the originally open site.

As a first step we should remove those menu options so users can't accidentally put the editor into a broken state.

Then eventually we should follow up with properly supporting the ability to simultaneously edit multiple sites in one editor instance.

PR #408, opened 2025-10-22
Prevent users from opening or creating new workspaces when one is already open

mxgrey:
This is a very simple PR to address the most critical problem raised by #354.

In the current implementation, if someone has a site open and then tries to open or create another site, the original site will remain loaded but permanently inaccessible because the site editor does not currently support managing multiple workspaces (sites) at once. I suspect the way we use resources will also be detrimental to loading a second site.

This PR prevents that from happening by simply removing the UI elements that would allow it. In the future we can follow up with more development to properly support multiple sites, perhaps using bevy's multi-world support.


xiyuoh (approved):
LGTM!

Nit: We might want to remove the unused imports `CreateNewWorkspace` and `WorkspaceLoader` [here](https://github.com/open-rmf/rmf_site/blob/no_workspace_changing/crates/rmf_site_editor/src/widgets/workspace.rs#L19)

PR #412, opened 2025-10-24
Update to released bevy_gltf_export

luca-della-vedova:
Set to 0.1.1 because I had a few minor dependency cleanups to add.

I bumped upstream the dependency to `image` that removed its duplication, reducing the total number of packages from 673 to 656.


mxgrey (approved):
Thanks for taking care of the release :bow:

PR #427, opened 2026-04-10
Export site file as JSON on wasm

xiyuoh:
This PR targets https://github.com/open-rmf/rmf_site/issues/423 and provides a way for users to export their site JSON on web via copy/paste. The objective is to allow users to store changes they made on the web editor.

Currently we do not support saving to file on wasm due to limitations in accessing the file system, so as a workaround this PR generates the site file to JSON string and allows the user to copy it. There could be better alternatives out there and I'm open to suggestions!

One thing to note - [`write_text()`](https://docs.rs/web-sys/0.3.77/web_sys/struct.Clipboard.html#method.write_text) is currently only enabled for secure context (e.g. `https://`, `localhost`), more details [here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations).

CC @JohnTGZ 

### To test

Follow instructions on the main README to run the editor on web. Under `File` in the top menu bar, you'll see an `Export JSON` button. Upon clicking on it a dialog will appear with the JSON string and button to copy all to clipboard.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5683,6 +5683,7 @@ dependencies = [
  "geo",
  "gz-fuel",
  "itertools 0.11.0",
+ "js-sys",
  "mapf",
  "nalgebra 0.32.6",
  "pathdiff",
@@ -5707,6 +5708,7 @@ dependencies = [
  "urdf-rs",
  "utm",
  "uuid",
+ "wasm-bindgen",
  "web-sys",
  "web-time",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5707,6 +5707,7 @@ dependencies = [
  "urdf-rs",
  "utm",
  "uuid",
+ "web-sys",
  "web-time",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ testdir = "0.9"
 file_diff = "1.0"
 web-sys = "0.3"
 web-time = "1"
+js-sys = "0.3"
 
 bevy = {version = "0.16", features = ["configurable_error_handler"]}
 bevy_a11y = "0.16"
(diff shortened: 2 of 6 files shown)


luca-della-vedova:
I'm guessing there is no way to do a "Download" instead?

xiyuoh:
@luca-della-vedova Added in [708d35c](https://github.com/open-rmf/rmf_site/pull/427/commits/708d35c45fc3d8ba8d0f7f2ff7d86e648824fea9)

JohnTGZ:
Hi XY, this feature is great! Not directly related to this PR, but now you can update the following line in the root `README.md` to highlight this new alternative feature:
```
TODO: The web assembly version is highly experimental, currently it lacks important features like saving/loading of map files.
```

mxgrey:
@JohnTGZ @xiyuoh 

Any opinions on whether the `Download file` button should also be included in the `File` menu when compiled to WASM? Is there a particular reason to have users click through two steps (`Export Json` -> `Download file`) instead of one?

xiyuoh:
@mxgrey I was thinking of both download and copy/paste features to be accessible within the `Export JSON` widget and let the users decide how they'd like to retrieve their site JSON. For example it's faster for me to copy the JSON over downloading a new file in my workflow, but others might prefer the download approach. I believe crossflow's diagram editor works the same. No strong opinions though, I'd be happy to move the `Download` option to the `File` menu directly.

JohnTGZ:
> @mxgrey I was thinking of both download and copy/paste features to be accessible within the `Export JSON` widget and let the users decide how they'd like to retrieve their site JSON. For example it's faster for me to copy the JSON over downloading a new file in my workflow, but others might prefer the download approach. I believe crossflow's diagram editor works the same. No strong opinions though, I'd be happy to move the `Download` option to the `File` menu directly.

I don't have a strong opinion either. But I think offering the ability to copy text could be convenient during development.

xiyuoh:
Added `Download File` to the main dropdown in [25d2f8d](https://github.com/open-rmf/rmf_site/pull/427/commits/25d2f8dd77ae86c379c06556951175faff583ac8)

image

xiyuoh:
Ready for review! @mxgrey

mxgrey (approved):

PR #428, opened 2026-05-13
Add diagnostic for duplicated location names

### LINKED ISSUE #346: Add diagnostic for locations with the same name
Location names are used, for example, in robot tasks to dispatch robots to go to specific places.
We should have a site diagnostic in place for multiple locations with the same name since the ambiguity might cause issues when tasks are assigned

Lots-ninety-nine:
Fixes #346.

## Problem

Location names are used as identifiers for dispatch destinations in RMF tasks. When two or more locations in a site share the same name, the dispatch target becomes ambiguous, which can cause robots to be routed to the wrong place. Today the editor accepts duplicate names silently — there is no warning surfaced by the diagnostics panel.

## Fix

Add a new `Validate` check that mirrors the existing duplicated door / lift name diagnostics:

- `crates/rmf_site_editor/src/site/location.rs` gains `DUPLICATED_LOCATION_NAME_ISSUE_UUID` and a `check_for_duplicated_location_names` system. The system reads `ValidateWorkspace` events, walks every entity with `LocationTags` that is a descendant of the validated site root, groups them by `NameInSite`, and spawns one `Issue` per name that appears more than once with the duplicate entities attached.
- `crates/rmf_site_editor/src/site/mod.rs` registers the new UUID with `IssueDictionary` (`"Duplicate location name"`) and adds the system to the same `PreUpdate` tuple that already runs `check_for_duplicated_door_names` and `check_for_duplicated_lift_names`.

The shape of the check (HashMap of name → BTreeSet<Entity>, ancestor-filter via `AncestorIter`, `Issue { brief, hint }` payload) is intentionally identical to `check_for_duplicated_door_names` so future maintainers can read the three together.

## Verification

Ran the four checks the existing CI workflows perform, on Ubuntu 24.04 / rustc 1.95:

```
cargo fmt --check                                          # style.yaml         → ok
cd crates/rmf_site_format && cargo check --no-default-features  # style.yaml    → ok
cd crates/rmf_site_format && cargo test                    # ci_linux.yaml      → ok (0 tests)
cd crates/rmf_site_editor && cargo test                    # ci_linux.yaml      → ok (0 unit + 1 doctest)
```

No new tests are added because `rmf_site_editor` does not currently host unit tests for the other `check_for_duplicated_*_names` systems either; the diagnostic is exercised through the Diagnostics panel UI. Happy to add a system test if reviewers would like one.

### DIFF
--- a/.github/workflows/ci_windows.yaml
+++ b/.github/workflows/ci_windows.yaml
@@ -3,7 +3,7 @@ on:
   pull_request:
   workflow_dispatch:
   push:
-    branches: [main]
+    branches: [main, xiyu/ci_windows]
 
 jobs:
   build:
--- a/.github/workflows/ci_windows.yaml
+++ b/.github/workflows/ci_windows.yaml
@@ -3,7 +3,7 @@ on:
   pull_request:
   workflow_dispatch:
   push:
-    branches: [main, xiyu/ci_windows]
+    branches: [main]
 
 jobs:
   build:
--- a/crates/rmf_site_editor/src/site/location.rs
+++ b/crates/rmf_site_editor/src/site/location.rs
@@ -15,9 +15,14 @@
  *
 */
 
-use crate::{layers::ZLayer, site::*};
-use bevy::{ecs::hierarchy::ChildOf, prelude::*};
+use crate::{issue::*, layers::ZLayer, site::*};
+use bevy::{
+    ecs::{hierarchy::ChildOf, relationship::AncestorIter},
+    prelude::*,
+};
 use rmf_site_picking::{Hovered, Select, Selected, VisualCue};
+use std::collections::{BTreeSet, HashMap};
+use uuid::Uuid;
 
 pub const BILLBOARD_LENGTH: f32 = 0.3;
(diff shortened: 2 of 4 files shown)


xiyuoh (commented):
Thanks for the contribution @Lots-ninety-nine ! Tested this out and works great, able to identify and diagnose duplicated location names. Can you resolve the CI build failure?

Lots-ninety-nine:
The failing `build (rolling, noble)` check looks unrelated — the failure happens in `headless_load_and_save_roundtrip` (in `save.rs`, untouched here) due to missing symlinks on Windows (see the `#[cfg(unix)]` guard in the test). My change only adds a `ValidateWorkspace`-driven system, which is a no-op in this test (the test asset has 0 Location entities and no UI to fire the event).

I re-ran `ci_windows` on my fork against the same SHA (17c8c63) and it passed: https://github.com/Lots-ninety-nine/rmf_site/actions/runs/26072356337

Pushing an empty commit to retrigger CI here.

xiyuoh (approved):
Hmm yeah I'll open another PR to fix it if it becomes a problem on main, thanks!

xiyuoh:
It still needed to be fixed before I'm able to merge this PR, done in https://github.com/open-rmf/rmf_site/pull/428/commits/5d852ce2228696f074bd06dd244f370c953c2587

PR #430, opened 2026-06-20
Customizable heights for walls and doors

mxgrey:
While playing with the site editor to think through the interior design of my upcoming home, I found that the uniform height of walls and doors creates a distorted impression of what a building really looks like.

This PR adds the ability to adjust the height of the tops and bottoms of walls and doors. The office demo site has been updated to take advantage of this feature:

Screenshot from 2026-06-20 19-17-17

To adjust the top/bottom height, select the wall or door that you want to modify and then find the height widget in the inspection panel:

Screenshot from 2026-06-20 19-21-59

You can choose between `ratio` or `fixed`:
* `ratio` - A ratio (recommended values from 0.0 to 1.0) of how high up to the next level. A value of 1.0 means it extends all the way to the elevation of the next level, if one exists.
  * If no level exists above the current one, then we fall back to `DEFAULT_LEVEL_HEIGHT` which is 3m. 
  * If the elevation of the next level up is less than 2.0m above the current one, we interpret that as a mistake and fall back on 3m.
  * If the elevation of the next level is changed, walls and doors with this setting will automatically change as well.
* `fixed` - A fixed value in meters.


crates/rmf_site_editor/assets/office.site.json
```diff
@@ -0,0 +1 @@
+../../../assets/demo_maps/office.site.json
```
arjo129, line 1: knit: While symlinks work fine on macOS/Linux they could be mildly problematic for windows developers. See: https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows

Windows needs to be put into "developer mode" for symlinks to work correctly. Im suspecting that our CI already is configured in developer mode and hence does not catch this. 

FWIW this comment might be slightly outdated in 2026 as Ive not touched windows for a while.

arjo129 (commented):
I tested this feature and it works as expected.

crates/rmf_site_editor/assets/office.site.json
mxgrey: There's a sibling of this file, [`office.building.yaml`](https://github.com/open-rmf/rmf_site/blob/main/crates/rmf_site_editor/assets/office.building.yaml) which is already a symlink used in the same way as this one: linking to a file in the assets directory so that the asset gets included in installs of the `rmf_site_editor` package. The other symlink worked for us for many years, so I assume this one won't be a problem.

However it seems Windows CI has been failing lately, and I haven't figured out why...

arjo129 (approved):

Issue #431, opened 2026-06-20
Bug: Panic when adding a directional light


mxgrey:
When adding a directional light to a site, or changing an existing light into directional, we immediately get a panic from `bevy_pbr`:

```
thread 'Compute Task Pool (5)' (172191) panicked at /home/grey/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bevy_pbr-0.16.1/src/render/light.rs:1526:18:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_pbr::render::light::prepare_lights`!
```

I assume this is related to an implementation change in one of the bevy versions, since directional lights used to work fine.

PR #432, opened 2026-06-20
Temporarily disable support for directional lights

mxgrey:
This is a temporary band-aid for https://github.com/open-rmf/rmf_site/issues/431

With this PR, we will skip the implementation of directional lights and print a warning to the user. Someone should follow up with an investigation into why directional lights have stopped working.


Cargo.lock
arjo129, line 1: Knit: If we are deleting this as a policy should we also add `Cargo.lock` should we add it to the .gitignore so that people dont accidentally commit their `Cargo.lock`

arjo129 (changes requested):
I was able to reproduce the panic. I also tested this PR to make sure that the user experience is reasonable. This temporary diabling+error message seems like a reasonable stop-gap.

I have one minor knit about `Cargo.lock`. If we are removing it we should add it to the `.gitignore` to prevent others from accidentally committing their lockfiles.

Cargo.lock
mxgrey: Fair point, and this is a source of recurring ambivalence for me across different Rust projects...

If we add `Cargo.lock` to the `.gitignore`, then I wonder if we should also delete the file from git history entirely? Otherwise various users might run `cargo update` at different times, not realize that their update isn't being committed, and then get confused when the repo can build for one person but not the other.

On the other hand, with no `Cargo.lock` at all, CI can start failing inexplicably because of dependency updates that don't follow semantic versioning as carefully as they should.

I'm not sure which situation is worse. If it were an option, I'd rather suppress these patch warnings altogether, but I can't find any way to do so.

arjo129 (approved):
I do see a use case for `Cargo.lock` for reproducible builds especially. I misread the PR on github thinking that the file was being deleted when we are actually removing unnecessary deps. Apologies for the confusion.

PR #433, opened 2026-06-22
Drag models with global coordinates, not local

mxgrey:
After rolling a model by 90 degrees, I discovered that we were accidentally dragging models in their local z-plane instead of the global z-plane. I think we never encountered this before because of how rare it is for us to pitch or roll a model.


arjo129 (approved):
I was able to reproduce the behaviour on main and this PR does indeed fix the issue. The windows CI seems to be flaky for some reason.

PR #436, opened 2026-06-27
Fix panel alignment for anchor inspector

mxgrey:
I noticed alignment for the anchor inspector widget was out of whack:

site_editor_bad_formatting

This probably happened after #424 and just wasn't noticed until now. This PR fixes the alignment:

site_editor_fixed_formatting

Also the the anchor dependents inspector widget was wrongly labeled as "Dependencies". This corrects it to "Dependents", because it lists items that depend on the currently selected anchor.

### DIFF
--- a/crates/rmf_site_editor/src/widgets/inspector/inspect_anchor.rs
+++ b/crates/rmf_site_editor/src/widgets/inspector/inspect_anchor.rs
@@ -137,45 +137,49 @@ fn impl_inspect_anchor(
                 }
             });
         } else {
-            match anchor {
-                Anchor::Translate2D(_) => {
-                    if !is_dependency {
-                        ui.label("x");
-                    }
-                    let mut x = tf.translation.x;
-                    ui.add(DragValue::new(&mut x).speed(0.01));
+            panel.orthogonal(ui, |ui| {
+                match anchor {
+                    Anchor::Translate2D(_) => {
+                        if !is_dependency {
+                            ui.label("x");
+                        }
+                        let mut x = tf.translation.x;
+                        ui.add(DragValue::new(&mut x).speed(0.01));
 
-                    if !is_dependency {
-                        ui.label("y");
-                    }
-                    let mut y = tf.translation.y;
-                    ui.add(DragValue::new(&mut y).speed(0.01));
+                        if !is_dependency {
+                            ui.label("y");
+                        }
+                        let mut y = tf.translation.y;
+                        ui.add(DragValue::new(&mut y).speed(0.01));
 
-                    if x != tf.translation.x || y != tf.translation.y {
-                        {}
-                        params.move_to.write(MoveTo {
-                            entity: id,
-                            transform: Transform::from_translation([x, y, 0.0].into()),
+                        if x != tf.translation.x || y != tf.translation.y {
+                            {}
+                            params.move_to.write(MoveTo {
+                                entity: id,
+                                transform: Transform::from_translation([x, y, 0.0].into()),
+                            });
+                        }
+                    }
+                    Anchor::CategorizedTranslate2D(_) => {
+                        warn!("Categorized translate inspector not implemented yet");
+                    }
+                    Anchor::Pose3D(pose) => {
+                        panel.align(ui, |ui| {
+                            if let Some(new_pose) = InspectPoseComponent::new(pose).show(ui) {
+                                // TODO(luca) Using moveto doesn't allow switching between variants of
+                                // Pose3D
+                                params.move_to.write(MoveTo {
+                                    entity: id,
+                                    transform: new_pose.transform(),
+                                });
+                            }
                         });
                     }
                 }
-                Anchor::CategorizedTranslate2D(_) => {
-                    warn!("Categorized translate inspector not implemented yet");
-                }
-                Anchor::Pose3D(pose) => {
-                    panel.align(ui, |ui| {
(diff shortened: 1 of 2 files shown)


crates/rmf_site_editor/src/widgets/inspector/inspect_anchor.rs
```diff
@@ -137,45 +137,49 @@ fn impl_inspect_anchor(
...
+                                });
+                            }
+                        });
+                    }
                 }
-            }
+            });
         }
+
+        ui.add_space(10.0);
```
arjo129: Interestingly, this line is panicking for me when I load the office demo world ([gist](https://gist.github.com/mxgrey/f71be4c16fbacbcc7070a87ee2289c8a))

arjo129 (changes requested):
Im getting a panic when running the demo world. 

Inside egui, `ui.add_space(amount)` allocates a sizing rectangle based on the layout's orientation. Im guessing when no anchor is selected the default grid layout is being passed as `ui` and causing the panic. FWIW, things look fine without the `add_space`. Commenting out the line stops the panic.

crates/rmf_site_editor/src/widgets/inspector/inspect_anchor.rs
mxgrey: This is super weird. I don't experience a panic at all when running the demo world and selecting anchors. I wonder if there's something specific related to your system that's leading to this?

For one thing, the call stack in the backtrace is surprisingly deep. I guess that's a consequence of the way egui uses nested callbacks to define the UI. I wonder if your system is reaching some kind of call stack limit? But I think there are function calls elsewhere in the widget system that run even deeper than this.

I'm concerned that if we just remove this line and move on with our lives then we're sweeping dust under the rug, and it's going to come out again when we least expect it.

mxgrey: I tested on a second machine and still didn't get a panic when selecting an anchor nor when using the anchor inspector UI.

I might suggest deleting your whole `target/` directory and rebuilding just in case there's some corrupted dependency that's leading to this.

arjo129: hmm... Its crashing both on my Mac OS and my linux machine. Could there be some race condition due to system ordering we are not catching? I won't block this PR on it but perhaps ticket it as an issue

arjo129: OK this is strange. When I build in `--release` mode the crash does not occur but just with `cargo run` I get a crash.

arjo129: So... upon closer inspection this is NOT a race condition but rather in debug mode `egui` adds the following `debug_assert` https://github.com/emilk/egui/blob/68b74530b7848cef6bff4efc5fc9906bfbd1e8ca/crates/egui/src/placer.rs#L152

So if you run it in release mode this will not trigger and the `ui.add_space` will silently not do anything in release mode. The fact is that `inspect_edge.rs` and `inspect_point` use a `Grid` layout which then gets passed to `InspectAnchor`. 

One option is to guard the `ui.add_space` with something when its being rendered from within the edge inspector.

mxgrey: > So if you run it in release mode this will not trigger and the ui.add_space will silently not do anything in release mode.

The add space does make a difference. If you remove it and run this in release mode (so it doesn't crash) you'll see `Dependents` squashed right below the coordinate grid with no breathing room. Admittedly this isn't a huge functional difference, but I think the subtle spacing makes a big psychological difference in the UX.

> The fact is that inspect_edge.rs and inspect_point use a Grid layout which then gets passed to InspectAnchor.

Do you mean that the `ui.add_space` is being called inside the rendering of a grid? I don't think that's supposed to be the current setup, but I'll check if that's happening accidentally.

arjo129: > Do you mean that the ui.add_space is being called inside the rendering of a grid? I don't think that's supposed to be the current setup, but I'll check if that's happening accidentally.

Yes this is exactly whats happening. We are using this widget in a grid here: https://github.com/open-rmf/rmf_site/blob/acb7134529e8ecbb00473c9fd221736526a3cde9/crates/rmf_site_editor/src/widgets/inspector/inspect_point.rs#L91

arjo129: Just putting a
```
if !is_dependency {
  ui.add_space(10.0); 
}
```

Should solve the problem.

mxgrey: Ohhhh I forgot this widget gets nested inside the edge inspector.

We ought to have a better way to automatically space out these inspector widgets, but that's a problem for another time. Adding the space conditionally is good enough for now. I'll update the PR.

mxgrey: This should do the trick: [b6567f7](https://github.com/open-rmf/rmf_site/pull/436/commits/b6567f71825c2110e5d7cd3343d12a6192914985)

Thanks for spotting this and figuring out the problem.

arjo129 (approved):
Thanks! I tested this and it no longer crashes. LGTM!

PR #437, opened 2026-07-07
Add support for specifying a custom base SDF file

arjo129:
### Implemented feature

End users may want to customize the set of sensors/plugins that get loaded at runtime. This commit adds support for loading and saving custom SDF files both in headless and GUI mode.

We expose a new export button for custom simulation commands. Clicking File > Export will now allow you to select a base SDF file. This is useful for loading our own custom plugins at runtime. 

Screenshot From 2026-07-06 09-09-20

On the CLI side you can now specify `--export-sdf-base` 

https://github.com/open-rmf/rmf_site/blob/3aa37da31d48d440f08547f8b581a58256bfda88/crates/rmf_site_editor/src/lib.rs#L105

### GenAI Use

- [x] I used a GenAI tool in this PR.
- [ ] I did not use GenAI

Generated-by: Gemini 3.1 Pro


mxgrey (commented):
Sorry for how long it's taken to get around to reviewing this.

There are a few things that surprised me about the UX on this:
1. Adding a dialog to select the base SDF just before exporting is not the path I would expect. I would have expected perhaps a separate SDF Export Settings dialog so it doesn't add clicks each time a user wants to export their SDF.
2. I'm surprised that the base SDF setting isn't saved as part of the .site.json and needs to be added again each time the site editor is restarted.
3. I would expect an option for the user to either save the base SDF filename _or_ the raw xml inside the .site.json depending on whether they want to link files or to embed the base SDF.

I like the idea of being able to specify a base SDF from the command line in headless mode, but if we follow my expectation of being able to save the base SDF information inside the .site.json, then what should the behavior be if someone specifies a base SDF from command line for a .site.json that already has a base SDF saved for it? My suggestion would be:
* If base SDF is specified from the command line and not saved inside the .site.json, apply the base SDF during export as a one-time operation (same as existing behavior)
* If base SDF is specified from the command line but is also saved inside the .site.json, exit with an error message, _unless_ the user adds `--override-base-sdf` flag at command line.

My rationale is that if someone is using headless mode to mash a base SDF onto a world, they are probably using some build tool or procedural generation tool which assumes the .site.json files are meant to have a base SDF added underneath them. If the .site.json already has a base SDF then it was probably made for that specific base SDF and not meant to have it replaced. If the user is confident in what they're doing, then they can just add the override flag.

Issue #394, opened 2025-09-24
Mismatch in glossiness of lane components


mxgrey:
The custom fragment shader introduced by #378 to show animated arrows has a subtle mismatch in its material properties compared to the rounded cap that we place at the endpoints of each lane. In the screenshot below you can see that the rounded half of the lane has a glossier texture than the straight portion of the lane.

<img width="522" height="475" alt="Image" src="https://github.com/user-attachments/assets/c00e2d07-da03-4d36-a852-523ce68c2d07" />

The material for the glossy endpoints is made [here](https://github.com/open-rmf/rmf_site/blob/ab6b005da764d04c4853d10e1be1f439dce1270f/crates/rmf_site_editor/src/site/display_color.rs#L28), and its implementation is [here](https://github.com/open-rmf/rmf_site/blob/ab6b005da764d04c4853d10e1be1f439dce1270f/crates/rmf_site_editor/src/site/assets.rs#L123-L131). It may be enough to adjust the `perceptual_roughness` and/or `metallic` fields to get them to match the color made by the shader.

PR #439, opened 2026-08-12
Update StandardMaterial used for lane shader

althea28:
## Bug fix
Difference in material glossiness for lane shader found in #394 
image

### Fix applied
The fix proposed in the original bug report is to update the material properties in the old_default_material to match the shader, but updating the lane shader material to use the same material as all the other lane components (outline, unselected/unhovered lane material, rounded ends) seems more reasonable. I've updated the standard material in the lane shader to also use the same old_default_material(...) used for the rounded ends.

Screenshot 2026-08-12 at 12 01 28 PM


arjo129 (approved):
Thanks for this. I tested this fix and it indeed solves the  visual consistency issue described.

Issue #355, opened 2025-08-04
Decluttering the UI


mxgrey:
There are some ongoing problems with the site editor related to how cluttered the UI is. The site editor aims to offer an enormous amount of functionality, but currently it is difficult for users to make sense of that functionality due in part to information overload.

We may want to consider adding tabs to the property panel to logically separate the many widgets it features. While we use egui, we could consider [`egui_dock`](https://crates.io/crates/egui_dock) to separate the different broad sets of functionality.

We could also consider affecting the 3D view based on which tab is selected. For example, if the scenarios tab is selected, then we could highlight which objects in the world are modified for the current scenario.

Issue #261, opened 2025-01-03
Reduce clutter in the right-hand panel


mxgrey:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

The core team's focus is more focused on functionality than UI/UX, so as we add capabilities we are sliding into an increasingly cluttered widget system. The right-hand UI panel in particular is becoming extremely packed with inconsistent spacing and poor visibility of widgets and features.

Users would benefit from a less cluttered and more intuitive layout of features. If someone from outside of the core team can recommend a layout or UI design that can improve the UX for the site editor's capabilities, that would carry a very high impact.

PR #442, opened 2026-08-20
Integrate egui_dock into properties panel

althea28:
## Revamping properties panel
Efforts to declutter the properties panel mentioned in #355 and #261. This PR consists the base implementation of [egui_dock](https://crates.io/crates/egui_dock/0.4.2/dependencies) into the properties panel, as suggested in #355. More discussions of possible enhancements/adjustments are included below. 

I might not be 100% familiar with all the use cases and functionalities in the properties panel, please let me know if there are more ideal setups / adjustments for better UX :D

Before: Collapsible Headers
image

After: Tabs using egui_dock
image

Video demo: https://github.com/user-attachments/assets/46b82072-68f3-47aa-9e19-a268cbe69305

### Current behaviour
- Each header in the original panel has been added as a tab
- On load, all tabs are opened and the panel is divided into 2 windows - upper window (levels, scenarios, models, navigation) + lower window (layers, inspector, groups, lights). This was added as a separator between these 2 groups of headers was included in the original UI
- Each tab can be closed, reopened, added to an existing window, created as a new standalone window (see video demo above)
- Each window is collapsible (via the arrow button on the left top corner of each window)
- 'Tabs' button included at the top left corner of the left panel to manage open/closed tabs. Can be used to reopen previously closed tabs
- The full properties panel is anchored to the right and is resizeable width-wise
- Individual windows can be resized both height- and width-wise

Additional note: The building preview plugin only consists of 1 button in the original UI. It has been added as the "Preview" tab in this implementation, but I'm not sure if having a tab just for 1 button is a bit awkward. Let me know if there is a better placement for this.

### Implementation overview
Key changes:
- building_preview.rs: Extraction of rendering logic into the new show_widget function to allow Building Preview to be rendered as a tab.
- inspector/mod.rs:  Extraction of rendering logic into the show_inspector function to allow Inspector to be rendered as a tab. Slight logic refactoring done within the new show_inspector function to avoid borrow checker errors.
- view_model_instances.rs: Removed height limit so that each tab fills up the entire window height
- Removed dead code in view_levels.rs, view_scenarios.rs, view_model_instances.rs, view_nav_graphs.rs, view_layers.rs, view_groups.rs, view_lights.rs, and building_preview.rs, namely the WidgetSystem since it is no longer called in the properties panel.
- widgets/mod.rs: Initialising resources & new properties panel
- New file properties_panel.rs added to initialise the new properties tab. The general implementation here is modelled after these [examples](https://github.com/anhosh/egui_dock/blob/main/examples) included in the egui_dock repo

Each tab state seems to be correctly persisting, let me know if there are edge cases I should be trying for this.

### Other alternatives
#261 requested recommendations for other UI layouts. Below are some screenshots from the Blender and Unity UI which I think are quite neat(!) and are similar to the functionality of the properties panel, for consideration.

Blender's UI:
image
- A few possible versions of how the tab headers can be stacked/compacted
- Note: egui_dock currently only supports horizontal scrollable string headers. If the above are of interest, I can look into modifying the mod to include vertical stacking and image icon headers 

Unity's UI:
image
- The tabs function similarly to egui_dock
- For collapsible headers within tabs, Unity includes icons next to the headers & more obvious formatting for these collapsible headers for better visibility. This may be applicable to some of the tabs in site editor's properties panel which also have internal collapsible headers, such as the Layers tab


arjo129 (commented):
I really like this UX improvement it is a huge clean up over the old stuff we had and much easier to navigate.

crates/rmf_site_editor/src/widgets/properties_panel.rs
```diff
@@ -0,0 +1,311 @@
...
+        view_model_instances::ViewModelInstances,
+        view_nav_graphs::ViewNavGraphs,
+        view_scenarios::ViewScenarios,
+    },
+    AppState,
+};
+use rmf_site_egui::*;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum PropertyTab {
```
arjo129: Food for future thought: Do you think there is a good way to make this dynamic? One common use case for the site editor is that it gets embedded in other applications.

althea28: Noted! I've made the following changes to update the implementation, following Tuesday's discussion:

- In rmf_site_egui/src/properties_panel/mod.rs:
  - Added a 'Name' field in to hold the tab name
  - Added a 'TabGroup' field to specify if the tab should be in the top or bottom dock on startup
- For all view_*.rs, inspector.rs, building_preview.rs widgets:
  - Reverted back to original implementation (with the original WidgetSystem implementation)
  - Removed collapsing header (used in the old properties panel)
  - Include tab name and group (for tabs in top dock) when initialising the PropertiesTilePlugin
- In properties_panel.rs,
  - Dropped the PropertyTab enum to use DockState\<Entity\> instead
  - Use world.try_show_in(...) to render tiles
  - Added logic to group tabs in the top or bottom dock based on the 'TabGroup' field

Let me know if this is suitable!

crates/rmf_site_editor/src/widgets/properties_panel.rs
```diff
@@ -0,0 +1,233 @@
...
+                ui.horizontal(|ui| {
+                    egui::ComboBox::from_id_salt("tabs_manager_menu")
+                        .selected_text("Tabs")
+                        .width(0.0)
+                        .height(400.0)
+                        .show_ui(ui, |ui| {
+                            for tab in &tabs {
+                                let is_open = state.dock_state.find_tab(tab).is_some();
+                                let title =
+                                    world.get::<Name>(*tab).map(|n| n.as_str()).unwrap_or("Tab");
```
arjo129, line 191: knit: I'd probably log a warning/error if we had someone who didnt add a name to their tab rather than defaulting to "Tab".

crates/rmf_site_editor/src/widgets/properties_panel.rs
```diff
@@ -0,0 +1,233 @@
...
+            if let Some(index) = state.dock_state.find_tab(&tab) {
+                state.dock_state.remove_tab(index);
+            }
+        }
+
+        let mut style = Style::from_egui(context.style().as_ref());
+        style.tab_bar.show_scroll_bar_on_overflow = true;
+        style.tab_bar.fill_tab_bar = false;
+        style.tab_bar.height = 24.0;
+        style.buttons.close_tab_active_color = egui::Color32::from_rgb(240, 80, 80);
```
arjo129: Knit: use a constant for this.

arjo129 (approved):
I really don't have much feedback now, other than this is awesome! Thanks for the quick iteration on the feedback we gave.

crates/rmf_site_editor/src/widgets/properties_panel.rs
mxgrey: How about we make the tab name a mandatory argument of `PropertiesTilePlugin::new(~)`?

That breaks backwards compatibility slightly, but the migration is very easy. Then the name field doesn't need the Option.

althea28: I've added the name as a mandatory argument of PropertiesTilePlugin! I've also added a missing tab name warning in fn title for the edge case where a downstream user manually spawns a child entity of Properties Panel directly with no name component (let me know if this is unnecessary).

crates/rmf_site_editor/src/widgets/properties_panel.rs
althea28: Done!

arjo129:
One more thing I forgot is to mention to update the screenshot in the readme.

althea28:
> One more thing I forgot is to mention to update the screenshot in the readme.

Done! My bad for missing this

arjo129:
NP, Ill merge this once the CI passes.

PR #443, opened 2026-09-10
Fixes a bug where all models were spawned at the origin

arjo129:
### Fixed bug

Currently adding a model using the add description spawns a preview of the model at the point of the mouse cursor. Upon clicking, one would expect the model to be spawned at the mouse cursor's position. However, it seems to be spawned at the origin. This commit fixes this behaviour.

Before:
bug_hd

After:
fixed_bug_hd

### Fix applied
Preserve the initial element values before spawning.

### GenAI Use

- [x] I used a GenAI tool in this PR.
- [ ] I did not use GenAI

Generated-by: Gemini

### DIFF
--- a/crates/rmf_site_editor/src/site/property_plugin.rs
+++ b/crates/rmf_site_editor/src/site/property_plugin.rs
@@ -389,11 +389,12 @@ mod tests {
         // Create a root scenario
         let root_scenario = app
             .world_mut()
-            .spawn((ScenarioModifiers::<Entity>::default(), Affiliation::<Entity>(None)))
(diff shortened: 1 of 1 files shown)


althea28:
LGTM!! I noticed the ci_windows build failed, but I ran the tests locally and they passed fine. Seems like it's just a CI environment constraint or a flaky runner. It probably just needs a re-run?

arjo129:
Yeah the windows run time has been flaky for a while. I think it might be some race condition in the gui startup world but Im not actually sure.

PR #444, opened 2026-09-10
Fix panic when two measurements share a vertex

R-sharing:
Loading a `.building.yaml` panics if two measurements share an endpoint:

```
thread 'Async Compute Task Pool (2)' panicked at
crates/rmf_site_format/src/legacy/building_map.rs:412:75:
called `Option::unwrap()` on a `None` value
```

The process segfaults right after, so nothing loads and `--export-nav` / `--export-sdf` write nothing.

In traffic_editor it is common to set the scale of both axes from one corner of the drawing, so both measurements point at the same vertex. The first one moves that anchor into the drawing, the second one calls `remove_entry` on it again and gets `None`.

`office.building.yaml` has all six measurement endpoints on separate vertices, so the demo map never hits this.

Fix: move the anchor only if it is still in `level_anchors`. Two measurements sharing one drawing anchor is fine. Added a test that reproduces the panic.

Found this on a real 13 level building map where every level sets both 90 m and 24.68 m from the same corner of the floor plan.

### DIFF
--- a/crates/rmf_site_format/src/legacy/building_map.rs
+++ b/crates/rmf_site_format/src/legacy/building_map.rs
@@ -407,11 +407,13 @@ impl BuildingMap {
                     let mut site_measurement = measurement.to_site(&vertex_to_anchor_id)?;
                     let edge = &mut site_measurement.anchors;
                     // Remove the measurement anchors from the level anchors, since they belong to
-                    // the drawing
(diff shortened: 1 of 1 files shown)


arjo129 (approved):
Reading through this, the fix is simple enough. LGTM.

